Issue #32, opened 2026-07-26
Complete match that prefixes a longer pattern always waits the full flush window — add a shorter prefix-hold or per-command override


jinwoo1601:
Follow-up to #25's eager flush. `eagerFlushOnCompleteMatch` correctly refuses to eager-commit a pattern that is a compatible prefix of a longer pattern — in The Long Burn's grammar, `orient to heading {heading}` prefixes `orient to heading {heading} mark {mark_sign} {mark}`, so every **plain** heading order (the common case) holds the buffer for the full flush window (2.0 s of dead air) to keep the mark form speakable.

The hold is structurally correct, but paying the *entire* silence window for it is more caution than the case needs: a speaker continuing into "mark…" does so almost immediately.

Request — either of:
- a separate, shorter configurable hold for the "complete match, but extendable" state (e.g. `prefixHoldSeconds` < the full flush window, ~0.5–0.8 s): if no continuation starts within it, commit the complete match;
- a per-command override to opt a pattern out of the prefix hold entirely.

(Push-to-talk via `ReleaseTalk` is the documented way around this and we intend to adopt it eventually, but it needs an interaction design for hand-tracking users first — a keyed flush shouldn't be the only path to low latency.)

Issue #47, opened 2026-07-29
Document the matching/scoring model — score formula, penalties, selection order, gates, eager-flush verdicts


jinwoo1601:
## Summary

The matching/scoring model — the thing that decides whether a spoken command fires — is undocumented. Its parameters are `internal const`s in `VoxrCommandParser`, its selection rules live only in code, and each release adds another interacting knob (dynamic denominator in 1.2.1, eager flush in 1.2.2, `skippedWordPenalty` and `prefixHoldSeconds` in 1.4.0). A consumer tuning `minScore`/`minConfidence`, diagnosing a rejection, or authoring patterns that won't shadow each other currently has to reverse-engineer the model from source — which one consumer project did repeatedly, across at least three separate debugging cycles in a single month.

This matters more since 1.3.0: the auto-exported session log surfaces per-attempt `score`, `minScore`, `aggregateConfidence`, and `rejectReason` for every utterance — exactly the right diagnostic data, but **uninterpretable without the formula behind it**. A rejection at `score 0.50 < minScore 0.60` only became actionable (a dropped function word on a 3-element pattern) after reading `TryMatchScored`.

## What a "How matching and scoring works" reference should cover

1. **The score formula**: raw score accumulation per element (`MatchScore` 1.0, `OptionalLiteralScore` 0.5, `RequiredSlotMissPenalty` −1.0, `RequiredLiteralMissPenalty` −0.5) and the dynamic denominator (required elements always count; optionals only when matched — the 1.2.1 change), with the consequence that short patterns are disproportionately fragile to a single dropped word (a 3-element pattern missing one word ≈ 0.50 and fails the default gate; a 7-element pattern absorbing the same drop ≈ 0.79 and passes).
2. **The skipped-word penalty** (1.4.0): applied after selection, `[unk]` never charged, counting restarting per extracted command — and the explicit note that it filters via `minScore` but does not change which pattern wins.
3. **Selection and tie-break order**: earliest start → highest score → most matched literals → (currently) pattern listing order, and that `TryEagerCommit` mirrors it. Authors need this to understand why sibling patterns shadow each other and why pattern order in the asset can matter.
4. **Sequential multi-command extraction**: how one utterance can yield several commands, and what that means for patterns that are prefixes of others.
[...]

PR #48, opened 2026-07-29
Prefer the longer consumed span on equal-score ties (#41)

### LINKED ISSUE #41: Equal-score tie-break has no span preference — first-listed pattern wins and can split one utterance into two commands
## Summary

`VoxrCommandParser.ParseInternal` selects the winning (command, pattern) by earliest start → highest score → most matched literals, with **no preference for the longer consumed span**. All comparisons are strictly-greater, so an exact tie keeps the first candidate encountered — and iteration order is registration order, meaning **whichever pattern is listed first in the asset wins**. When a command has a bare pattern and a tailed sibling (`intercept track {track}` / `intercept track {track} {burn_level}`), an utterance carrying the tail ties both at score 1.0 with equal literal counts; if the bare pattern is listed first, it wins, and sequential extraction then matches the orphaned tail as a **separate command**.

## Observed in-headset (2026-07-26)

Grammar: `Cmd_InterceptTarget` with `intercept track {track}` listed before `intercept track {track} {burn_level}`.

Spoken: **"intercept track one two zero one maximum burn"**.

Result (confirmed in `Library/VoxrDebugLogs/session-2026-07-26_14-34-48.json`): both patterns matched at token 0 with score 1.0 and 2 literals → tie → first-listed bare pattern won, consuming through the track digits. The remaining "maximum burn" then matched a bare `{burn_level}` pattern as a second, independent command.

Consequence: the consuming game applies a policy at the command level ("an intercept at maximum burn is refused as a whole order"). Because the parser delivered *two* commands instead of one, the policy check never saw a maximum-burn intercept — the intercept executed and the refusal was silently bypassed. The failure is invisible: no warning, no reject, two "successful" commands.

## Root cause

`ParseInternal` selection loop (and the mirrored loop in `TryEagerCommit`):

```csharp
if (matchResult.Score > 0f &&
    (bestScore <= 0f ||
     startIdx < bestStartIdx ||
     (startIdx == bestStartIdx &&
      (matchResult.Score > bestScore ||
       (matchResult.Score == bestScore && matchResult.LiteralCount > bestLiteralCount)))))
```

No term involves `matchResult.EndIdx`. Asset array order is effectively load-bearing for correctness, which is documented nowhere and collides with `MatchedPatternIndex` consumers (reordering patterns to work around this renumbers the indices content may key on).

## Proposed fix

At equal start **and** equal score, prefer the match with the greater consumed span (`EndIdx`), then literal count, keeping first-listed only as the final deterministic fallback. Apply identically in `TryEagerCommit` so the eager verdict keeps matching what `FlushBuffer` will fire.

## Tradeoff note

Longer-span preference is greedier: a speaker who intended two chained commands in one utterance ("intercept track 1201" + a standalone "maximum burn") now gets the single tailed command. Scoping the change to *exact ties only* (not a general coverage bonus) limits it to outcomes that are currently order-dependent coin flips — grammars without tied sibling patterns see zero change.

## Workaround in use

Reordering `patterns[]` longest-first in every tailed command asset. Fragile (one designer reorder away from regressing, silently) and coupled to `MatchedPatternIndex`-keyed content.

jinwoo1601:
Closes #41

## What

Pattern selection now prefers the match with the **greater consumed span** when two candidates tie on start index *and* score, ahead of the existing literal-count tie-break. Registration order remains the final deterministic fallback.

The comparison was duplicated verbatim in `ParseInternal` and `TryEagerCommit` — the very drift hazard the issue calls out ("apply identically in `TryEagerCommit`") — so it is now a single `IsBetterCandidate` helper both scans call. Extraction is behaviour-preserving; the semantic change is the new span term.

Span is compared on a new `MatchResult.ConsumedEndIdx` — the index after the last element that actually matched — rather than on `EndIdx`. `TryMatchScored` skips `[unk]` before *every* pattern element, including a trailing optional that then matches nothing, so `EndIdx` overstates coverage and a pattern could otherwise win a tie by absorbing filler. `EndIdx` keeps its existing meaning for `searchStart` and the eager whole-buffer gate.

## Why

`intercept track {track}` and `intercept track {track} {burn_level}` both score 1.0 with equal literal counts on an utterance that carries the tail. With no span term the tie fell through to array order, so whichever pattern the asset listed first won. When the bare one won, sequential extraction matched the orphaned tail as a **separate** command — one spoken order delivered as two, both marked successful, no warning. Any policy the game applies at the command level (e.g. "an intercept at maximum burn is refused as a whole order") never saw the combination.

Asset array order was effectively load-bearing for correctness, and the workaround (listing patterns longest-first) collides with content keyed on `MatchedPatternIndex`.

## Blast radius — wider than "just the ties this fixes"

The span term sits **above** literal count, which means it does not only settle previously order-dependent ties. Any two candidates at the same start index with equal scores now resolve by span, **including pairs that literal count used to decide on its own** — deterministically, in either declaration order. Two consequences worth reviewing against your grammars:

- A pattern with fewer literals but wider coverage now wins where it used to lose: `fire at {target}` beats `fire at hotel` on "fire at hotel one" (2 literals/4 tokens vs 3 literals/3 tokens), changing both `MatchedPatternIndex` and the extracted slot value.
- Because the comparison runs across the whole command list, this can change which **intent** fires, not just which pattern index: `go {place}` beats `go {dir}` on "go north pole".

Grammars whose equal-scoring candidates always agree on span and literal count are unaffected. The stated tradeoff also stands: a speaker who intended `intercept track 1201` followed by a standalone `maximum burn` in one breath now gets the single tailed command.

## Eager flush

Applying the same ordering in `TryEagerCommit` keeps the eager verdict naming the command the subsequent flush will fire. Two verdict transitions follow, only the first of which is a pure win:

- `None → Commit` — a tailed utterance whose winning match used to stop short of the buffer end now commits early instead of paying the full `bufferWindow`.
- `None → HoldExtendable` — where the newly-preferred pattern is itself a prefix of a longer sibling, the buffer now reports a complete-but-extendable match. That arms the shortened `prefixHoldSeconds` hold (#32) instead of the full window, so it shortens the window in which a legitimate continuation can still arrive. With a three-deep grammar and a pause before the final word, this can fire the shorter pattern and orphan the tail. This is `prefixHoldSeconds` doing its designed job on an honest whole-buffer match — previously the same buffer got conservative treatment only as a side effect of the selection bug — but it is a real behaviour change and needs both `eagerFlushOnCompleteMatch` **and** a non-zero `prefixHoldSeconds` to be observable.

## Validation

- [x] Full Unity suite via the host project `VoXR TestGround` (Unity 6000.4.7f1), **both** platforms, at the head of this branch: EditMode 97/97 and PlayMode 271/271, 0 failures, no compile errors.
- [x] 10 tests covering the change:
  - `SpanTieBreak_TailedPatternWins_WhenBareSiblingListedFirst` — the reported repro; asserts **one** command with both slots filled, not two.
  - `SpanTieBreak_OutcomeIndependentOfPatternOrder` — same outcome with the patterns declared in the opposite order.
  - `SpanTieBreak_BareUtterance_StillMatchesBarePattern` — without the tail the tailed pattern misses a required slot and scores `(1+1+1-1)/4 = 0.5`, so the bare sibling wins on score outright; the preference really is tie-scoped.
  - `SpanTieBreak_StandaloneTailStillMatchesItsOwnCommand` — a tail spoken alone is still its own command.
  - `SpanTieBreak_LongerSpanBeatsHigherLiteralCount` — pins the demotion of literal count described above.
  - `SpanTieBreak_ChoosesBetweenCommands_NotJustPatterns` — pins the cross-command intent flip.
  - `SpanTieBreak_TrailingUnkCannotWinTheTie` — a stray `[unk]` must not flip the reported pattern index.
  - `TryEagerCommit_PrefersLongerSpanOnTie_LikeParseInternal` — the eager scan selects the same pattern `ParseInternal` fires, returning `Commit` where it previously returned `None`.
  - `TryEagerCommit_SpanTieCanYieldHoldExtendable` — pins the second transition.
  - `TryEagerCommit_TrailingUnkStillHoldsTheFullWindow` — the whole-buffer gate still treats a trailing `[unk]` as an in-progress tail.
- [x] No native-bridge changes, so no `libvosk-bridge.so` rebuild was required.
- [ ] On-device (Quest) check for the **timing** effects only: the parser logic itself is pure C# and fully covered by PlayMode tests, but the two eager-flush verdict transitions above change *when* the buffer flushes. Worth confirming in a headset for any grammar that sets `eagerFlushOnCompleteMatch` together with a non-zero `prefixHoldSeconds` — particularly the `None → HoldExtendable` case, where a shortened hold could cut off a continuation.

## Review

An independent multi-agent review ran on this PR (slim profile, 6 angles, every finding adversarially verified) and posted its verdict as a comment. It raised 6 confirmed findings, no blockers. All 6 are addressed in this branch: the two claim defects (this body and the CHANGELOG understating the blast radius; a wrong 0.75 score figure) are corrected above, the three coverage gaps have tests, and the trailing-`[unk]` span inflation is fixed in code via `ConsumedEndIdx`.


jinwoo1601 (commented):
## Independent review — PR #48

**Depth profile:** `slim` (issue lane). Angles run: `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP` — all six, none dropped.
**Head reviewed:** `38fec01604d2cbc0af41d9b34f51f66bdb37b64b`. **Merge-base:** `0adcbecb8d300a475f49f5a1f26643d531472285`.

Finder agents ran with no context from the session that wrote the code; every surviving finding was then passed to a separate verifier instructed to refute it and to default to REFUTED when uncertain. **6 findings survived verification, all CONFIRMED. No blockers.**

The core of the change verified clean. `LINE` re-derived `IsBetterCandidate` term-for-term against the two inline conditions it replaces and found them equivalent apart from the intended `EndIdx` term; `GONE` accounted for all 12 deleted lines; `PIT` cleared the `in`-on-mutable-struct defensive-copy question (field-only reads, no copy), confirmed the float-equality tie is exact rather than lucky (scores accumulate only from ±1.0/±0.5/0.0, so a full match is bitwise `1.0f`), and confirmed the helper sits outside `#if UNITY_EDITOR` so player builds compile; `XFILE` found no cross-file contract broken and — usefully — checked whether any *existing* grammar flips: `Samples~/CommandRecognition/AssetAuthoring/Commands/Cmd_SetHeading.asset` and the `MakeNumericCommands` test grammar both have sibling pairs, but in both the longer sibling already carried the higher literal count, so it won before and still wins.

---

### F1 — `CHANGELOG.md:12` and PR body: the "only coin flips" scope claim is false — **major**, CONFIRMED

The entry states the preference *"is scoped to exact ties, so it only decides outcomes that were previously order-dependent coin flips; grammars without tied sibling patterns are unaffected."* That inference is invalid. The span term is inserted **above** literal count (`Runtime/Commands/VoxrCommandParser.cs:624-626`), and the base comparator used strictly-greater on `LiteralCount` over the key `(start↑, score↓, literals↓)`, so at equal start and score with **unequal** literal counts the winner was fixed **in either registration order** — deterministic, not a coin flip. Exactly that class can now flip.

Reachable with ordinary grammar features, no sibling patterns required: `set maximum` (2 literals, `EndIdx` 2, score 1.0) vs `set {burn_level}` where `burn_level = "maximum burn"` (1 literal, `EndIdx` 3, score 1.0) on "set maximum burn". Before: the literal-anchored pattern won regardless of order. After: the slot pattern wins regardless of order.

The **code is not implicated** — issue #41 prescribes "prefer the greater consumed span (`EndIdx`), then literal count", and the code implements exactly that. The defect is the scope claim. `CHANGELOG.md` is currently the *only* shipped description of selection order (no `Documentation~/` file states a tie-break order; that is deferred to #47), so this wording is the sole signal a user gets about a silent selection change. Suggested correction: state the actual blast radius — equal-score candidates at the same start index whose spans differ, **including** cases literal count used to decide.

### F2 — no test covers span-vs-literal-count disagreement — **minor**, CONFIRMED

F1's behaviour change has no regression signal. The verifier compiled the real parser plus 124 tests from `Tests~/Runtime/VoxrCommandParserTests.cs`, `Tests~/Runtime/VoxrEagerCommitTests.cs` and the two Editor diagnostic suites into a standalone harness, instrumented `IsBetterCandidate`, and ran everything under both the new and the pre-PR comparator. Across all 124 tests, **every** comparison that reaches the span/literal stage has span and literal count pointing the *same* direction, except the two new ones — which tie at equal literal counts by construction. Toggling the span term off fails exactly **2 of 124** tests, both added by this PR.

So no test anywhere — new or pre-existing, including `Cmd_SetHeading.asset` and `MakeNumericCommands` — constrains the case where span and literal count disagree. That case is reachable and user-visible: patterns `["fire","at","hotel"]` (3 literals, span 3) and `["fire","at","{target}"]` (2 literals, span 4) on "fire at hotel one", both scoring 1.0 at start 0:

```
span term OFF -> patIdx=0, target=<empty>
span term ON  -> patIdx=1, target="hotel one"
```

Both `MatchedPatternIndex` **and extracted slot content** change. (Detail correction to the finding as filed: the new eager test ties at 1 literal each, not 2 — substance unaffected.)

### F3 — `EndIdx` counts trailing `[unk]` the pattern never matched — **minor**, CONFIRMED

`TryMatchScored` runs its `[unk]`-skip loop before *every* element including a final one that then matches nothing (`VoxrCommandParser.cs:656`), and the unmatched-optional branch does not restore `tokenIdx`, which is returned as `EndIdx` (`:732`). So on tokens `["fire","[unk]"]`, `["fire"]` gives score 1.0 / 1 literal / `EndIdx` 1 while `["fire","{?mode}"]` gives the same score and literal count with `EndIdx` 2 — and the new term now deterministically prefers the candidate that "consumed" only filler.

Two consequences, of different weight:

- **Default config, no settings required:** with the bare pattern declared first, `"fire"` reports `MatchedPatternIndex` 0 but `"fire [unk]"` reports 1. Stray recogniser noise flips the reported pattern index for the identical spoken command — against the grain of this PR, whose stated purpose includes making that index stop depending on incidental factors.
- **Eager flush:** it defeats the gate whose own comment says *"anything left over (including trailing `[unk]`) means an in-progress tail"* (`:1071`). Bounded, though: a trailing optional is never terminal (`IsTerminalPattern`, `:1188-1190`), so `Commit` stays unreachable and the verdict is at most `HoldExtendable`; it needs `eagerFlushOnCompleteMatch` **and** `prefixHoldSeconds`, both non-default; and pre-change it was an order-dependent coin flip.

Reachability caveat, stated plainly: the tie requires a grammar declaring a bare pattern *and* a sibling whose extra tail is **optional** (`{?slot}` / `?literal`) — a redundant authoring shape. Issue #41's own repro has a **required** tail, which misses and scores below 1.0, so it never ties and is unaffected. No new test uses a trailing `[unk]`.

Possible remedy if you want it closed: compare on a span that excludes trailing tokens no pattern element consumed, rather than on the raw end index.

### F4 — score figure is 0.5, not 0.75 — **minor**, CONFIRMED

`Tests~/Runtime/VoxrCommandParserTests.cs:1139-1140` (and the matching PR-body validation bullet) says the tailed pattern *"misses a required slot and scores 0.75"*. `RequiredSlotMissPenalty` is `-1.0f` applied as a **negative numerator** term (`VoxrCommandParser.cs:34, 686-690`), so the real value is `(1+1+1-1)/4 = 0.5`. `0.75` is what you get if a miss contributed `0` — a model the code does not implement and `Documentation~/command-recognition.md:102` contradicts.

No assertion depends on it (the test checks intent, slots and pattern index, not score), and the figure is inert inside `Parse`, which gates only on `score <= 0f`. It misleads only when carried up to the recogniser, where `minScore = 0.6` sits between the two numbers and 0.5-vs-0.75 genuinely flips accept/reject.

### F5 — cross-command span ties flip **intent**, untested — **minor**, CONFIRMED

`IsBetterCandidate` is called inside `for (int ci = 0; ci < _commands.Length; ci++)` and its true-branch rewrites `bestCommandIdx` (`:356, 369, 383`), so the new term moves winners across commands, not just across patterns of one command. Because the denominator is dynamic, any two commands matching fully at the same start tie at 1.0.

Concrete, from a grammar already in this repo: `go_dir = ["go","{dir}"]` / `go_place = ["go","{place}"]` with `dir = "north"`, `place = "north pole"` on "go north pole" — 1 literal each, spans 2 vs 3. Base fired `go_dir` and orphaned "pole"; head fires `go_place`. (`dial3`/`dial5` on "dial 1 2 3 4 5" is the same shape — arguably improved by the change, but equally unclaimed.) Those grammars exist at `Tests~/Runtime/VoxrEagerCommitTests.cs:150-174` but their tests only assert `CanCommitEarly`, a static precompute that never enters the selection loop.

All five new tests tie two patterns of a *single* command, so no assertion anywhere depends on the span term choosing between two commands.

### F6 — `None → HoldExtendable` is untested, and the deferral note is too narrow — **minor**, CONFIRMED

The new eager test pins only the `None → Commit` transition. The verifier compiled `VoxrCommandParser.cs` standalone and ran both commits to establish a second, unclaimed transition:

```
grammar : intercept {track} | intercept {track} {burn_level} | intercept {track} {burn_level} now
utterance: "intercept alpha maximum burn"
BASE 0adcbec : verdict=None            Parse -> intercept#0
HEAD 38fec01 : verdict=HoldExtendable  Parse -> intercept#1
```

A control with the bare sibling removed returns `HoldExtendable` on both commits, proving the flip comes from the new `EndIdx` term rather than from `CanCommitEarly` (unchanged). A two-deep grammar with word-prefix ambiguity in the trailing slot flips the same way.

This is **materially distinct from the tested path**, and it is the sharpest item in this review. `None → Commit` fires a pattern nothing can extend — a pure latency win. `None → HoldExtendable` shortens the window in which a *legitimate continuation* may still arrive, so it can change **which** pattern fires: in the grammar above with a 1s pause before "now", base waited `bufferWindow` and flushed `intercept#2`; head flushes at `prefixHoldSeconds` and fires `intercept#1`, **orphaning the tail — the same failure shape #41 exists to remove.**

Requires `eagerFlushOnCompleteMatch = true` (default `false`) **and** `prefixHoldSeconds ∈ (0, bufferWindow)` (default `0f`), so default configurations are unaffected. The PR's deferral note names only the first of those two preconditions, and both the note and the CHANGELOG describe the eager-flush effect solely as "commits early", which does not reach this case.

---

### Considered and killed

- **"CHANGELOG entry is ~600 words; house style is much shorter"** (`LINE`, out-of-angle) — refuted on measurement: the new entry is 243 words against existing 1.4.0 entries of 103 / 211 / 216 / 253. It is in range.
- **"`in MatchResult` on a non-`readonly` struct causes a defensive copy in the hot loop"** (raised as an aside by `GONE` and `GAP`) — refuted by `PIT`: the body reads only fields, never an instance method or property, so the compiler emits direct loads through the readonly ref.
- **"`ZeroAllocPollPathTests` does not cover the parse path"** (`GAP-4`) — accurate, but it is an error in *my own review brief*, which cited that file as enforcing the zero-alloc invariant when its two tests exercise `VoxrJsonParser` only. A defect in the brief, not in this PR, so it is not carried as a finding. Worth knowing that the parser's zero-alloc invariant is currently held by review, not by the suite.

### Noted, pre-existing, not caused by this PR

`Score_BetterMatchWins` (`Tests~/Runtime/VoxrCommandParserTests.cs:359-367`) claims to pin literal-count tie-breaking, but on "cease fire" the `["disengage"]` pattern takes a required-literal miss and never becomes a candidate, so the test passes trivially and asserts nothing about tie-breaks. This is why no existing test caught the literal-count demotion in F1.

- **On-device (Quest) behaviour** — mic capture and real buffer timing in a headset. The PR's own deferral covers this; note F6 widens what should be checked there, and adds `prefixHoldSeconds` as a second precondition the note omits.
- **Re-running the Unity suite.** This is a bare UPM package; `Tests~` needs the host project to run. The PR's "EditMode 97/97, PlayMode 266/266" was **not** re-executed by this review. `CLAIM` did corroborate it statically: `Tests~/Editor` holds exactly 97 NUnit cases (89 `[Test]` + 8 `[TestCase]`) and `Tests~/Runtime` exactly 266 (254 + 10 + 2 `[UnityTest]`) at this head, and all five claimed new tests are present.

### Disposition

Per the issue lane, nothing here is a blocker, so **nothing in this review has been fixed** — all six findings are posted unruled for the human to decide. Not approved, not merged.

jinwoo1601:
Pushed `e70c938` — addresses **all six** findings from the review above. The human ruled on every finding rather than only the blockers (there were none), so this goes further than the issue lane's default.

| Finding | How it is addressed |
|---|---|
| **F1** — "only coin flips" scope claim is false | Rewritten in `CHANGELOG.md` and the PR body. Both now state that the span term sits above literal count and therefore also settles equal-score pairs literal count used to decide on its own, with the two concrete examples (`fire at {target}` over `fire at hotel`; `go {place}` over `go {dir}`) and an explicit call to check existing grammars. |
| **F2** — no coverage for span-vs-literal disagreement | `SpanTieBreak_LongerSpanBeatsHigherLiteralCount` |
| **F3** — trailing `[unk]` inflates the span | **Fixed in code.** `MatchResult` gains `ConsumedEndIdx` — the index after the last element that actually matched — and `IsBetterCandidate` compares that instead of `EndIdx`. `EndIdx` keeps its existing meaning for `searchStart` and the eager whole-buffer gate, so the fix is confined to the tie-break. Covered by `SpanTieBreak_TrailingUnkCannotWinTheTie` (the default-config `MatchedPatternIndex` flip) and `TryEagerCommit_TrailingUnkStillHoldsTheFullWindow` (the gate). |
| **F4** — score stated as 0.75 | Corrected to `(1+1+1-1)/4 = 0.5`, with the reason (a required-slot miss is a negative numerator term) rather than just the number. |
| **F5** — cross-command ties flip intent, untested | `SpanTieBreak_ChoosesBetweenCommands_NotJustPatterns`, using the `go {dir}` / `go {place}` case from the review. |
| **F6** — `None → HoldExtendable` untested, deferral note too narrow | No behaviour change: the shortened hold now fires on an honest whole-buffer match, where previously the same buffer got conservative treatment only as a side effect of the selection bug. That is `prefixHoldSeconds` doing its designed job, so it is pinned by `TryEagerCommit_SpanTieCanYieldHoldExtendable` and documented instead — the deferral note now names `prefixHoldSeconds` as the second precondition and the orphaned-tail risk the earlier wording did not reach. |

**Verification at `e70c938`:** full Unity suite via the `VoXR TestGround` host project (Unity 6000.4.7f1), both platforms — **EditMode 97/97, PlayMode 271/271, 0 failures**, no compile errors. That is the previous 363-test suite plus exactly the 5 new tests, with **no regressions**: every previously-passing test still passes under the `ConsumedEndIdx` comparison.

Still not verifiable from this environment, and unchanged by this push: on-device (Quest) confirmation of the two eager-flush timing transitions, which needs `eagerFlushOnCompleteMatch` together with a non-zero `prefixHoldSeconds`.

Not approving and not merging — this remains yours to rule on.

PR #50, opened 2026-07-30
feat-wav-replay: editor WAV replay seam + TTS fixture corpus (Tier B)

jinwoo1601:
## What this is

Tier B of the locked automated-verification design (feature backlog row 1): an editor WAV-replay seam in `EditorMicBackend`, a committed TTS fixture corpus with a reproducible generation script, and a PlayMode acoustic regression suite that replays every fixture through the real pipeline (Downsampler → AGC → int16 → libvosk.dll → command parser) and asserts intent/slots/transcript.

Process docs (disk-only, `Planning~/` is gitignored): `Planning~/features/wav-replay/{requirements,architecture}.md` — requirements F1–F10 trace to design §6; the architecture doc carries the decision register, build plan, and as-built deltas.

## Commits (build-plan phases)

- **Phase 0** — `VoxrWavReader` (byte-level RIFF/PCM gate: 48 kHz mono 16-bit only, distinct message-bearing rejections) + `VoxrAudioTestCase`/`VoxrAudioTestSuiteAsset` containers.
- **Phase 1** — pure code-motion extraction of `ProcessChunk` from `Tick()` (the design's own testability addition).
- **Phase 2** — playback mode: `StartPlayback`/`TickPlayback`/`StopPlayback` on a caller-driven pump; full pipeline-state reset per replay; mic/playback mutual exclusion; two internal editor-only test accessors on `VoxrSpeechRecogniser`. Zero added work on the live mic path (separate pump, no branch).
- **Phase 3** — 16-fixture corpus (all 11 demo intents, slot/alias/number variants, homophone pair, filler, split command, silence negative) at Quest-realistic amplitude (peaks 0.04–0.4); `Tests~/Fixtures/generate.sh` + manifest regenerate it end-to-end (piper → soxr → normalize); EditMode corpus verification tests.
- **Phase 4** — manifest-driven `Replay_ManifestCase` suite + the Phase 3a acoustic go/no-go probe.

## Verification

Host project (VoXR TestGround, Unity 6000.4.7f1), both platforms, real model + `libvosk.dll`:
- **EditMode 115/115** (13 WAV-reader + 5 corpus tests new)
- **PlayMode 286/286** on the committed tip (8 seam smoke + 16 replay cases + probe new; 261 pre-existing untouched) — evidence preserved at `VoXR TestGround/TestResults/{EditMode,PlayMode}-final.xml`.
- **Regeneration proof (F7):** `generate.sh` run end-to-end from a clean cache; suite against the regenerated corpus: 285/286 at intent/slot level — see finding 2 below. Evidence: `TestResults/{EditMode,PlayMode}-regen.xml`.
- PlayMode executes in ~25 s warm.

One-time host fix: TestGround was missing `StreamingAssets/vosk-model-small-en-us-0.15.zip` (the pre-existing suite never loaded a real model); installed per `Documentation~/getting-started.md`.

## Findings worth your attention

1. **Homophone loss (pinned as negative baseline):** grammar-constrained Vosk mis-hears **"switch to navigation"** as **"switch two navigation"** (to→two; both in the demo vocabulary) and no command matches — while "switch to weapons" survives the same trap. The fixture is pinned as a negative homophone baseline (expects no command; fires if recognition ever improves) and a new `navigation mode` fixture keeps the intent positively covered. Candidate follow-up issue: homophone-tolerant matching in the parser (out of scope here — no shipped-runtime changes beyond the seam).
2. **TTS regeneration is intent-stable, not transcript-stable:** a regenerated `fall back from target bravo two` decoded as "**pull** back from target bravo two" — the *same intent* (`retreat_from_target`, score 1.00, correct slot) via a synonymous pattern; only the strict raw-transcript assertion failed. Requirements F7 is amended (flagged, not silent) to an intent/slot-level bar for regenerated corpora: raw-transcript assertions are a property of the committed baseline and re-baseline on regeneration. The committed baseline passes 286/286 with transcripts asserted.

## G2 ratification bundle (design §6.1 wording readings)

Three deliberate readings of §6.1's carried-over v3.4 wording, argued in the architecture doc's decision register (D-2, D-6, D-7) — please ratify or push back:
1. **Separate `TickPlayback` pump** instead of a mode-branch inside `Tick()` ("Tick() reads successive chunks" read as the tick-cadence contract, giving literally zero live-path overhead and free faster-than-real-time replay).
2. **Fixed 4800-sample (100 ms) chunk** as "the same chunk size per tick" (determinism over frame-jitter simulation).
3. **`Voxr` prefix** for the container types (`VoxrAudioTestCase`/`VoxrAudioTestSuiteAsset`) — §6.1 prints v3.4's `Vosk*` names, but as provenance; the codebase is fully rebranded.

Plus the F7 acceptance amendment from finding 2 above.

Product docs (`Documentation~`, `CHANGELOG.md`, `KNOWN_LIMITATIONS.md` homophone entry) follow on this branch after G2, per the workflow bindings.


jinwoo1601 (commented):
## Independent review — full profile (pre-G2 evidence)

**Head reviewed:** `ab4e6c287c45109d2a2a7efdbcad8dc8b543b190` (merge-base `0adcbec`). **Profile:** full — `CLAIM` + all 11 roster angles, no angles dropped. Every finding below was adversarially verified by an independent verifier agent; refuted findings are listed at the end. Reviewing agents inherited nothing from the sessions that wrote the code.

**Overall:** the seam itself is solid — the `ProcessChunk` extraction was verified as character-level pure code motion, the zero-live-path-overhead invariant holds byte-for-byte, the DemoGrammar copy and all 16 committed WAVs verified faithful (peaks to 1 LSB), and the decimator was proven chunking-transparent by simulation. The findings concentrate on **what the suite proves vs. what the docs claim it proves**, one API-surface disclosure gap, and small robustness/hygiene items. Nothing here is a correctness defect in the shipped runtime path.

---

### 1. MAJOR — Replay time-base is decoupled from the command layer's wall-clock windows; the split fixture's pinned baseline exists only under time compression

`TickPlayback` advances audio 100 ms per call and the suite pumps once per frame (`Tests~/Runtime/VoxrWavReplayTests.cs:101-105`), while every command-layer timing decision runs on `Time.time` (`VoxrCommandRecogniser.cs:438,483`). Verified consequences:

- The split fixture's two Vosk finals are separated by **1.84 s of audio** (measured from the WAV envelope: seg1 ends 1.68 s, seg2 ends 3.52 s) but only ~43 ms of `Time.time` at the measured pump rate. Against the suite's own `BufferWindow = 1.5f`, **replaying the identical audio at real-time pacing would flush the buffer mid-gap and fail** — the pinned "current-correct" split behavior (requirements OQ6) is reachable only under compression.
- The manifest/design claim that this fixture "exercises buffer-window behavior" (§6.2 category) is not delivered: the case passes for any window ≳0.05 s. Multi-final concatenation *is* exercised — but that semantic is already covered deterministically by `VoxrCommandRecogniserInjectionTests.EagerFlush_SplitCommand_FiresWhenSecondHalfCompletes` at the identical knobs. The gap-vs-window relationship is what remains untested.
- Architecture D-6's "machine- and frame-rate-independent" holds up to the Vosk feed but not for the assertions: the real failure threshold is sustained ≲12 FPS (not single hitches — `Time.maximumDeltaTime` clamps those; a failure needs ~5 consecutive maximally-clamped frames).
- The suite's non-default knobs (`BufferWindow=1.5`, `CommandCooldown=0`, `EagerFlush=true`) match the pre-existing injection-test harness convention — deliberate parity, not drift — but no comment or doc states that rationale.

**Suggested disposition (human's call):** either re-scope the split fixture's claim to "multi-final concatenation" (doc + manifest description edit) and record the time-base limitation as a known blind spot of Tier B, or drive the command layer from an injectable clock during replay (larger change, arguably design-level). The current suite is green and stable at any realistic frame rate; the issue is what the docs say it proves.

### 2. MINOR — The determinism reset in `StartPlayback` is verified deletable with the suite staying green

`EditorMicBackend.cs:477-480` (downsampler/AGC/partial-cache/recognizer reset) is provably inert in the current suite: every PlayMode test builds a fresh backend per case, and the only same-backend re-arm test asserts chunk counts decided before the reset runs. The invariant that makes the corpus a regression detector (D-3, "every replay starts identical") is implemented but untested — it breaks silently if per-case model loads are ever consolidated into a shared setup. *Fix shape:* one test replaying fixture → contaminating audio → same fixture on one backend, asserting identical outcomes.

### 3. MINOR — No test asserts `OnPartialResult`/`OnResult` during replay

Architecture's replay flow says the test "asserts the captured `OnFinalResult`/`OnResult`/`OnCommandRecognised` payloads" and the G2 checklist says "fires the full event chain"; the suite subscribes only to `OnFinalResult`/`OnCommandRecognised`/`OnUnrecognisedSpeech`. `ProcessChunk`'s entire partial branch (partial results + dedup cache) executes on every replay asserted by nothing — which is also what makes finding 2's `_lastPartialLength` reset unobservable. Real-Vosk word payloads (`OnResult`) are parsed on every replay and never checked.

### 4. MINOR — `StartPlayback` passes a possibly-NULL recognizer to libvosk

`SetGrammar` frees and nulls `_recognizer`; on a failed `vosk_recognizer_new_grm` (reachable via the **public** `VoxrSpeechRecogniser.SetGrammar` with caller-supplied grammar JSON) it returns with `IsInitialised` still true. `StartPlayback` then arms and calls `vosk_recognizer_reset(IntPtr.Zero)` — and the next `TickPlayback` would push NULL into `accept_waveform_s`. The same PR guards this pointer in `TickPlayback`'s flush but not here; base `Start()`/`Reset()` carry the identical pre-existing gap (replicated hazard, not novel; editor-only). *Fix shape:* a `_recognizer != IntPtr.Zero` refusal in `StartPlayback` (optionally also the pre-existing sites).

### 5. MINOR — Two new public types + `Create > VoXR > Audio Test Suite` menu ship in the runtime assembly, undisclosed in the ratification bundle

`VoxrAudioTestCase`/`VoxrAudioTestSuiteAsset` are public with `[CreateAssetMenu]`, consumed only by the two new test files. Not technically forced: the same PR's `VoxrWavReader` is internal via `InternalsVisibleTo`, and both `JsonUtility` (public *fields* suffice — the private nested `JsonWrapper` proves it in-repo) and `CreateInstance<T>` work on internal types. Canon points both ways — v3.4 prints `public class VoskAudioTestCase` and the house siblings are public, but requirements NFR says "no new public game-facing API" and a package consumer who uses the Create-menu entry gets an asset nothing shipped can consume (the reader is internal, the seam is editor-only, the tests are stripped by release.yml). **This belongs in the G2 ratification bundle alongside D-2/D-6/D-7** — rule public (document it) or internal (drop the menu entry).

### 6. MINOR — Undisclosed whole-file reflow of `Runtime/VoxrSpeechRecogniser.cs`

112+/43− across 20 hunks; exactly one hunk is the declared change (two internal accessors + a `using` move). Verified semantically inert by whitespace-stripped comparison — but no formatter config exists in the repo, the sibling `EditorMicBackend.cs` was *not* reformatted (the two files now carry incompatible wrap styles), and neither the commit message nor the PR body discloses it. Disclosure/diff-hygiene: a reviewer certifying a shipped runtime file must verify 19 hunks of noise.

### 7. MINOR (clarity) — "committed baseline passes 286/286 with transcripts asserted" is 12-of-14

The transcript assertion is conditional on non-empty `expectedTranscript`, and the two hardest command-bearing fixtures — filler and split — ship with it empty, already sitting at the intent/slot bar the F7 amendment proposes for regenerated corpora only. The sentence appears in both the PR body and the as-built delta being ratified. Whether those two were left transcript-free by design or calibration convenience should be stated at G2.

### 8. MINOR — Four of the WAV reader's rejection branches are unreachable by its test builder

Build-plan claim: "crafted-byte fixtures for **every** rejection branch." `BuildWav` cannot emit a negative chunk size, a short `fmt `, `data`-before-`fmt `, or a `fmt `-less file — those four throws are untested (the other nine are covered). The same doc's Components section states the narrow truth; the plan sentence was never amended.

### 9. NIT — WAV reader truncation guard integer-overflows on hostile sizes

`dataStart + chunkSize > wav.Length` wraps negative for sizes near `int.MaxValue`; verified by compiled repro (20- and 44-byte crafted files → `IndexOutOfRangeException`/OOM instead of the contract's `InvalidDataException`). Fix verified: `chunkSize > wav.Length - dataStart` cannot overflow. Internal type, committed fixtures only — soundness, not operational risk.

### 10. NIT — Silent zero-case mode in the PlayMode suite (verifier-discovered)

`VoxrAudioTestSuiteAsset.FromJson` tolerates a zero-case parse; `[ValueSource]` would then yield no `Replay_ManifestCase` tests and the run stays green. EditMode guards this (`Manifest_ParsesAndIsNonEmpty`); PlayMode has no equivalent and nothing pins the expected case count.

### 11. NIT — Regeneration-story hygiene (three items)

(a) `generate.sh` pins piper-tts and soxr but **not numpy** — while the architecture doc claims all three pinned (doc/code disagreement). (b) The F7 re-baseline policy (regeneration invalidates transcript assertions) exists only in gitignored `Planning~/` — neither `generate.sh`'s header nor the assertion site mentions it; the next regenerator gets a red suite with no in-repo explanation. (c) The split fixture's manifest description still reads "Expectation provisional until Phase 4 calibration" — Phase 4 is done, and the description echoes into every failure message.

### 12. NIT — Mechanical cleanups (all verified behavior-preserving)

Dead `finals`/`unrecognised` parameters on `Replay(...)` (found by six independent angles); the category-coverage `ToList()` hoistable out of its loop (the conversion itself is forced by NUnit's non-generic `ICollection` overload); `generate.py`'s `"segments" in case and case["segments"]` → `case.get("segments")`; `TickPlayback`'s end-flush is byte-identical to `Stop()`'s flush — a shared helper is extraction-safe, and if extracted, disarming *before* flushing would also close a latent re-entrancy double-flush. Also: the re-arm test hardcodes `5` where `PlaybackChunkSamples` is derivable; the manifest's top-level `"voice"` field is read by nothing (the authoritative copy is `generate.sh`'s variable — either wire it or label it provenance); `ToJson` on the audio suite asset is public, caller-free, test-free, and lossy against the manifest (a hypothetical future export would silently produce a manifest that regenerates all fixtures as silence).

---

### Refuted during verification (considered and killed)

- **Per-entry peak drift** — no doc requires per-entry peak verification; `peak` is a generation input and content assertions bind audio↔manifest more tightly. REFUTED.
- **Untested exclusion guards as claim-contradiction** — no canon claims all four guard directions tested; the mic-running direction is structurally untestable in a batchmode host (worth one doc line, not a test). REFUTED as stated.
- **`category` string as fragile policy switch** — every mislabel path fails loudly through one of three assertions; the double-duty is declared canon. REFUTED. (Residue: design §6.2 says "silence/**noise** negative"; the corpus ships silence-only.)
- **Divisibility premise unpinned** — the decimator was proven chunking-transparent by simulation; the "divisible by 3" comment is cosmetic, not load-bearing. Consequence refuted.
- **Test-helper duplication "silently" diverging** — divergence fails loudly with the path in the message; downgraded to the DRY nit in item 12.

### Verified as claimed (positive evidence)

Phase→commit mapping exact; Phase 1 literally pure code motion (character-level); zero added work on the live mic tick (no branch, no field read, no allocation); `DemoGrammar` token-identical to `CommandDemo.cs` today; all 16 WAVs 48 kHz/mono/16-bit with peaks matching the manifest to 1 LSB; manifest↔disk 1:1; silence fixture true zeros; EditMode 115/115 and PlayMode 286/286 XMLs fresh, on this head, with the claimed new-test counts; the regen run's single failure exactly as described (intent-stable transcript drift); no premature G2 checkbox ticks anywhere.

Unity test execution itself (result XMLs were read as evidence, not re-run); actual piper regeneration (network + toolchain bootstrap); on-device behavior (out of Tier B's scope by design); `vosk_recognizer_reset(NULL)` behavior inside prebuilt libvosk (the API header documents no NULL tolerance — treated as contract-violating).

---

*Review produced by the project's review-pr workflow: 12 finder angles fanned out against a pinned brief, findings deduplicated, every survivor adversarially verified by an independent agent (3 findings killed in verification). G2 disposition of all findings belongs to the human; nothing has been fixed on the branch.*

jinwoo1601:
G2 ruled in conversation 2026-07-30: **accepted as-is** (including the D-2/D-6/D-7 §6.1 readings, the F7 intent/slot-level amendment for regenerated corpora, and the public container types). Review findings above remain on record unfixed per the as-is ruling — candidates for future issue-lane work. Product docs pushed as `f322dc8` (CHANGELOG, KNOWN_LIMITATIONS homophone extension, API reference for the audio containers, editor-testing guide). PR is ready for the human merge.

jinwoo1601:
Post-merge integration verification on `f796cca` (after merging main's #41 span tie-break): **EditMode 115/115, PlayMode 296/296** — all 16 acoustic replay fixtures individually confirmed Passed, so #41 changed no acoustic outcome; our suite and #48's new parser tests coexist green. Evidence: `VoXR TestGround/TestResults/{EditMode,PlayMode}-merged.xml`. PR is conflict-free and ready for the human merge.

PR #52, opened 2026-07-30
Add PlayMode coverage for VoxrPushToTalkController.OnApplicationPause (#49)

### LINKED ISSUE #49: Add PlayMode coverage for VoxrPushToTalkController.OnApplicationPause
Tier A of the locked automated-verification design (Planning~/design-docs/automated-verification.md, local process doc — routed through the issue lane per its DR-1).

`OnApplicationPause` (Runtime/VoxrPushToTalkController.cs:133-147) is the only lifecycle hook with no test coverage. Add PlayMode tests in `Tests~/Runtime/` driving the controller with `SendMessage("OnApplicationPause", true/false)`:

- Pause while recognising (continuous mode) → recognition stops.
- Resume with `_wantRecognising` set → recognition restarts.
- Resume when already recognising → no double-start.
- Pause while not recognising (PTT idle) → no stop call, no state corruption.
- Resume ordering doesn't fight the `Update()` permission-race reconciliation (`:149-155`).

**Mechanism (decide in this issue):** `VoxrSpeechRecogniser`'s `IsRecognising`/`StartRecognition`/`StopRecognition` are non-virtual and the controller's internal setters (`:157-158`) are typed to the concrete class, so no seam exists for a test double today. Either
(a) exercise a real initialised recogniser and assert `IsRecognising` transitions — on the Windows test host this routes through `EditorMicBackend` and needs a model plus a working `Microphone` device in batchmode; or
(b) add a minimal internal seam (internal virtuals or an internal control interface) so the controller is testable in isolation.
Test-only preferred; a minimal internal seam is acceptable.

**Scope note:** this verifies the C# state machine only. Whether Android actually delivers the pause callback on HMD doff is covered by the Tier D device rig (separate feature, not this issue).

jinwoo1601:
Closes #49

Tier A of the locked automated-verification design (`Planning~/design-docs/automated-verification.md` §5, routed to the issue lane by its DR-1). Tiers B (#50) and C (#51) are already merged; A was scheduled first and slipped.

## Mechanism decision

The issue left the observation mechanism open: **(a)** drive a real initialised recogniser and assert `IsRecognising` transitions, or **(b)** add a minimal internal seam. **Decided (b)**, for a reason sharper than "needs a mic in batchmode":

- **Two of the five required cases are call-count assertions, which no `IsRecognising` reading can express.** "Resume when already recognising → no double-start" and "pause while idle → no stop call" are about calls the controller must *not* make. `IsRecognising` is a bool: after a correct resume and after a double-start it reads `true` either way. Mechanism (a) cannot distinguish them, even with a perfectly working recogniser and a real microphone.
- `StartRecognition()` is fire-and-forget (`_ = StartRecognitionAsync()`) and awaits `InitialiseAsync()` when uninitialised, so under (a) a pause/resume assertion would race model extraction. Tier B's `StartPlayback` seam does not rescue this: playback is armed directly on `EditorMicBackend` and bypasses `StartRecognition()` entirely, so it cannot make the controller's own start path observable.

## What

**Seam — `Runtime/VoxrSpeechRecogniser.cs` (+16/−3).** `IsRecognising`, `StartRecognition()` and `StopRecognition()` now delegate to three new `internal virtual` members `IsRecognisingCore` / `StartRecognitionCore()` / `StopRecognitionCore()`; the existing bodies moved into them unchanged.

- Public signatures and runtime behaviour are byte-identical — pure indirection. No new public API, and no new public extension point: the overridable surface is `internal`, reachable only through the `InternalsVisibleTo` grants already in `Runtime/AssemblyInfo.cs`. The class was already unsealed (`public class VoxrSpeechRecogniser`), so nothing was unsealed for this.
- **The controller is not touched.** The state machine under test ships exactly as tested — that is why the seam went on the collaborator rather than on `VoxrPushToTalkController` (extract-and-override there would have changed the code under test) and rather than on an internal control interface (a second, interface-typed field alongside `[SerializeField] VoxrSpeechRecogniser` can desync, and Unity's fake-null semantics make `== null` unreliable through an interface reference to a destroyed component).

**Tests — `Tests~/Runtime/VoxrPushToTalkPauseTests.cs` (new, 8 PlayMode tests).** A nested `FakeSpeechRecogniser : VoxrSpeechRecogniser` overrides the three cores to count `Start`/`Stop` calls and to stage what `IsRecognising` reports. No override calls `base`, so no test touches a model, the native bridge, or a microphone; the suite is deterministic and platform-agnostic. `OnApplicationPause` is driven via `SendMessage("OnApplicationPause", true/false)` per the issue, on a GameObject holding only the controller.

The five cases from the issue map to tests as follows:

| Issue case | Test |
|---|---|
| Pause while recognising (continuous) → stops | `Pause_WhileRecognisingInContinuousMode_StopsRecognition` |
| Resume with `_wantRecognising` set → restarts | `Resume_WithWantRecognisingSet_RestartsRecognition` |
| Resume when already recognising → no double-start | `Resume_WhenAlreadyRecognising_DoesNotDoubleStart` |
| Pause while not recognising (PTT idle) → no stop, no corruption | `Pause_WhilePushToTalkIdle_DoesNotStopAndPreservesState` |
| Resume ordering vs the `Update()` reconciliation (`:149-155`) | `Resume_ThenUpdate_ReconcilerLeavesTheResumedSessionRunning` **and** `PauseResume_WhileIdleWithRecogniserRunning_ReconcilerStopsItNextFrame` |

Plus two that fell out of the same fixture: `Pause_WithNullRecogniser_IsNoOp` (the `:135` guard) and `PauseResume_WhilePushToTalkHeld_StopsThenRestarts` (the realistic doff-while-holding path, and the other way `_wantRecognising` becomes true).

Two notes on how the assertions are made:

- The reconciliation case is covered in **both** directions, because "doesn't fight the reconciler" has two failure modes: the reconciler must not stop a session the resume just restarted (`_wantRecognising == true`), and it must still stop an orphaned session after an idle pause/resume (`_wantRecognising == false`, recogniser running — the staged permission-race state). Both are `[UnityTest]` so real frames elapse and `Update()` actually runs.
- `_wantRecognising` is private, so "no state corruption" is proved behaviourally: after an idle pause/resume, a subsequent `PressTalk()` must still start recognition — the same indirect-probe idiom the existing `OnDisable_InContinuousMode_PreservesWantRecognisingFlag` test uses.

The lifecycle-silence contract documented in `Documentation~/push-to-talk.md:79` (resume must not re-fire `OnTalkStarted`/`OnTalkEnded`) is now asserted rather than assumed.

## Validation

- [x] **PlayMode: 304 tests, 0 failures, 0 skipped** (Unity exit 0) — host project `VoXR TestGround`, Unity 6000.4.7f1, batchmode.
- [x] **EditMode: 116 tests, 0 failures, 0 skipped** (Unity exit 0) — same host project.
- [x] All 8 new tests confirmed **present and passing** in the PlayMode result XML — the `VoxrPushToTalkPauseTests` fixture reports `total=8 passed=8 failed=0`. Checked by name, because a fixture that failed to compile would be *absent* from the XML rather than red.
- [x] Baseline before this change was 116 EditMode + 296 PlayMode; the delta is 296 → 304 PlayMode and EditMode unchanged — the 8 new tests and nothing else.
- [x] Two compile-time risks in this approach were unknowns before the run and are now settled empirically, not assumed: `internal override` **does** bind across the `InternalsVisibleTo` boundary under Unity's Roslyn (no CS0507/CS0115/CS0122; the test assembly rebuilt and ran), and `AddComponent<T>()` **does** work for a private *nested* MonoBehaviour subclass in a file whose name doesn't match the class (Unity's filename rule constrains serialization-based loading, not generic `AddComponent`).
- [ ] **On-device (Quest) check not needed for this PR, and deliberately not claimed.** Whether Android actually *delivers* `OnApplicationPause` on HMD doff or system-menu entry is Tier D's lifecycle test (design §8.3.4), explicitly out of scope per the issue's scope note. This PR verifies the C# state machine only. That split is the point of landing A before D: a later Tier D failure isolates to callback *delivery* rather than state-machine *logic*.

No `NativeBridge~/` change, so no bridge rebuild and no `.so` to commit.

**No `CHANGELOG.md` entry**, deliberately: public API, public behaviour and the shipped package are unchanged (`Tests~/` is stripped at release, and the seam is `internal`), so there is nothing user-facing to record. Flagging it so the omission reads as a decision rather than an oversight.

## Follow-ups noticed, not fixed here

- `Documentation~/push-to-talk.md:75` says the permission-race guard "cannot fire in the editor test runner — without the native DLL, `IsRecognising` is always false … Verification is manual, on a Quest device". That was already imprecise before this PR (the Windows editor backend can report `IsRecognising == true`), and this PR now covers the guard's *logic* in the editor — while the real dialog race still needs a device, since the permission coroutine is Android-only. Left alone to keep this diff surgical; worth a one-line issue if you want the doc tightened.
- The seam is deliberately scoped to the three members Tier A needs. If Tier D's rig wants the same interception for observability (e.g. asserting the controller's calls on-device), it extends this pattern rather than replacing it — but I have not designed for that here, and no Tier D requirement exists yet.

### DIFF
--- a/Runtime/VoxrSpeechRecogniser.cs
+++ b/Runtime/VoxrSpeechRecogniser.cs
@@ -87,7 +87,16 @@ namespace VoXR
             }
         }
 
-        public bool IsRecognising
+        public bool IsRecognising => IsRecognisingCore;
+
+        // Test seam (issue #49). VoxrPushToTalkController's lifecycle state machine
+        // reads recognition state and drives start/stop only through these three
+        // internal virtual members, so a test double can subclass this component and
+        // count the calls — "resume must not double-start" is a call-count assertion
+        // no IsRecognising reading can express. Internal keeps the public surface
+        // free of an extension point; overriding needs the InternalsVisibleTo grant
+        // in Runtime/AssemblyInfo.cs.
+        internal virtual bool IsRecognisingCore
         {
             get
             {
@@ -198,7 +207,9 @@ namespace VoXR
             IsModelReady = false;
         }
 
-        public void StartRecognition()
+        public void StartRecognition() => StartRecognitionCore();
+
+        internal virtual void StartRecognitionCore()
         {
             _ = StartRecognitionAsync();
         }
@@ -282,7 +293,9 @@ namespace VoXR
 #endif
         }
 
-        public void StopRecognition()
+        public void StopRecognition() => StopRecognitionCore();
+
+        internal virtual void StopRecognitionCore()
         {
             _isRecognising = false;
 #if UNITY_EDITOR_WIN
--- a/Tests~/Runtime/VoxrPushToTalkPauseTests.cs
+++ b/Tests~/Runtime/VoxrPushToTalkPauseTests.cs
@@ -1,0 +1,329 @@
+// ============================================================================
+// Purpose:  PlayMode tests for the VoxrPushToTalkController.OnApplicationPause state machine
+// Layer:    Tests.Runtime
+// Owns:     VoxrPushToTalkPauseTests (public class), FakeSpeechRecogniser (nested test double)
+// Depends:  VoxrPushToTalkController, VoxrSpeechRecogniser, VoxrListeningMode
+// ============================================================================
+using System.Collections;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.TestTools;
+
+namespace VoXR.Tests.Runtime
+{
+    public class VoxrPushToTalkPauseTests
+    {
+        // Subclasses the real component through the internal seam
+        // (VoxrSpeechRecogniser.IsRecognisingCore/StartRecognitionCore/StopRecognitionCore)
+        // so the controller cannot tell the difference, while the test counts calls
+        // and stages recognition state. No override calls base, so nothing here
+        // touches a model, the native bridge, or a microphone.
+        sealed class FakeSpeechRecogniser : VoxrSpeechRecogniser
+        {
+            internal int StartCalls { get; private set; }
+            internal int StopCalls { get; private set; }
+
+            // What IsRecognising reports. Settable so a test can stage a session the
+            // controller did not start itself.
+            internal bool Running { get; set; }
+
+            internal override bool IsRecognisingCore => Running;
+
+            internal override void StartRecognitionCore()
+            {
+                StartCalls++;
+                Running = true;
+            }
+
+            internal override void StopRecognitionCore()
+            {
+                StopCalls++;
+                Running = false;
+            }
+
+            internal void ResetCalls()
+            {
+                StartCalls = 0;
+                StopCalls = 0;
+            }
+        }
+
+        GameObject _recogniserGo;
+        GameObject _controllerGo;
+        FakeSpeechRecogniser _fake;
+        VoxrPushToTalkController _controller;
+
+        int _startedCount;
+        int _endedCount;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _recogniserGo = new GameObject("PauseTestFakeRecogniser");
+            _fake = _recogniserGo.AddComponent<FakeSpeechRecogniser>();
+
+            // The controller lives on its own GameObject so SendMessage("OnApplicationPause")
+            // reaches the state machine under test and nothing else.
+            _controllerGo = new GameObject("PauseTestController");
+            _controller = _controllerGo.AddComponent<VoxrPushToTalkController>();
+            _controller.SpeechRecogniser = _fake;
+            _controller.InitialiseOnStart = false;
+
+            _startedCount = 0;
+            _endedCount = 0;
+            _controller.OnTalkStarted.AddListener(() => _startedCount++);
+            _controller.OnTalkEnded.AddListener(() => _endedCount++);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            if (_controllerGo != null)
+                Object.DestroyImmediate(_controllerGo);
+            if (_recogniserGo != null)
+                Object.DestroyImmediate(_recogniserGo);
+        }
+
+        // -------- Drivers --------
+
+        // OnApplicationPause is private, so drive it the way the OS does.
+        void Pause() => _controllerGo.SendMessage("OnApplicationPause", true);
+
+        void Resume() => _controllerGo.SendMessage("OnApplicationPause", false);
+
+        // Continuous mode is how _wantRecognising gets set without a button press.
+        // The mode setter starts recognition, so call counts are cleared afterwards
+        // and every assertion below is absolute.
+        void ArrangeContinuousAndRunning()
+        {
+            _controller.ListeningMode = VoxrListeningMode.Continuous;
+            Assume.That(
+                _fake.Running,
+                Is.True,
+                "Precondition: switching to continuous mode should have started recognition"
+            );
+            _fake.ResetCalls();
+        }
+
+        // -------- Pause --------
+
+        [Test]
+        public void Pause_WhileRecognisingInContinuousMode_StopsRecognition()
+        {
+            ArrangeContinuousAndRunning();
+
+            Pause();
+
+            Assert.AreEqual(
+                1,
+                _fake.StopCalls,
+                "Pause while recognising must stop recognition exactly once"
+            );
+            Assert.IsFalse(_fake.Running);
+            Assert.AreEqual(0, _fake.StartCalls);
+            Assert.AreEqual(
+                0,
+                _endedCount,
+                "Pause is a lifecycle event — OnTalkEnded must not fire"
+            );
+        }
+
+        [Test]
+        public void Pause_WhilePushToTalkIdle_DoesNotStopAndPreservesState()
+        {
+            Assume.That(_controller.ListeningMode, Is.EqualTo(VoxrListeningMode.PushToTalk));
+            Assume.That(
+                _fake.Running,
+                Is.False,
+                "Precondition: an untouched push-to-talk controller must not be recognising"
+            );
+
+            Pause();
+
+            Assert.AreEqual(0, _fake.StopCalls, "Pause must not stop what was never started");
+            Assert.AreEqual(0, _fake.StartCalls);
+
+            Resume();
+
+            Assert.AreEqual(
+                0,
+                _fake.StartCalls,
+                "Resume must not start recognition the user never asked for"
+            );
+
+            // _wantRecognising is private; a press that still takes effect proves the
+            // pause/resume pair left it false rather than corrupting it.
+            _controller.PressTalk();
+
+            Assert.AreEqual(
+                1,
+                _fake.StartCalls,
+                "PressTalk after an idle pause/resume cycle must still start recognition"
+            );
+            Assert.AreEqual(1, _startedCount);
+        }
+
+        [Test]
+        public void Pause_WithNullRecogniser_IsNoOp()
+        {
+            _controller.SpeechRecogniser = null;
+
+            Assert.DoesNotThrow(() =>
+            {
+                Pause();
+                Resume();
+            });
+            Assert.AreEqual(0, _fake.StartCalls);
+            Assert.AreEqual(0, _fake.StopCalls);
+        }
+
+        // -------- Resume --------
+
+        [Test]
+        public void Resume_WithWantRecognisingSet_RestartsRecognition()
+        {
+            ArrangeContinuousAndRunning();
+            Pause();
+            Assume.That(
+                _fake.Running,
+                Is.False,
+                "Precondition: pause should have stopped recognition"
+            );
+            _fake.ResetCalls();
+
+            Resume();
+
+            Assert.AreEqual(
+                1,
+                _fake.StartCalls,
+                "Resume must restart recognition when the user still wants it"
+            );
+            Assert.IsTrue(_fake.Running);
+            Assert.AreEqual(
+                1,
+                _startedCount,
+                "Resume is a lifecycle event — OnTalkStarted must not re-fire "
+                    + "(the 1 is the continuous-mode switch)"
+            );
+        }
+
+        [Test]
+        public void Resume_WhenAlreadyRecognising_DoesNotDoubleStart()
+        {
+            // A resume that arrives while the session is still up — the guard at
+            // VoxrPushToTalkController.cs:144 is the only thing preventing a second
+            // native start, and only a call count can prove it held.
+            ArrangeContinuousAndRunning();
+
+            Resume();
+
+            Assert.AreEqual(
+                0,
+                _fake.StartCalls,
+                "Resume must not start a second session over a running one"
+            );
+            Assert.IsTrue(_fake.Running);
+            Assert.AreEqual(0, _fake.StopCalls);
+        }
+
+        [Test]
+        public void PauseResume_WhilePushToTalkHeld_StopsThenRestarts()
+        {
+            _controller.PressTalk();
+            Assume.That(
+                _fake.Running,
+                Is.True,
+                "Precondition: PressTalk should have started recognition"
+            );
+            _fake.ResetCalls();
+
+            Pause();
+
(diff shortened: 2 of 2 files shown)


jinwoo1601 (commented):
## Independent review — PR #52

**Depth profile:** slim (issue lane). Angles run: `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP` — the full slim roster, none dropped.
**Head reviewed:** `3230e316235d249f45fcf7a80ea2a122d0cb4481`, against merge-base `f87df96b09f87fe498cd906a33c550618b268d89`.
**Method:** six finder agents with no context from the session that wrote the code, then one adversarial verifier per surviving finding. 5 findings survived dedup; **4 CONFIRMED, 1 REFUTED. No blockers.**

Nothing here blocks merge. The five cases the linked issue requires are all genuinely asserted — `LINE` and `GAP` independently confirmed each one goes **red** if the production branch it targets is deleted, including both directions of the reconciler case. The findings below are about a sixth, extra test and about the accuracy of this PR's own stated reasons.

---

### 1. `Pause_WithNullRecogniser_IsNoOp` is vacuous, and the PR body misattributes it — minor, high confidence — **CONFIRMED**

`Tests~/Runtime/VoxrPushToTalkPauseTests.cs:167-178`. The body claims this test covers "the `:135` guard". It cannot. `SetUp` leaves `_listeningMode` at its `PushToTalk` default, so `Awake()` (`VoxrPushToTalkController.cs:111-112`) never sets `_wantRecognising` and the test never presses or switches mode. Both branches of `OnApplicationPause` are gated on that flag — `if (_wantRecognising)` at `:139`, and `if (_wantRecognising && !_speechRecogniser.IsRecognising)` at `:144`, where `&&` short-circuits before the dereference. Delete the `:135` guard and no `NullReferenceException` is reachable, so `Assert.DoesNotThrow` still passes.

The verifier strengthened this: the test survives deleting **the entire body** of `OnApplicationPause`, not merely the guard. The two `Assert.AreEqual(0, _fake.StartCalls/StopCalls)` lines are additionally tautological — `_controller.SpeechRecogniser = null` at `:169` detaches the fake before the act, so nothing can move either counter.

Consequence: the guard's only load-bearing state — `_wantRecognising == true` with a null or destroyed recogniser — is the one `paused × _wantRecognising × IsRecognising × null` combination of `:133-147` that no test in the fixture reaches. (`GAP` enumerated the rest and found them all covered.)

Smallest fix: arrange the state first — `ArrangeContinuousAndRunning(); _controller.SpeechRecogniser = null;` then pause — so the guard's removal actually produces the NRE; and drop the two tautological count assertions. Or keep the test and strike "(the `:135` guard)" from the body, since what it documents is the short-circuits, not the guard.

*Caveat, honestly flagged:* one limb of this finding — that Unity's `SendMessage` logs a receiver exception rather than propagating it, which would make `Assert.DoesNotThrow` decorative even in a state where the NRE does occur — came back **UNSETTLED**. No repo evidence either way and no receiver in the suite throws. It does not change the finding: limb one stands on control flow alone. But if the fix above is applied, that caveat decides whether the regression surfaces through the assertion or through the test framework's unexpected-log failure.

### 2. The test double is not inert by construction, and "platform-agnostic" is false — minor, high confidence (mechanism) — **CONFIRMED**

The body and the fixture's own header comment (`Tests~/Runtime/VoxrPushToTalkPauseTests.cs:19-20`) both say: "No override calls `base`, so no test touches a model, the native bridge, or a microphone; the suite is deterministic and platform-agnostic."

`FakeSpeechRecogniser` is a live `VoxrSpeechRecogniser` component, so it inherits the base class's privately-declared Unity messages — `void OnDestroy() { ReleaseNativeResources(); }` (`Runtime/VoxrSpeechRecogniser.cs:515`) — independently of what the three overrides do. `TearDown`'s `DestroyImmediate` therefore runs the real `ReleaseNativeResources()` after every test. On the Windows editor host that takes `#if UNITY_EDITOR_WIN` → `_editorBackend?.Release()` with a null backend and is genuinely benign. The `#else` branch is `BridgeNative.vosk_bridge_destroy()` — a real P/Invoke, with `_bridgeAvailable` true from its field initialiser (`:44`) and never lowered on this path.

So hermeticity is guaranteed by a **compile-time platform define**, not by the seam or by the overrides. The verifier's decisive point: the finding survives either resolution of whether Unity dispatches a private base `OnDestroy` to a derived instance — if it does, safety comes from `UNITY_EDITOR_WIN`; if it doesn't, from Unity's dispatch rules. Neither is "no override calls `base`". "Platform-agnostic" is false as written: on Android that line is a live native call, and on a macOS/Linux editor it throws `DllNotFoundException` into the `:200` catch.

Not purely hypothetical. This test assembly has empty `includePlatforms`/`excludePlatforms`, so it builds into an Android test player, and the locked design runs Tier D as `-testPlatform Android` over it (§8.2) with no stated filter. `vosk_bridge_destroy()` (`NativeBridge~/src/vosk_bridge.cpp:247-273`) stops capture, joins the recognition thread and frees the process-global recognizer/model — i.e. this fixture's TearDown could tear down a neighbouring fixture's initialised bridge. Two honest caveats: it early-returns unless `g_initialised`, so damage needs a neighbour holding an initialised bridge across this TearDown; and Tier D is undesigned, so a category filter would sidestep it. A plausible hazard, not a certainty.

Also note: `Tests~/Runtime/VoxrPushToTalkControllerTests.cs:24,43` already builds and destroys a **real** recogniser at base, so this PR adds an instance of an existing hazard, not a new class of one.

Smallest fix, **no production change needed**: either correct the wording in both places, or declare an empty `void OnDestroy() { }` on the fake — hiding the base's private message makes the double inert by construction rather than by platform define.

### 3. The fixture records a false reason for its own safety property — minor, high confidence — **CONFIRMED (as clarity)**

Same comment, second defect, and the root cause shared with finding 2: not calling `base` neutralises only the three overridden members. `Initialise()`, `InitialiseAsync()`, `StartRecognitionAsync()` and `ReleaseNativeResources()` are all public, all non-virtual, all inherited live by the fake. What actually keeps the suite off the model is `_controller.InitialiseOnStart = false` at `:70` — a load-bearing safety line carrying no comment saying so.

The verifier confirmed the payload is real: on this host the model is present and pre-extracted, so `ExtractModelAsync` returns synchronously and a reached `InitialiseAsync` would perform a real ~150 MB `vosk_model_new` load whose continuation outlives a synchronous `[Test]`. It also established the trigger is **not** this PR's doing — the pre-existing `VoxrPushToTalkControllerTests.cs:96` already walks that exact path on a real recogniser today with the suite green, and `StartRecognitionAsync` was already public and non-virtual at base. This PR inherits the class rather than introducing it.

It also **disproved the fix the finder proposed**: an `IsModelReady == false` tripwire in TearDown would pass vacuously (the flag is set after a thread hop, and `ReleaseNativeResources` forces it false anyway), and a recording double cannot intercept non-virtual calls made through the controller's `VoxrSpeechRecogniser`-typed field. So the correct remedy is the honest comment, not a tripwire: state that inertness rests on the three overrides **plus** `InitialiseOnStart = false`, and mark `:70` as a safety line.

### 4. `Assume` at `ArrangeContinuousAndRunning` can mask a real red — nit, medium confidence — **CONFIRMED, narrowed to one site**

`Tests~/Runtime/VoxrPushToTalkPauseTests.cs:100`. NUnit turns an unsatisfied `Assume` into *Inconclusive*, and this project's green criterion is literally `failed="0"` (`CLAUDE.md:40`); the result XML confirms `inconclusive` is a separate attribute.

The verifier found this matters at exactly one of the six `Assume` sites, for a specific reason: `:100` is the **only** assertion in the entire `Tests~` tree that continuous mode actually starts recognition. `StartRecognition()` and `_onTalkStarted?.Invoke()` are separate statements (`VoxrPushToTalkController.cs:59-60`), and the sibling test `ListeningMode_SetToContinuous_FiresOnTalkStarted` asserts only the event count — so deleting `:59` leaves it green. Under that regression `Resume_WhenAlreadyRecognising_DoesNotDoubleStart` *would* fail hard, but the helper throws `InconclusiveException` first and converts the red into an inconclusive that `failed="0"` reports as green.

The other five `Assume` sites are each backed by a hard assertion elsewhere and match the sibling file's established idiom — **do not sweep them**; that is churn the project conventions warn against. Recommended action is one word: `Assume.That` → `Assert.That` at `:100`. "No action, log as a follow-up" is also defensible, since the regression class requires a future production edit.

Minor precision note from the same verifier: this PR's validation line says "0 skipped", but `inconclusive` is a distinct NUnit attribute from `skipped` — the claim is true, and the XML did record `inconclusive="0"`, but that wording alone would not have caught inconclusives.

---

### Considered and killed

- **"The pure-indirection invariant rests on diff-reading, not on the suite" — REFUTED.** Two premises false. `Tests~/Runtime/VoxrPushToTalkControllerTests.cs:24` builds a real recogniser and reaches the moved bodies indirectly ~13 times via `PressTalk`/`ReleaseTalk`/`ListeningMode`, so `StartRecognitionCore` is not "executed by nothing"; and the PR body grounds the byte-identical claim in the diff, never in the test counts, so "lets the green run appear to carry it" was a strawman. One narrow nuance survives as a pre-existing observation, not a finding: those ~13 tests assert only `UnityEvent` counts, so a body gutted to a no-op would not redden them.

### Angles that found nothing, and what they establish

- **`GONE` — zero findings, mechanically proved.** It normalised the head file by rewriting only the three new declaration headers back to base form, then diffed against the base revision: **identical, zero hunks**. All three `#if`/`#else`/`#endif` blocks stay wholly inside their members, every `catch (DllNotFoundException)`/`MarkBridgeUnavailable()` path survives, all seven `_isRecognising` assignments keep their count/order/enclosing method, and all three load-bearing comments in the stop path survive verbatim. The diff removes exactly three lines, all declaration headers.
- **`XFILE` — zero findings.** Full caller census (`VoxrPushToTalkController`, `Commands/GrammarManager.cs:43,46,52`, the recogniser's own `Update():456`, both `Samples~` demos, `VoxrSpeechRecogniserLifecycleTests`) all bind unchanged public members; no reflective or `nameof` binding to these names exists anywhere. Override signatures, the `InternalsVisibleTo` ↔ asmdef name match, and the `"OnApplicationPause"` string against the private `bool`-arg method all check out — the last with in-repo precedent that Unity's message lookup reaches private methods (`Commands/VoxrCommandRecogniser.cs:841`). No `.meta` is needed: `Tests~` carries none in git.
- **`PIT`** additionally cleared, by derivation: `internal virtual`/`internal override` binds legally across the friend grant; `Awake`/`OnEnable` running inside `AddComponent` before the `[SetUp]` setters is safe given the `PushToTalk` default; Unity fake-null is not in play (the field is concretely typed and the null is a real null); one `yield return null` guarantees exactly one controller `Update()` and the reconciler's stop is idempotent, so `StopCalls == 1` is not frame-count-fragile; and the inherited `Update()` is inert on **both** preprocessor branches, unconditionally, because the fake never sets the base's private `_isRecognising`.

### Not verifiable from this environment — stated, not assumed

- **The Unity runs could not be re-executed by the review.** The claimed 116 EditMode / 304 PlayMode / 0 failures and the per-fixture `total=8 passed=8` come from the authoring session's `compile-check` run. `CLAIM` did reconcile the totals independently from source — 107 `[Test]` + 9 `[TestCase]` = 116 EditMode, and 289 declared Runtime cases with one `[ValueSource]` expanding 1→16 = 304 — and found no internal inconsistency.
- **Compilation itself** was judged by reading, not by invoking a compiler.
- **On-device (Quest) behaviour** — real `OnApplicationPause` delivery on HMD doff, mic capture, permission-dialog timing. No device attached; the PR explicitly defers this to Tier D and does not claim it. That deferral is consistent with the issue's own scope note.

---

Per the issue lane, only a CONFIRMED blocker would be fixed in this run. There are none, so **nothing was changed in response to this review** — all four findings are posted unfixed and unruled, for the human to decide. Findings 1 and 4 are small code edits; findings 2 and 3 are one corrected comment plus matching body wording, and share a root cause.

PR #54, opened 2026-07-30
Tidy the #49 pause-test fixture per PR #52 review (#53)

### LINKED ISSUE #53: Tidy the #49 pause-test fixture per PR #52 review (4 confirmed findings)
Follow-up to PR #52 (issue #49, Tier A of the automated-verification design). Four findings were CONFIRMED by that PR's review pass and merged unfixed, as accepted-as-is. All four sit in `Tests~/Runtime/VoxrPushToTalkPauseTests.cs` and total roughly 15 lines of change. Full evidence is in the review comment on PR #52; this issue is self-contained so it can be picked up cold.

**None of these affect shipped behaviour.** The production seam is sound (independently verified as byte-identical to base), the five state-machine cases issue #49 required are all genuinely asserted, and the suite is green. These are test-power and rationale-accuracy defects.

## 1. `Pause_WithNullRecogniser_IsNoOp` is vacuous (`:167-178`)

The test claims to cover the null guard at `Runtime/VoxrPushToTalkController.cs:135`. It cannot. `SetUp` leaves `_listeningMode` at its `PushToTalk` default, so `Awake()` (`:111-112`) never sets `_wantRecognising`, and the test neither presses nor switches mode. Both branches of `OnApplicationPause` are gated on that flag — `if (_wantRecognising)` at `:139`, and `if (_wantRecognising && !_speechRecogniser.IsRecognising)` at `:144` where `&&` short-circuits before the dereference. So no `NullReferenceException` is reachable and `Assert.DoesNotThrow` passes even with the guard deleted. The verifier established the stronger form: **the test survives deleting the entire body of `OnApplicationPause`.** The two `Assert.AreEqual(0, _fake.StartCalls/StopCalls)` lines are separately tautological, since `_controller.SpeechRecogniser = null` at `:169` detaches the fake before the act.

This leaves the `:135` guard's only load-bearing state — `_wantRecognising == true` with a null or destroyed recogniser — as the one `paused × _wantRecognising × IsRecognising × null` combination of `:133-147` that no test reaches. (The review enumerated the rest; all covered.)

Fix: arrange the state before nulling — `ArrangeContinuousAndRunning(); _controller.SpeechRecogniser = null;` then pause — so removing the guard actually produces the NRE, and drop the two tautological assertions.

**Caveat that affects how to write the fix:** one limb of this finding came back UNSETTLED — whether Unity's `SendMessage` logs a receiver exception rather than propagating it. If it does, `Assert.DoesNotThrow` is decorative even in a state where the NRE occurs, and the regression would surface as the test framework's unexpected-error-log failure instead. Settle that before relying on the assertion; invoking the private `OnApplicationPause` by reflection is the assertion-visible alternative.

## 2 & 3. The fixture states a false reason for its own safety (`:19-20`, `:70`)

One root cause, two consequences. The header comment says: *"No override calls `base`, so nothing here touches a model, the native bridge, or a microphone"* — and the (now-merged) PR body added *"the suite is deterministic and platform-agnostic"*. Both are wrong:

- `FakeSpeechRecogniser` is a live `VoxrSpeechRecogniser`, so it **inherits the base class's privately-declared Unity messages** regardless of the overrides. `TearDown`'s `DestroyImmediate` runs the real `OnDestroy() → ReleaseNativeResources()` (`Runtime/VoxrSpeechRecogniser.cs:515`) after every test. That is benign only on the `#if UNITY_EDITOR_WIN` branch (`_editorBackend?.Release()`, null backend). The `#else` branch is `BridgeNative.vosk_bridge_destroy()` — a real P/Invoke, with `_bridgeAvailable` true from its field initialiser (`:44`). So hermeticity comes from a **compile-time platform define**, not from the seam. "Platform-agnostic" is false: on Android that line is a live native call; on a macOS/Linux editor it throws `DllNotFoundException`.
- Not calling `base` neutralises only the three overridden members. `Initialise()`, `InitialiseAsync()`, `StartRecognitionAsync()` and `ReleaseNativeResources()` remain public, non-virtual and live on the fake. What actually keeps the suite off the model is `_controller.InitialiseOnStart = false` at `:70` — a load-bearing safety line with no comment saying so. (Payload confirmed real: on the test host the model is pre-extracted, so a reached `InitialiseAsync` would perform a real ~150 MB `vosk_model_new` load whose continuation outlives a synchronous `[Test]`.)

Fix: state the true reason in the comment — inertness rests on the three overrides **plus** `InitialiseOnStart = false` — and mark `:70` as a safety line, not cosmetics. Optionally make inertness structural by declaring an empty `void OnDestroy() { }` on the double, which hides the base's private message. No production change is required either way.

**Do not** add an `IsModelReady == false` tripwire: the review disproved it. The flag is set after a thread hop, and `ReleaseNativeResources` forces it false anyway, so the assertion passes vacuously. A recording double also cannot intercept non-virtual calls made through the controller's `VoxrSpeechRecogniser`-typed field.

## 4. One `Assume` can mask a real red (`:100`)

`ArrangeContinuousAndRunning()` gates four of the eight tests with `Assume.That(_fake.Running, Is.True, …)`. NUnit turns an unsatisfied `Assume` into *Inconclusive*, and this project's green criterion is `failed="0"` — `inconclusive` is a separate XML attribute.

This matters at exactly this one site, for a specific reason: `:100` is the **only** assertion in the whole `Tests~` tree that continuous mode actually starts recognition. `StartRecognition()` and `_onTalkStarted?.Invoke()` are separate statements (`Runtime/VoxrPushToTalkController.cs:59-60`), and the sibling test `ListeningMode_SetToContinuous_FiresOnTalkStarted` asserts only the event count — so deleting `:59` leaves it green. Under that regression `Resume_WhenAlreadyRecognising_DoesNotDoubleStart` *would* fail hard, but the helper throws `InconclusiveException` first and converts the red into a green-reporting inconclusive.

Fix: one word — `Assume.That` → `Assert.That` at `:100`. **Do not sweep the other five `Assume` sites** in this file; each is backed by a hard assertion elsewhere and matches the established idiom in the sibling `VoxrPushToTalkControllerTests.cs`. Changing them is churn with no coverage gain.

## Verification when this is picked up

Both host suites per the project's verification bindings — EditMode **and** PlayMode; these are PlayMode tests, so an EditMode-only run misses them entirely. Baseline to beat: 116 EditMode + 304 PlayMode, 0 failures. Watch the `inconclusive` attribute as well as `failed`, per finding 4.

jinwoo1601:
Closes #53

## What

Four test-power and rationale-accuracy defects in `Tests~/Runtime/VoxrPushToTalkPauseTests.cs`, confirmed by PR #52's review and merged unfixed. **No production code changes** — the diff touches exactly one test file.

1. **`Pause_WithNullRecogniser_IsNoOp` was vacuous.** `SetUp` leaves the controller in its `PushToTalk` default, so `_wantRecognising` is false and both branches of `OnApplicationPause` short-circuit before ever dereferencing the recogniser — the test passed with the entire method body deleted. It now calls `ArrangeContinuousAndRunning()` before nulling the reference, and drops the two assertions that were tautological once `SpeechRecogniser = null` detached the fake.

   The issue flagged one limb as UNSETTLED: whether `SendMessage` logs a receiver exception instead of propagating it, which would make `Assert.DoesNotThrow` decorative even in a throwing state. Rather than settle it, the test now sidesteps it — a new `PauseDirect(bool)` driver invokes the private message by reflection, so the throw is visible to the assertion either way (reflection's `TargetInvocationException` wrapper still fails `DoesNotThrow`). The other seven tests keep the `SendMessage` drivers.

2 & 3. **The fixture stated a false reason for its own safety.** "No override calls `base`, so nothing here touches a model, the native bridge, or a microphone" was wrong twice over: a base class's privately-declared Unity messages still run on a subclass, so `TearDown`'s `DestroyImmediate` reached the real `OnDestroy → ReleaseNativeResources()` on every test — benign only on the `#if UNITY_EDITOR_WIN` branch, a live `vosk_bridge_destroy()` P/Invoke off it. And not calling `base` neutralises only the three overridden cores; `Initialise()` and friends are public non-virtual, so what actually keeps the suite off the model is `InitialiseOnStart = false`.

   The header comment now names the real load-bearing set, `InitialiseOnStart = false` is annotated as a safety line, and an empty `void OnDestroy() { }` on the double makes inertness structural rather than resting on a compile-time platform define. No `IsModelReady` tripwire — the issue disproved it.

4. **One `Assume` could mask a real red.** `ArrangeContinuousAndRunning`'s `Assume.That` → `Assert.That`. It is the only check in the whole `Tests~` tree that continuous mode starts recognition at all, and an unsatisfied `Assume` reports *Inconclusive*, which this project's `failed="0"` green criterion reports as a pass. The other five `Assume` sites are left alone deliberately, per the issue.

## Why

All four are test-power defects, not behaviour defects. The production seam is sound and the five state-machine cases issue #49 required are genuinely asserted — but three of these four defects meant a future regression could land green.

## Validation

Unity 6000.4.7f1, host project `VoXR TestGround`, both platforms per the project's verification bindings:

- [x] **EditMode: 116/116 passed** — 0 failed, 0 inconclusive, 0 skipped
- [x] **PlayMode: 304/304 passed** — 0 failed, 0 inconclusive, 0 skipped
- [x] Combined 420/420, exact match with the PR #52 baseline (this change adds and removes no tests). `inconclusive` checked explicitly on both platforms, per finding 4 — it is a distinct NUnit attribute from `failed` and the reason finding 4 exists.
- [x] Zero `error CS` in both logs, so the totals are real runs rather than empty ones.

**Both fixes verified by mutation testing rather than by inspection** — the point of this issue is that these tests looked like they asserted things they did not, so "it still passes" is not evidence:

- [x] **Mutation A** — deleted the `if (_speechRecogniser == null) return;` guard at `VoxrPushToTalkController.cs:135`. Result: **exactly 1 failure**, `Pause_WithNullRecogniser_IsNoOp`, with an inner `NullReferenceException` at `:140` (the precise line the guard protected) wrapped in `TargetInvocationException`. The other 7 tests in the fixture stayed green, so the failure is causally tied to the mutation. Under the old version this test survived deleting the whole method body.
- [x] **Mutation B** — deleted `_speechRecogniser?.StartRecognition();` at `VoxrPushToTalkController.cs:59`. Result: **5 failed, 0 inconclusive**, all five with the message `Switching to continuous mode must start recognition`, all originating in `ArrangeContinuousAndRunning`. The `failed`/`inconclusive` split is the whole point: under `Assume` these would have been inconclusive and reported green.
- [x] Mutation B also **confirmed the coverage hole** finding 4 describes: the sibling `VoxrPushToTalkControllerTests.ListeningMode_SetToContinuous_FiresOnTalkStarted` **passed** under the same mutation (that suite stayed 16/16 green), because it asserts only the event count while `:59` and `:60` are separate statements. Recognition never starts and it still reports green.

Both mutations were reverted; the branch diff touches no production file (`git diff origin/main --name-only` outside `Tests~/` returns nothing).

### Not verified

- **The `OnDestroy()` shim's shadowing is not empirically proven.** Whether redeclaring a Unity message on a subclass suppresses the base's privately-declared one cannot be demonstrated from the only platform available here: on `UNITY_EDITOR_WIN` the base's `OnDestroy` is already benign (`_editorBackend?.Release()` on a null backend), so a green run is consistent with either outcome. The shim is correct-or-inert — it cannot make things worse — but its benefit lands on Android/macOS/Linux and is unverified there.
- **On-device (Quest) verification: not applicable, not claimed.** No production code changed, so shipped pause behaviour is untouched by this PR. The fixture remains a headless proxy for a real OS pause callback, exactly as it was before.

### DIFF
--- a/Tests~/Runtime/VoxrPushToTalkPauseTests.cs
+++ b/Tests~/Runtime/VoxrPushToTalkPauseTests.cs
@@ -5,6 +5,7 @@
 // Depends:  VoxrPushToTalkController, VoxrSpeechRecogniser, VoxrListeningMode
 // ============================================================================
 using System.Collections;
+using System.Reflection;
 using NUnit.Framework;
 using UnityEngine;
 using UnityEngine.TestTools;
@@ -16,8 +17,18 @@ namespace VoXR.Tests.Runtime
         // Subclasses the real component through the internal seam
         // (VoxrSpeechRecogniser.IsRecognisingCore/StartRecognitionCore/StopRecognitionCore)
         // so the controller cannot tell the difference, while the test counts calls
-        // and stages recognition state. No override calls base, so nothing here
-        // touches a model, the native bridge, or a microphone.
+        // and stages recognition state.
+        //
+        // Overriding the cores is not on its own what keeps this fixture off the model,
+        // the native bridge and the microphone. Inertness rests on all three of:
+        //   - no override calling base, so nothing the seam covers reaches native code;
+        //   - the empty OnDestroy() below, because a base class's privately-declared
+        //     Unity messages still run on a subclass and the real one releases native
+        //     resources (benign only under UNITY_EDITOR_WIN, a live
+        //     vosk_bridge_destroy() P/Invoke off it). The base Update() needs no such
+        //     shim — it returns early while the base's own _isRecognising stays false;
+        //   - SetUp's InitialiseOnStart = false, since Initialise()/InitialiseAsync()/
+        //     StartRecognitionAsync() are public non-virtual and the seam misses them.
         sealed class FakeSpeechRecogniser : VoxrSpeechRecogniser
         {
             internal int StartCalls { get; private set; }
@@ -46,6 +57,11 @@ namespace VoXR.Tests.Runtime
                 StartCalls = 0;
                 StopCalls = 0;
             }
+
+            // Keeps the base's private OnDestroy — and its ReleaseNativeResources() call
+            // (VoxrSpeechRecogniser.cs:515) — off this double. No `new` keyword: the base
+            // declaration is private, so nothing is being hidden as far as C# is concerned.
+            void OnDestroy() { }
         }
 
         GameObject _recogniserGo;
@@ -67,6 +83,8 @@ namespace VoXR.Tests.Runtime
             _controllerGo = new GameObject("PauseTestController");
             _controller = _controllerGo.AddComponent<VoxrPushToTalkController>();
             _controller.SpeechRecogniser = _fake;
+            // Load-bearing, not tidiness: Initialise() is public and non-virtual, so the
+            // seam does not cover it. Without this, Start() loads the real model.
             _controller.InitialiseOnStart = false;
 
             _startedCount = 0;
@@ -91,16 +109,32 @@ namespace VoXR.Tests.Runtime
 
         void Resume() => _controllerGo.SendMessage("OnApplicationPause", false);
 
+        // Same message, called directly. Only the null-guard test needs this: SendMessage
+        // may log a receiver exception rather than propagate it, which would leave
+        // Assert.DoesNotThrow unable to see the very throw it exists to rule out.
+        // Reflection's TargetInvocationException wrapper still fails DoesNotThrow.
+        void PauseDirect(bool paused) =>
+            typeof(VoxrPushToTalkController)
+                .GetMethod("OnApplicationPause", BindingFlags.Instance | BindingFlags.NonPublic)
+                .Invoke(_controller, new object[] { paused });
+
         // Continuous mode is how _wantRecognising gets set without a button press.
         // The mode setter starts recognition, so call counts are cleared afterwards
         // and every assertion below is absolute.
+        //
+        // Assert rather than Assume, unlike the other preconditions in this file: this is
+        // the only check anywhere in Tests~ that continuous mode starts recognition at all.
+        // VoxrPushToTalkController.cs:59 is a separate statement from the :60 event that
+        // the sibling ListeningMode_SetToContinuous_FiresOnTalkStarted covers, so deleting
+        // :59 is invisible everywhere else. An unsatisfied Assume reports Inconclusive,
+        // which this project's failed="0" green criterion would report as a pass.
         void ArrangeContinuousAndRunning()
         {
             _controller.ListeningMode = VoxrListeningMode.Continuous;
-            Assume.That(
+            Assert.That(
                 _fake.Running,
                 Is.True,
-                "Precondition: switching to continuous mode should have started recognition"
+                "Switching to continuous mode must start recognition"
             );
             _fake.ResetCalls();
         }
@@ -166,15 +200,20 @@ namespace VoXR.Tests.Runtime
         [Test]
         public void Pause_WithNullRecogniser_IsNoOp()
         {
+            // The guard at VoxrPushToTalkController.cs:135 only carries weight when the
+            // user still wants recognition: both branches of the state machine are gated
+            // on _wantRecognising (:139, :144), so from the PushToTalk-idle default this
+            // test would pass with the whole method body deleted. Arrange the wanting
+            // state first, then drop the reference — now deleting :135 dereferences null
+            // on the pause branch and again on the resume branch.
+            ArrangeContinuousAndRunning();
             _controller.SpeechRecogniser = null;
 
             Assert.DoesNotThrow(() =>
             {
-                Pause();
-                Resume();
+                PauseDirect(true);
+                PauseDirect(false);
             });
-            Assert.AreEqual(0, _fake.StartCalls);
-            Assert.AreEqual(0, _fake.StopCalls);
         }
 
         // -------- Resume --------


jinwoo1601 (commented):
# Independent review — PR #54

**Depth profile:** slim (issue lane, `issue/53-*`) — angles `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Head reviewed:** `12ebe89bdc3e7115e594fa324e596d11499c9afc` · **Base (merge-base):** `ce25164`
**Method:** six finder agents with no context from the session that wrote the code, then one adversarial verifier per surviving finding, each instructed to refute and to default to REFUTED when unable to substantiate.

**Outcome: 13 raw findings → 9 after dedup → 5 CONFIRMED (3 minor, 2 nit), 4 REFUTED. No blockers.**

Invariants confirmed independently: the diff touches exactly one file and no production code (`git diff ce25164..12ebe89 --name-only`); all eight tests survive with unchanged names and assertions; `Assume` count 6 → 5 with exactly one converted and no sweep; no `IsModelReady` tripwire; the null-guard test is genuinely non-vacuous (statically re-derived, independent of the reported test run).

---

## CONFIRMED

### 1. Header comment mis-attributes `StartRecognitionAsync()` — the attribution is inverted, not just misgrouped
**minor · high confidence · `Tests~/Runtime/VoxrPushToTalkPauseTests.cs:30-31`**

```
//   - SetUp's InitialiseOnStart = false, since Initialise()/InitialiseAsync()/
//     StartRecognitionAsync() are public non-virtual and the seam misses them.
```

`StartRecognitionAsync()` is reachable in this fixture **only** through `StartRecognitionCore()` (`Runtime/VoxrSpeechRecogniser.cs:214`), which is one of the three seam members the double overrides without `base`. The seam does not miss it — the seam is its only door. `grep -rn "StartRecognitionAsync" Runtime/ Editor/` returns exactly two hits: the declaration and that single call site.

Worse, the causation is backwards: `StartRecognitionAsync()` self-initialises —

```csharp
// Runtime/VoxrSpeechRecogniser.cs:217-223
public async Task StartRecognitionAsync()
{
    if (!_bridgeAvailable) return;
    if (!IsInitialised)
        await InitialiseAsync();     // :223
```

— so `InitialiseOnStart = false` gives that path **zero** protection. The bullet also contradicts the same comment's first bullet three lines above, which correctly claims the seam covers this path.

This is not inherited from issue #53. The issue's sentence is a surface inventory ("remain public, non-virtual and live on the fake") and is true as written; the PR re-derived the mapping — moving `ReleaseNativeResources()` into the `OnDestroy()` bullet and correctly naming only `Initialise()` at the SetUp site (`:86-87`) — and misfiled this one name in the process.

Why it lands at the top of `minor` rather than as a nit: this comment's factual accuracy *is* the PR's headline deliverable. Findings 2 & 3 of issue #53 exist because the previous comment stated a false reason for the fixture's own safety, and the replacement still contains a false clause. Harm path is narrow but real — a maintainer reworking the double would read this as saying `InitialiseOnStart = false` backstops the async start. It does not; `:223` would load the real model and `:236` would open the mic (or P/Invoke `vosk_bridge_start()` off `UNITY_EDITOR_WIN`).

**Fix:** attribute `Initialise()`/`InitialiseAsync()` to `InitialiseOnStart = false`, and either drop `StartRecognitionAsync()` from that bullet or move it to the first.

### 2. `InitialiseOnStart = false` is annotated as load-bearing, but its removal is *silent* — and this PR deepens what the silence hides
**minor · high confidence · `Tests~/Runtime/VoxrPushToTalkPauseTests.cs:86-88`**

The new comment is accurate but incomplete in the direction that matters. Deleting the line it annotates produces no failed assertion, no propagated exception, and no logged error for the test framework to fail on:

- `Runtime/VoxrPushToTalkController.cs:31,115-119` — `_initialiseOnStart` defaults `true`; `Start()` calls `Initialise()`.
- `Runtime/VoxrSpeechRecogniser.cs:121-124` — `public void Initialise() { _ = InitialiseAsync(); }`, fire-and-forget.
- Every terminal failure funnels to `FireError → OnError?.Invoke` (`:520-523`), which this fixture never subscribes.
- `grep -rn "LogError\|LogException\|LogAssertion" Runtime/` → **zero matches repo-wide**. (Only five `LogWarning` sites, which UTF does not fail on.)

Three escape hatches were opened and closed by the verifier: no `MissingReferenceException` past `TearDown` (post-await code touches only managed fields); the host's extracted model cache and `libvosk.dll` are both present, so `vosk_model_new` would *succeed* — a real ~150 MB native load, 1–3 s per test, invisible to NUnit; and `Start()` does run inside the two `[UnityTest]` cases, which `yield return null`.

The PR did not create this silence — the bare line exists at the merge-base and still sits uncommented in `VoxrPushToTalkControllerTests.cs:30`. What makes it a finding *against this PR* is the new shim: pre-PR, had `Initialise()` ever fired, `DestroyImmediate` → base `OnDestroy` → `ReleaseNativeResources()` → `_editorBackend?.Release()` bounded the damage. With `void OnDestroy() { }` in place that release path is gone, so the silent failure would now also leak the backend per test.

The only in-scope remedy is documentation — say that removal is silent. Making the seam cover `Initialise()` is a production change, and issue #53 states none is required.

### 3. Only one of the two states Unity's `==` conflates is covered
**minor · high confidence · `Tests~/Runtime/VoxrPushToTalkPauseTests.cs:210`**

`VoxrSpeechRecogniser : MonoBehaviour` (`:20`), so `if (_speechRecogniser == null)` at `VoxrPushToTalkController.cs:135` resolves through `UnityEngine.Object.operator==`, true for **two** states: a real CLR null, and a live wrapper whose native object was destroyed. The test arranges only the first. Issue #53 names the guard's load-bearing state as "`_wantRecognising == true` with a null **or destroyed** recogniser".

With the guard deleted and a *destroyed* fake still assigned, the double's overrides touch only managed state (`StopCalls++; Running = false;`), so both branches execute cleanly and `Assert.DoesNotThrow` **passes**. Mutation A is therefore evidence for one limb only. A `== null` → `is null`/`ReferenceEquals` refactor — what an IDE "simplify null check" suggests — would keep all 420 tests green.

Two corrections to the finding as originally raised, both from the verifier: the destroyed-limb harm is a *stale `vosk_bridge_stop()` native call*, not the clean `MissingReferenceException` first claimed (the only `MissingReferenceException` risk is the resume branch, Android-only, via `StartCoroutine` at `VoxrSpeechRecogniser.cs:231`); and invariant 2 is still *satisfied* — the test does die under Mutation A via the null limb. This is residual coverage, not a blocker.

Fixable in the test alone: destroy `_recogniserGo` leaving `SpeechRecogniser` pointed at `_fake`, pause, assert `_fake.StopCalls == 0`. Note that assertion is **not** tautological in this arrangement, because destroying does not detach the fake. The gap is between issue #53's *diagnosis* ("null or destroyed") and its *prescription* (null only), which the PR implemented verbatim.

### 4. Rationale comment claims a second observation point the test cannot reach
**nit · high confidence · `Tests~/Runtime/VoxrPushToTalkPauseTests.cs:207-208`**

> `// now deleting :135 dereferences null on the pause branch and again on the resume branch.`

Both `PauseDirect` calls sit in one `Assert.DoesNotThrow` lambda (`:212-216`), so the first throw aborts the delegate and `PauseDirect(false)` never runs — the resume dereference at `:144` is unreachable under the mutation the comment describes. The PR's own Mutation A evidence names `:140` only. With the guard present, `:135` returns before `:137` reads `paused`, so the resume call executes an identical path.

Fairness to the code: the resume call is not worthless — a guard *relocated* into the `if (paused)` block would be caught only by it. It just does not do what the comment says. Below-behavioural, but salient because "rationale-accuracy" is this PR's own stated theme.

### 5. Mutation A's prose says "deleted"; the evidence only coheres with an in-place replacement
**nit · PR description only, no shipped code affected**

Verdict returned as CONFIRMED-AS-CLARITY. The arithmetic: a literal `sed '135d'` puts `_speechRecogniser.StopRecognition();` at `:139`, so a stack trace could not report `:140`. The verifier could not determine which edit was made and correctly declined to resolve it.

**Author-side disambiguation, for the record:** the mutation was `sed -i '135s|.*|// MUTATION-A: guard removed|'` — a line-preserving in-place replacement. So `:140` is genuine and read off the log; the underlying evidence is sound and only the word "deleted" is imprecise. Mutation A's substantive claims (exactly 1 failure, the test name, the `TargetInvocationException`-wrapping-`NullReferenceException` pair) are unaffected.

---

## REFUTED — considered and killed

- **The two deleted assertions should have been kept** (raised independently by two angles). Refuted: under the proposed `_wantRecognising = false; return;` mutation the fake is *detached*, so both counts stay 0 and the assertions pass anyway — keeping them would not have backed the "IsNoOp" name. Relative to the SUT, no mutation of `OnApplicationPause` can make them fail; only a break in the test's own internal detach seam could. Tautological with respect to the code under test, which is what issue #53's word ranges over. *(A genuine residual was logged: the `_wantRecognising = false; return` mutation is caught by no test — pre-existing at `ce25164`, outside issue #53's four items, future-issue candidate.)*
- **The shipped comment asserts the `OnDestroy` shim works without the PR body's hedge.** Refuted: the shim's efficacy is not a separate unproven proposition but the same single Unity mechanism the finding already accepts as established — Unity resolves one method pointer per message name per type by a most-derived-first walk, which is *why* the base's private `OnDestroy` runs on a subclass at all. Holding the first as fact and the second as unproven is incoherent.
- **Unchecked `GetMethod` misattributes a future rename.** Refuted: the two failure modes differ on the first line of NUnit output (`TargetInvocationException` vs bare `NullReferenceException`), it can never fail silently, and a rename would red seven sibling `SendMessage` tests unambiguously. Speculative future-proofing against an impossible-today scenario.
- **The two unproven engine-behaviour caveats need a tracking issue.** Refuted: issue #53's wording is two-branch — "Settle that before relying on the assertion; invoking the private `OnApplicationPause` by reflection is *the assertion-visible alternative*." The PR took branch two. The unsettled `SendMessage` question is now load-bearing for zero tests.

---

Stated so these read as unverified rather than as passing checks:

- **Every Unity test run in the PR body.** This is a bare UPM package with no `Assets/`/`ProjectSettings/`; `Tests~` runs only inside the host project, and the result XMLs are not in this repo and were deleted after the run. The reviewers could not re-execute them. They *did* check internal consistency against source, and it held: the assertion message string matches `:137` character-for-character; `ArrangeContinuousAndRunning()` has exactly 5 call sites, matching "5 failed"; the sibling `ListeningMode_SetToContinuous_FiresOnTalkStarted` exists at `VoxrPushToTalkControllerTests.cs:184`, asserts only mode and event count, and sits in a 16-test fixture; and "the only check in the whole `Tests~` tree" survives a tree-wide grep across all 31 files.
- **On-device (Quest) behaviour.** Not applicable here — no production code changed — and not claimed by the PR.
- **Whether `void OnDestroy() { }` actually shadows the base's private message.** A Unity engine question, unanswerable from this repo. Both outcomes are harmless: if it shadows, the P/Invoke is avoided; if not, behaviour equals today's `main`.

One factual correction to the PR body, noted rather than silently edited so the reviewed artifact stands: it says the shim's benefit "lands on Android/macOS/Linux". `ReleaseNativeResources` catches `DllNotFoundException` (`VoxrSpeechRecogniser.cs:199-203`) and only `Runtime/Plugins/Android/arm64-v8a/libvosk-bridge.so` ships, so the hazard is **Android-only**.

---

**Disposition.** No CONFIRMED blocker, so nothing was fixed on the branch — per the issue lane's rule, everything below blocker stays posted here for the human, unfixed and unruled. Findings 1 and 4 are comment-text edits; finding 2 is one clarifying clause; finding 3 is an optional extra test case; finding 5 touches no code. This review is evidence, not a ruling, and not an approval.

PR #56, opened 2026-08-08
Keep the PlayMode test assembly out of Android players (#55)

### LINKED ISSUE #55: PlayMode test assembly is built for all platforms — on-device runs would free the process-global bridge mid-suite
`Tests~/Runtime/Jinwoo1601.VoXR.Tests.Runtime.asmdef` sets `"includePlatforms": []`, so the assembly is built for **every** platform. That is harmless today, because the suite only ever runs in the Windows editor. It stops being harmless as soon as the suite is run on a device (`-testPlatform Android`), which is planned for the on-device verification rig.

### Why it breaks on device

`VoxrSpeechRecogniser._bridgeAvailable` defaults to `true`:

https://github.com/jinwoo1601/VoXR-Speech-Recognition/blob/main/Runtime/VoxrSpeechRecogniser.cs#L44

so on any non-`UNITY_EDITOR_WIN` platform, destroying *any* recogniser component calls into the native bridge, whether or not that instance ever initialised it:

```csharp
public void ReleaseNativeResources()
{
#if UNITY_EDITOR_WIN
    ...
#else
    if (_bridgeAvailable)
    {
        try { BridgeNative.vosk_bridge_destroy(); }
        ...
```

`vosk_bridge_destroy()` is process-global. It early-returns only when nothing is initialised; otherwise it stops the capture, joins the recognition thread, and frees the shared recognizer and model (`NativeBridge~/src/vosk_bridge.cpp:247-272`).

The result on an Android run: every fixture that creates a real `VoxrSpeechRecogniser` and lets it be destroyed tears down the bridge belonging to whatever else is mid-test. The fixtures that do so are `VoxrSpeechRecogniserInjectionTests`, `VoxrSpeechRecogniserLifecycleTests`, `VoxrCommandRecogniserInjectionTests`, and `VoxrPushToTalkControllerTests`. (`VoxrPlaybackSeamTests` and `VoxrWavReplayTests` are already `#if UNITY_EDITOR_WIN`; the `FakeSpeechRecogniser` double in `VoxrPushToTalkPauseTests` was made inert in #53.)

Nothing here fails to compile — there is no `using UnityEditor` in the suite — so this would surface as confusing cross-test flakiness on device, not a build error.

### Proposed fix

Scope the existing PlayMode assembly so it is not built into Android players, leaving on-device tests to their own assembly. The exact asmdef mechanism should be chosen by whichever one preserves the current EditMode/PlayMode split (116 EditMode + 304 PlayMode) — `includePlatforms: ["Editor"]` risks reclassifying the suite as EditMode, so it needs verifying against a real run rather than assuming.

### Out of scope, noted separately

The same code path is a latent product-level bug independent of tests: two `VoxrSpeechRecogniser` components in a user's scene share one process-global bridge, so destroying either one frees the other's recognizer and model. Worth its own issue; not addressed here.

jinwoo1601:
Fixes #55.

`Tests~/Runtime`'s asmdef had an empty `excludePlatforms`, so the PlayMode suite was built for every platform. That is inert today — the tests only ever run in the Windows editor — but it blocks on-device test runs, because every fixture would be shipped into the player and `vosk_bridge_destroy()` is process-global (details and file references in #55).

### The change

One field: `"excludePlatforms": ["Android"]`.

### Why this mechanism

`includePlatforms: ["Editor"]` is the more obvious-looking fix and is wrong here: Unity classifies a test assembly whose only platform is Editor as an **EditMode** assembly, which would move all 304 PlayMode tests into EditMode and change how the `[UnityTest]` coroutines run. Excluding Android instead keeps the assembly a PlayMode assembly while dropping it from Android players.

### Verification

Full suite on the host project (Unity 6000.4.7f1), both platforms, after the change:

| Platform | Total | Passed | Failed | Inconclusive |
|---|---|---|---|---|
| EditMode | 116 | 116 | 0 | 0 |
| PlayMode | 304 | 304 | 0 | 0 |

Unchanged from the recorded baseline, which is the point — the split is preserved.

### Not verified

The runs above had the editor's active build target on Windows Standalone. I did not verify that the assembly still compiles for the editor while the **active build target is Android** — the configuration an on-device run leaves behind. The platform filter should not affect editor compilation (Editor remains in the assembly's effective platform set), but that is reasoning, not a measurement, and it belongs to the on-device rig's runbook rather than this fix.

### Left alone deliberately

The underlying `ReleaseNativeResources()` behaviour is also a product-level bug — two `VoxrSpeechRecogniser` components in a user's scene share one process-global bridge, so destroying either frees the other's recognizer and model. That is a shipped-runtime semantics change and wants its own issue; this PR only stops the test suite from tripping over it.

### DIFF
--- a/Tests~/Runtime/Jinwoo1601.VoXR.Tests.Runtime.asmdef
+++ b/Tests~/Runtime/Jinwoo1601.VoXR.Tests.Runtime.asmdef
@@ -7,7 +7,9 @@
         "UnityEditor.TestRunner"
     ],
     "includePlatforms": [],
-    "excludePlatforms": [],
+    "excludePlatforms": [
+        "Android"
+    ],
     "allowUnsafeCode": false,
     "overrideReferences": true,
     "precompiledReferences": [


jinwoo1601 (commented):
## Automated review — PR #56

**Depth profile:** slim (issue lane) — angles `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Reviewed at:** head `92c2b8bb13e1c17aec0174e0c8057498b01e5196`, against merge-base `dfeb06599c17863cf48312519abdeebed6b7f0ad`.
**Method:** six finder agents ran in parallel with no context from the session that wrote the change, reading only a brief and the diff. Every surviving finding then went to an independent verifier prompted to *refute* it. Six findings were raised; **one survived verification**.

---

### Finding 1 — the promised follow-up issue was never filed (minor, high confidence) — **CONFIRMED**

Both issue #55 (*"Worth its own issue; not addressed here."*) and this PR (*"That is a shipped-runtime semantics change and wants its own issue"*) defer a live shipped-runtime defect to a ticket that does not exist. A search of all issues finds no such ticket. Merging this PR closes #55 via `Fixes #55.`, at which point no **open** item tracks the defect — and this project binds `gh issue list` / `gh pr list` as its only orientation surface ("No workflow board"), so an unfiled issue drops out of the backlog entirely.

The verifier also established the defect is **worse than either the issue or this PR describes**. Not merely "destroying one recogniser frees the other's resources":

- `Runtime/VoxrSpeechRecogniser.cs:44` — `bool _bridgeAvailable = true;` (default)
- `:188-207` — `ReleaseNativeResources()` calls `BridgeNative.vosk_bridge_destroy()` in the `#else` arm of `#if UNITY_EDITOR_WIN`, i.e. live on Android
- `:515-517` — `OnDestroy()` calls it unconditionally
- `:130` — `InitialiseAsync` early-returns on `if (IsInitialised)`, which reads the **process-global** `vosk_bridge_is_initialised()`

So a second `VoxrSpeechRecogniser` in a scene never creates its own native state — it silently shares the first's, and *either* component's `OnDestroy` tears down both. `NativeBridge~/src/vosk_bridge.cpp:247-272` frees the file-static `g_recognizer`/`g_model`. The `DllNotFoundException` catch does not mitigate this on Android, because `Runtime/Plugins/Android/arm64-v8a/libvosk-bridge.so` ships, so the P/Invoke resolves and executes. Nothing in the public API forbids two components.

**Suggested remedy** (no code change, does not touch this diff or the 116/304 split): file the product-bug issue and replace this PR body's "wants its own issue" with its number, so closing #55 does not retire the only open record.

---

### Considered and killed

Five findings were raised and refuted on the evidence. Named here so it is visible what was weighed:

- **The headline claim is unverified** (the verification run had the active build target on Windows Standalone, where the change is a no-op) — *refuted*: `defineConstraints: ["UNITY_INCLUDE_TESTS"]` means a normal Android player never contained this assembly regardless; only a Test-Runner Android build is affected, and that does not exist yet. The proposed remedy also would not have measured the claim — an assembly found *absent* with Android active would violate the "editor path still works" invariant, not confirm the exclusion.
- **The body attributes the behaviour to the wrong asmdef field** (`excludePlatforms` vs the operative `includePlatforms`) — *refuted*: elliptical, not wrong. An empty `excludePlatforms` is a genuine necessary condition for Android inclusion, and no path from the wording leads to an incorrect change; the very next paragraph analyses and rejects `includePlatforms: ["Editor"]` on correct grounds.
- **Whole-assembly exclusion is broader than the stated hazard**, losing device coverage for ~300 managed tests — *refuted*: the cited exhibit `Utf8ParserAvailabilityTests` names its own method `Utf8Parser_IsAvailable_InEditorRuntime` (`:13`) and a stripped profile cannot remove a symbol the assembly statically references; `ZeroAllocPollPathTests` budgets allocations against **Mono** object layout (`:13-16`), so shipping it to IL2CPP would be a flakiness source, not coverage. Also arithmetically off: the bridge-free remainder is ~236 cases, not ~300.
- **An empty run is indistinguishable from a passing run** (a zero-test Android run satisfies `failed="0"` vacuously; UTF drops missing assemblies silently and exits 0) — *refuted*: the cited `continue` in `EditorLoadedTestAssemblyProvider.cs:56-59` is not the path a platform-filtered assembly takes (it is never compiled, so never loaded, so never reaches that loop), and the vacuity of `failed="0"` is pre-existing on `main` — dropping `testables` produces the same total=0/failed=0 today. Not created or worsened here.
- **The deferred check is handed to a runbook in no shared location** — *refuted*: this PR body is itself the durable artifact, and the project's planning layer being local-only is a deliberate, documented binding rather than a gap this PR introduced.

---

### Corrections to reasoning offered during review

One finder argued that Android-only exclusion is the right cut because the `vosk_bridge_destroy()` call sites are guarded by `#if UNITY_ANDROID && !UNITY_EDITOR`. **That reasoning is wrong** — `Runtime/Native/BridgeNative.cs` carries no platform guards at all, and the call at `VoxrSpeechRecogniser.cs:198` sits in the `#else` arm of `UNITY_EDITOR_WIN`, so it compiles for every non-editor-Windows target. The conclusion still holds, for a different reason: on any platform that does not ship `libvosk-bridge.so` the call raises `DllNotFoundException`, which is caught at `:200-203`, and Android arm64 is the only supported player target that ships it.

### Corroborations worth recording

- The **116 / 304** split was re-derived independently from source by two agents: EditMode 107 `[Test]` + 9 `[TestCase]` = 116; Runtime 289 declared cases with `Replay_ManifestCase`'s `[ValueSource]` expanding 1 → 16 fixtures = 304. Both land exactly.
- The PlayMode classification is preserved **structurally**, not just empirically: UTF decides from `AssemblyFlags.EditorOnly` alone (`com.unity.test-framework/.../EditorLoadedTestAssemblyProvider.cs:62`). An assembly still built for non-Android players cannot carry that flag.
- `"Android"` is the exact token Unity accepts, confirmed against a shipped Unity package asmdef and against `UnityEditor.dll` string literals. Unity itself ships the identical shape (`Unity.Collections.PerformanceTests`: `includePlatforms: []` + `excludePlatforms: ["iOS"]`).
- No dangling references: nothing in the repo references `Jinwoo1601.VoXR.Tests.Runtime` except its own `name` field and the `InternalsVisibleTo` grant in `Runtime/AssemblyInfo.cs`, which is name-only.

- Any behaviour of a real Android player build or a connected Quest — no device is attached.
- Whether the assembly still compiles for the editor while the **active build target is Android**. This PR already discloses that gap. Agents found supporting evidence (`UnityEditor.dll` exposes `IsCompatibleWithEditor` as a method distinct from the active-target checks) but this remains reasoning, not measurement.
- The asserted test totals were not re-run; they were corroborated from source only.

---

**Verdict: no blocker.** The one confirmed finding is a tracking gap, not a code defect, and asks for an issue rather than a change to this diff. Ruling on it is the human's call; nothing here has been fixed or merged automatically.

Issue #42, opened 2026-07-29
Bare pattern out-scores its slot-filled sibling when a required literal is dropped — spoken slot value silently discarded


jinwoo1601:
## Summary

When VOSK drops a required literal from an utterance that otherwise fills a specific pattern, the specific pattern is penalized (`RequiredLiteralMissPenalty`) while a bare, less-specific sibling still scores a clean 1.0 — so the bare pattern wins and the **spoken slot value is silently discarded**. The command executes with a default the captain never asked for, and nothing signals that a slot value was heard and thrown away.

## Observed in-headset

Grammar: `Cmd_Decelerate` with patterns `decelerate` and `decelerate by {burn_level}`.

Spoken: **"decelerate hard burn"** (the "by" elided by the speaker, or dropped by VOSK — short unstressed function words are the most-dropped tokens in practice; compare "time to target" → "time target" from the same test cycle).

Result: `decelerate by {burn_level}` was charged for the missing required "by" and scored below the bare `decelerate`'s 1.0 → bare pattern won → the heard "hard burn" was discarded and the command executed at the default level. Diagnosed from the session log; worked around content-side by rewording the pattern to `decelerate {burn_level}` (no literal to drop).

## Why this is worth a parser-level look

The scoring is locally correct — a required literal *was* missing. The problem is the comparison across siblings: the utterance contains recognized, in-grammar slot content that one candidate explains and the other ignores, and the ignoring candidate wins. The same word-drop fragility shows up elsewhere (a dropped function word pushed a 3-word pattern to 0.50 < the 0.60 default `minScore`, while a 7-word pattern absorbed the identical drop at 0.79) — short/specific patterns are structurally the most fragile to single-word drops.

## Possible directions (enhancement, not a prescriptive fix)

- A coverage term in sibling comparison: when candidates start at the same token, weigh how much of the utterance each explains, so a pattern that accounts for the spoken slot content isn't beaten by one that leaves it unconsumed. (Related to the tie-break span issue, but this case is not a tie — the specific pattern scores *lower*.)
- Softer miss penalty for optional-izable function words, or authoring guidance: the practical content-side fix was removing the droppable literal (`decelerate {burn_level}`), which may be worth documenting as a pattern-authoring rule ("don't place a required function word between the verb and its slot") in KNOWN_LIMITATIONS/command-recognition docs.
[...]

Issue #31, opened 2026-07-26
Sliding-start matching lets stray utterance tails execute short patterns


jinwoo1601:
Observed in-headset (The Long Burn, VoXR 1.3.0, constrained-grammar VOSK), twice in one session:

1. **"thrusters port"** (a retired form, no longer in the grammar) — VOSK snapped the out-of-vocab "port" to **"report"**; the parser's sliding start skipped the unmatched leading word "thrusters" and matched the lone "report" against the report-query pattern at full score. The query executed.
2. **"half closure three clicks on track…"** (mid-order stumble) — the sliding start found "closure" mid-utterance and executed the closure query.

Both absorbed patterns happened to be read-only queries, so no harm — but the mechanism is general: any stray utterance whose **tail** resembles a short pattern can fire that pattern, because skipped leading words carry no score penalty. If a short *actuating* pattern ever exists in a grammar, a sentence fragment could trigger it.

Suggested directions (any one would do):
- score-penalize skipped leading words, so a 1-word match inside a 2+-word utterance no longer scores 1.0;
- require short (1–2 token) patterns to anchor at utterance start;
- a configurable minimum matched-fraction of the utterance.

PR #58, opened 2026-08-11
Warn when a droppable required literal guards a slot (#42)

jinwoo1601:
Addresses #42.

## What

`VoxrCommandParser`'s constructor now warns when a grammar carries the shape the issue reports: a bare pattern plus a longer one that extends it with **required literals followed by a slot** (`decelerate` / `decelerate by {burn_level}`). The warning names the intents, both patterns, the literal, and the slot at risk, and states the fix — mark the literal optional (`?by`).

The scan follows what `ParseInternal` actually compares, so it catches the hazard in every form it takes — widened from the first draft after review:

- **Across commands**, not only within one command's pattern set. Selection runs over every pattern of every command through a single `IsBetterCandidate`, so splitting the two phrasings into separate intents (`decelerate` / `decelerate_by`) reproduces the hazard exactly.
- **Over a run of required literals**, not exactly one. Dropping any single word in `decelerate by the {burn_level}` strands the value just as dropping `by` alone does.
- **Over expanded optional forms**, as `ComputeCanCommitEarly`'s prefix analysis already does. `fire {?quantity} {weapon}` is not literally a prefix of `fire {weapon} at {target}`, but it is once its own optional is omitted — which is the form the parser matches when no quantity is spoken.

Widening cost was measured rather than assumed: **0 new warnings** on the 11-intent/32-pattern demo grammar for every variant, and exactly the one expected warning on the pre-#42 grammar. It does find real things — the cross-command scan immediately surfaced a genuine `status report {target}` hazard in an existing test grammar that the narrow version could not see. Patterns carrying more than six optional elements are compared unexpanded, since this scan runs on every parser rebuild and expansion is exponential; that costs recall on those patterns only.

Also in this PR:

- **Docs.** A new authoring rule in `Documentation~/command-recognition.md` ("Never leave a required function word between a bare pattern and its slot") with the hazardous/safe code pair, and a `KNOWN_LIMITATIONS.md` entry with the repro, the score arithmetic, and the workaround. The existing "Validation warnings re-emit on every active-set switch" entry names `RunValidationWarnings()` as the sole ctor validation pass; broadened by one clause since that is no longer true.
- **Sample + demo grammar.** `orient heading {heading} mark {?elevation}` carries the hazardous shape — a dropped "mark" scores 0.7 against the bare `orient heading {heading}`'s 1.0 and strands the elevation. Switched to `?mark` in `Tests~/Runtime/DemoGrammar.cs`, `Samples~/CommandRecognition/CommandDemo.cs`, and `Cmd_SetHeading.asset`, so the shipped sample demonstrates the guidance instead of tripping the new warning. **This has a cost**, surfaced by the review and now documented and tested: an optional literal stops anchoring the slot behind it, so a spurious 4th digit ("orient heading two seven zero **four**") is absorbed as `elevation = "four"` where the required form dropped it. Reverting these three edits is a reasonable alternative — see the review's "Left for you".
- **Tests.** Ten PlayMode tests: the warning fires; the reported failure is reproduced end-to-end (bare pattern wins, `burn_level` absent); the `?by` remedy keeps the slot with *and* without the spoken literal while a bare "decelerate" still matches the bare pattern; non-hazard shapes stay silent; an optional trailing slot warns too; the shipped `?mark` grammar is pinned on both sides — the elided-literal recovery it buys and the stray-digit cost it carries; and each of the three widened detector variants warns.

**No _parser_ behaviour changed.** Selection, scoring, and eager flush are untouched — the parser only logs. The `?mark` grammar edit above is a separate matter: it does change which pattern the sample's `set_heading` selects on some utterances, deliberately, and its costs are documented below and in `KNOWN_LIMITATIONS.md`.

## Why

The issue offers three directions and is explicit that none is prescriptive. Two of them — the validation warning ("at minimum") and the authoring guidance — are fully specified, and are what this PR implements. The third, a coverage term in sibling comparison, is deliberately **not** taken, for a reason worth recording:

- The bare pattern scores a clean **1.0** (`1/1`). The slot-filled sibling, charged `RequiredLiteralMissPenalty` for the dropped "by", scores **0.5** (`1.5/3`). Since every score normalises to a 1.0 ceiling, **no penalty or threshold tuning can make the specific pattern outrank the bare one** — the issue is right that this is not a tie-break problem.
- The mechanism the issue leaves open spans at least two shapes, and they behave differently. A new ordering key *above* `Score` in `IsBetterCandidate` flips selection but leaves 0.5 at the `minScore` gate, so the command stops firing rather than fires with the slot filled. Folding coverage *into* the score can do both — an additive weight `w > 0.75` puts the sibling ahead of the bare pattern and over 0.6 — but at that weight coverage is nearly as heavy as the match score itself, and it collides with the #31 skipped-word penalty (already a coverage fraction over leading tokens) and with the eager-flush scan, which gates the unpenalised selection score. Either way this is a scoring-model redesign with grammar-wide blast radius: a maintainer's design decision, not an issue-lane fix.
- Meanwhile the authoring fix reaches the desired outcome exactly, with machinery already in place: with `?by`, an omitted optional leaves both sides of the ratio, the sibling scores 1.0 whether or not the word was spoken, and the consumed-span tie-break from #41 gives it the win. That is what the warning tells authors to do, and the tests prove it.

So this PR closes the "nothing signals that a slot value was heard and thrown away" gap at authoring time, for the shape the issue reports. Whether that is enough to close the issue is the maintainer's call — hence "Addresses" rather than "Closes". What stays open, stated explicitly:

- **The scoring-model direction**, above.
- ~~Wider forms of the same shape.~~ **Closed** — the detector was widened after review to all three variants, described under "What" above.
- **The issue's second symptom**, which has no bare sibling at all and so is out of reach of both the detector and the authoring rule: "a dropped function word pushed a 3-word pattern to 0.50 < the 0.60 default `minScore`, while a 7-word pattern absorbed the identical drop at 0.79 — short/specific patterns are structurally the most fragile to single-word drops."

## Validation

- [x] Unity **EditMode 116/116** and **PlayMode 314/314** = **430/430 passed, 0 failed, 0 skipped, 0 inconclusive**, via the host project `VoXR TestGround` (Unity 6000.4.7f1), re-run after the review fixes and again after the detector was widened. Both result XMLs written to a fresh unique directory and their UTC `start-time` checked against a baseline stamped immediately before arming, since stale result files sit in `TestResults/` and would otherwise read as current. Each of the ten new tests was confirmed by name in `playmode.xml` with `result="Passed"`, not by trusting the total.
- [x] Score arithmetic in the docs/changelog traced against `TryMatchScored` by hand for all four cases (`decelerate`, `decelerate by hard burn`, `decelerate hard burn`, and the `?by` variants of each).
- [x] Grammar vocabulary is unaffected by the `?mark` change — `GenerateGrammarJson` strips the leading `?` from optional literals, so the harness's pinned 98-word list is unchanged.
- [ ] No on-device (Quest) check needed: this change only adds a construction-time log and edits docs/sample grammar; no runtime, lifecycle, mic, or native-bridge path is touched.


jinwoo1601 (commented):
## Independent review — issue lane, **slim** profile

**Reviewed:** `ead39d4adaa9e5745e721a81ebb56d01653b0386` against merge-base `ad5bd48`.
**Angles run (6):** `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP` — the full slim profile, nothing dropped. Each ran in a fresh context with no knowledge of the session that wrote the code; every surviving finding was then handed to a separate verifier prompted to *refute* it.

**Outcome: 4 CONFIRMED (3 minor, 1 nit), 1 REFUTED, 2 angles clean.** No blockers. Every confirmed finding is a false or incomplete statement in prose I wrote, not a defect in the matching logic — `LINE` and `XFILE` both came back clean after re-deriving the new method and tracing every cross-file contract.

---

### CONFIRMED — minor

**1. `GONE-1` — the `?mark` sample edit lets an unmarked stray digit become an elevation.**
`Tests~/Runtime/DemoGrammar.cs:123` (twins in `CommandDemo.cs`, `Cmd_SetHeading.asset`)

A required literal is the only element that must be *spoken* before the next one can consume anything. Making it optional removes that anchor. On `orient heading two seven zero four` (a spurious 4th digit past the maxed-out 3-word heading):

| | pattern 1 score | winner | result |
|---|---|---|---|
| base (`mark` required) | `3.5/5 = 0.7` | pattern 0 at `1.0` | stray digit correctly dropped |
| head (`?mark`) | `4/4 = 1.0`, span 6 | **pattern 1** (span 6 > 5) | `elevation = "four"` — fabricated |

Verified independently, arithmetic exact. The reachability is not hypothetical: `KNOWN_LIMITATIONS.md:147-166` names `"four"` as one of the three words a cough or hum most often becomes in grammar mode. `ComputeConfidence` also now spans the extra token, so a low-confidence stray digit can sink the whole command under `minConfidence`.

The verifier's sharpening is the important part: this is not a bug in the sample edit, it is the **inseparable price of the remedy this PR documents** — you cannot buy elided-literal recovery without also letting the slot claim tokens the literal never introduced. Nothing in the diff disclosed that price.

**2. `GONE-2` — the remedy lowers the score of imperfect matches, and the warning string said it didn't.**
`Runtime/Commands/VoxrCommandParser.cs:345`

A matched required literal adds `1.0` to both sides of the ratio; a matched optional literal adds `0.5`. Equivalent only when everything else matches — for `r < d`, `(r−0.5)/(d−0.5) < r/d` always. The shipped warning told authors the swap makes the sibling "score the same", which is false for any partial match. Concretely, `orient heading mark one five` (heading digits lost, "mark" survives) scored exactly `3.0/5 = 0.600` at base — bit-identical to `minScore = 0.6f`, so `<` admitted it — and `2.5/4.5 = 0.556` at head, which nothing clears, routing the utterance to `OnUnrecognisedSpeech`. The verifier simulated all 11 demo commands × every pattern × every start index in both revisions to confirm no other candidate rescues it.

Worth saying plainly: this particular change is an **improvement** (what base fired was `set_heading` with its required heading slot empty — the very "fires with a slot missing" failure this PR documents). The finding is that the guidance claimed a free swap.

**3. `SCOPE` — the detector is strictly narrower than the changelog and source comment claimed.** (merged from `CLAIM-1`, `GAP-1`, `GAP-2`)
`VoxrCommandParser.cs:296,307,332,334` · `CHANGELOG.md:12`

"The **one** shape that silently discards a slot value" is false. Three variants produce the identical silent discard and are not detected — all three confirmed with independent arithmetic:

- **(a) two or more literals before the slot** — `["fire"]` / `["fire","at","the","{target}"]`: bare `1.0` vs sibling `0.625`, `hotel one` stranded.
- **(b) the pair split across two intents** — detection loops sit inside `foreach (var command in commands)`, but `ParseInternal` selects across *all* commands.
- **(c) an optional element inside the bare pattern** — `IsElementPrefix` compares raw elements ordinally. The sharpest form of this: `ComputeCanCommitEarly`'s prefix analysis **in the same file** expands optionals via `ExpandOptionals` and scans cross-command; the new detector does neither.

The verifier costed the widening rather than asserting it: running the shipped detector and three widened variants over the real 11-intent/32-pattern demo grammar produced **0 warnings in every variant** (and on the pre-PR grammar, both shipped and fully-widened emit exactly the one expected warning). So (a) and (b) look close to free; (c) is costlier — it needs its own recursion bound, since `MaxOptionalExpansion` lives in a method this `static` one can't reach.

### CONFIRMED — nit

**4. `CLAIM-2` — the PR body's argument for *not* implementing the scoring fix contained a false step.**

The body claimed a coverage term "would have to rank coverage above score", making the command fall under `minScore` and stop firing. That only holds for a separate ordering key. Issue #42 says "**weigh** how much of the utterance each explains" — a blend. Under an additive blend, any weight `w > 0.75` puts the sibling both ahead of the bare pattern *and* over 0.6, which is exactly the outcome the body called unreachable. The file itself already separates the two knobs: the #31 skipped-word penalty is applied *after* selection specifically so it "filters via minScore without changing which pattern wins".

The **conclusion** (defer to the maintainer; "Addresses", not "Closes") is unshaken — a coverage blend at that weight collides with the #31 penalty and with the eager-flush scan, which gates the unpenalised score. Only the recorded justification was wrong, and that justification is what gets read when deciding whether to keep #42 open. It did not ship in code or docs — PR body only.

### REFUTED

**`PIT-1` — "the new warning spams release Android builds on every `NotifySlotChanged()`".** The mechanism is real (`NotifySlotChanged()` → `RebuildParser()` → ctor; docs do recommend calling it on spawn/death; nothing gates the pass on `UNITY_EDITOR`), but its differentiator collapses: the demo grammar already ships `{ "a", "one" }`, which fires an unsuppressable single-character-alias warning on *every* rebuild today, on `main`. This is the pre-existing "validation warnings re-emit on every active-set switch" limitation, which the brief declared out of scope — and which this PR already broadened to name the new pass. Allocation cost sits behind four `continue` guards, so clean grammars allocate nothing.

### Clean angles

`LINE` — re-derived every changed statement: the `extended.Length < bare.Length + 2` guard is sufficient for every reachable shape (including zero-length, single-element, and equal-length patterns), `IsElementPrefix` has no unguarded call site, the element-classification predicates match `TryMatchScored`'s rules exactly on degenerate inputs (`"?{x}"`, `"{}"`, `"{?}"`), no duplicate or symmetric warnings are possible, and all five tests' asserted indices and slot values follow from the code.

`XFILE` — the `.asset` DSL and the C# array form yield byte-identical elements for `?mark`; the three grammar copies stay parallel and no fourth declaration exists; `GenerateGrammarJson` strips the `?`, so the pinned 98-word harness vocabulary is unchanged and no fixture re-baseline is owed; the asmdef already references `UnityEngine.TestRunner`.

---

### Fixed on this branch in response (pushed as a follow-up commit)

All four confirmed findings were statements I wrote that were false or incomplete, so they are corrected rather than left for you to arbitrate:

- The warning string no longer claims the remedy "scores the same" — it now states the swap's two costs (`GONE-2`).
- `Documentation~/command-recognition.md` and `KNOWN_LIMITATIONS.md` disclose both costs, with the stray-digit case spelled out (`GONE-1`, `GONE-2`).
- "The one shape" → "a common shape" in `CHANGELOG.md` and the source comment, each now naming the three undetected variants explicitly (`SCOPE`).
- The PR body's "Why" section is reworded to drop the false impossibility step (`CLAIM-2`).
- Two new PlayMode tests pin the shipped `?mark` grammar — the recovery it buys *and* the stray-digit cost — closing a coverage gap `GAP` raised: the combination actually shipped had no parse-level test.
- Two self-inflicted defects also fixed: a formatter had re-indented `CommandDemo.cs:117` 8 columns out of alignment, and a test comment cited the sample's arithmetic on a different grammar.

### Left for you — not fixed, not ruled

- **Whether to widen the detector** to variants (a) and (b). Measured at 0 false positives on the demo grammar, but it is a scope decision, and the prose now honestly describes the narrow behaviour either way.
- **Whether the sample should keep `?mark` at all.** The verifier recommended disclosing over reverting, since reverting restores the exact shape the new warning flags. I followed that, but the trade is yours: the sample now recovers an elided "mark" and also claims an unmarked stray digit.
- The pre-existing "run validation once per `Configure()` rather than per rebuild" cleanup, already filed in `KNOWN_LIMITATIONS.md` as a low-priority follow-up.

- **On-device (Quest) behaviour** — never claimable from here. Nothing in this change touches mic capture, lifecycle, or the native bridge, so there is no obvious device exposure, but that judgement is yours.
- **The Unity test run is not verifiable *by the review agents*** — this is a bare UPM package with `Tests~` hidden from Unity, so no agent compiled anything. The 427/427 figure comes from the host-project run reported separately, not from the review.

jinwoo1601:
Follow-up push `e32cb77` addresses the four CONFIRMED findings from the review above (`GONE-1`, `GONE-2`, `SCOPE`, `CLAIM-2`), plus the two self-inflicted defects noted in passing. Suite re-run green at 427/427. The two design questions in "Left for you" — whether to widen the detector, and whether the sample keeps `?mark` — are untouched and unruled.

jinwoo1601:
### Detector widened (`425b5d9`) — resolves the `SCOPE` finding rather than documenting around it

The review's `SCOPE` finding is now fixed in code instead of narrowed in prose. The scan follows what `ParseInternal` actually compares, on all three axes:

| Variant | Before | Now |
|---|---|---|
| Pair split across two intents | missed — both loops sat inside `foreach (var command in commands)` | caught; flattened `(command, pattern)` scan |
| Run of 2+ required literals before the slot | missed — required the slot at exactly `bare.Length + 1` | caught; walks the literal run |
| Bare form reachable only by omitting its own optional | missed — raw ordinal element compare | caught; `ExpandOptionals`, as `ComputeCanCommitEarly` already did |

**One correction to my review above.** It relayed the verifier's claim that the expansion variant was costly because `MaxOptionalExpansion` is "unreachable from this `static` method". That is wrong — it is a class-level `const` and both `ExpandOptionals` and `CountOptionalElements` are `static`, so the variant was considerably cheaper than I reported when you made the call.

**Measured, not assumed:** 0 new warnings across the 11-intent/32-pattern demo grammar for every variant, and exactly the one expected warning on the pre-#42 grammar. It does find real hazards — the cross-command scan immediately surfaced a genuine one the narrow version could not see, in an existing test grammar:

```
... is a bare form of "status report {target}" (intent 'status_report'),
which extends it with the required literal "report" in front of slot "{target}"
```

**Bounded:** expansion is capped at 6 optionals per pattern, well under `MaxOptionalExpansion`, because this scan runs unconditionally in the ctor (and so on every parser rebuild) where `ComputeCanCommitEarly`'s expansion is lazy. Past the cap a pattern is compared unexpanded, costing recall on that pattern alone. Messages are deduped through a `HashSet` — one hazard can surface from several form pairs.

Three new tests pin the three variants. Suite green at **430/430** (116 EditMode + 314 PlayMode), fresh XMLs verified against a pre-arm timestamp, all three confirmed by name.

The docs, `KNOWN_LIMITATIONS.md`, and the changelog now describe the wide behaviour — the "deliberately narrow" wording they briefly carried was accurate only for the first draft.

**Still open for you:** whether the sample keeps `?mark`. Unchanged by this push, and analysed in full in the conversation — short version: it trades a lost elevation when "mark" is elided for a fabricated one when a stray digit follows a saturated 3-word heading, with the confidence gate offering partial cover on the second but nothing covering the first.

PR #59, opened 2026-08-11
Skip a leading [unk] run in the eager-flush gate (#43)

### LINKED ISSUE #43: Leading [unk] disables eager flush — station-prefixed utterances always pay the full bufferWindow
## Summary

`TryEagerCommit` requires the winning match to span the whole buffer starting at token 0:

```csharp
// The match must span the whole buffer: anything left over (including trailing
// [unk]) means an in-progress tail that more speech could still complete.
if (bestStartIdx != 0 || bestEndIdx != tokens.Length)
    return EagerCommitVerdict.None;
```

A leading `[unk]` run — e.g. an out-of-grammar station prefix, "Helm, coast" — pushes the best match's start to 1, so the verdict is `None` and the utterance pays the **full `bufferWindow`** (2.0s on Quest 3). `prefixHoldSeconds` (#32) doesn't help: it applies to `HoldExtendable`, not `None`.

This makes eager flush and station-address prefixes mutually exclusive: the same command is instant when spoken bare and pays ~2s of dead air when politely addressed — surprising, since leading-`[unk]` tolerance is otherwise a deliberate feature (the sliding start exists to absorb out-of-grammar preamble, and the #31 skipped-word penalty explicitly never charges `[unk]`).

## Proposed fix

Trim (or skip past) the **leading** `[unk]` run before the whole-buffer check: require the match to start at the first non-`[unk]` token and consume through the end. Trailing `[unk]` must keep blocking exactly as today — a trailing unknown is legitimately an in-progress tail that more speech could still complete, which is the comment's own rationale. Leading `[unk]` has no such ambiguity: nothing that arrives later can retroactively extend the utterance *leftward*.

This was flagged in our earlier eager-flush analysis as "candidate v1.2.3: trim leading unks" when `eagerFlushOnCompleteMatch` first shipped (#25 → 1.2.2), and remains the one common utterance shape that gets no benefit from either eager flush or `prefixHoldSeconds`.

## Scope note

`ComputeConfidence` and the skipped-word accounting already treat `[unk]` as free, so the trim should be a localized change to the whole-buffer span check in `TryEagerCommit` — no scoring or grammar changes involved.

jinwoo1601:
Closes #43

## What

`TryEagerCommit`'s whole-buffer check now measures from the **first recognised token** instead of token 0. A leading `[unk]` run is skipped; the match must still consume everything through the end of the buffer.

- `Runtime/Commands/VoxrCommandParser.cs` — the gate (one loop + one changed condition, plus comment).
- `Tests~/Runtime/VoxrEagerCommitTests.cs` — 5 new PlayMode cases.
- `CHANGELOG.md` — `[Unreleased] / Fixed`.
- `Documentation~/command-recognition.md` — the eager-flush bullet list and the "spans the whole buffer" line under prefix hold, which no longer described the gate accurately.

## Why

The gate required `bestStartIdx == 0`, so an out-of-grammar station prefix ("Helm, coast") pushed the match start past 0 and forced `None` — the utterance paid the full `bufferWindow` (2.0s on Quest 3). `prefixHoldSeconds` did not help: it arms on `HoldExtendable`, not `None`. So eager flush and station-address prefixes were mutually exclusive, even though leading-`[unk]` tolerance is a deliberate feature elsewhere (the sliding start absorbs it, the #31 skipped-word penalty never charges it, `ComputeConfidence` skips it).

Trailing leftovers keep blocking, unchanged — a tail, recognised or `[unk]`, is an in-progress utterance more speech could complete, which is the whole-buffer rule's own rationale. Nothing arriving later extends an utterance leftward, so the leading run carries no equivalent ambiguity.

**Why `[unk]`-only, and not "skip any leading leftover":** `ParseInternal` charges the skipped-word penalty (#31) against the winner *after* selection (`VoxrCommandParser.cs:591`); this scan does not. That asymmetry was masked before because `bestStartIdx == 0` meant there were never skipped words. It stays masked only because `CountRecognisedTokens` excludes `[unk]` and the new gate admits nothing else ahead of the match — so the skipped count is 0 in exactly the admitted cases and the eager score still equals the flush-time score. Widening it further would let the gate commit a buffer the subsequent flush then scores below `minScore`. `TryEagerCommit_LeadingRecognisedLeftover_StillReturnsNone` pins that boundary.

No scoring, selection, or grammar-analysis change; this only widens which buffers reach a verdict.

## Validation

- [x] Unity **EditMode** via host project `VoXR TestGround` (6000.4.7f1): 116/116 passed, 0 failed.
- [x] Unity **PlayMode**: 319/319 passed, 0 failed — including all 17 `TryEagerCommit_*` cases (12 pre-existing + 5 new), with `TryEagerCommit_AgreesWithParseInternalSelection` and `TryEagerCommit_TrailingUnkStillHoldsTheFullWindow` guarding the blast radius.
- [x] New cases: leading `[unk]` (single and multi-token run) commits and agrees with `Parse`; leading `[unk]` + prefix command now reaches `HoldExtendable`; leading `[unk]` with a trailing leftover (word *or* `[unk]`) still `None`; leading *recognised* leftover still `None`; all-`[unk]` buffer still `None`.
- [x] Working tree restored after the run (`Tests` → `Tests~`, `.meta` orphans removed, host manifest untouched).
- Note: after that green run, the only further edits were the added code comment and the two doc lines — no behavioural change was made post-test.
- No on-device Quest check needed: no mic-capture, lifecycle, or native-bridge code is touched. The change is pure parser logic covered by the PlayMode suite.


jinwoo1601 (commented):
## Independent review — PR #59

**Depth profile:** slim (issue lane) — angles run: `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Head reviewed:** `4f809387eb93c0813bdfce034e6da18f77a14d7e` · **merge-base:** `7e69e0ed1e47c84f8e49c16bf60c16daf0272605`
**Method:** six finder agents with no context from the session that wrote the code, then one adversarial verifier per surviving finding. The PR body was treated as claims to audit, not as context.

**4 findings, all CONFIRMED: 1 major, 1 minor, 2 nits. None is a blocker — the shipped gate is correct.** Three angles independently re-derived the production logic and agreed: the change is a strict widening (no buffer that used to `Commit` can now be refused), eager/flush score parity holds, and invariant "leading *recognised* leftovers still block" is enforced by the code. The findings are about the **evidence** for the change, not the change.

---

### 1. `major` · CONFIRMED (5 of 6 angles) — the new condition's restrictive half has no discriminating test, and the PR body claims it does

`Tests~/Runtime/VoxrEagerCommitTests.cs` — `TryEagerCommit_LeadingRecognisedLeftover_StillReturnsNone`

The PR body ends its safety argument with: *"Widening it further would let the gate commit a buffer the subsequent flush then scores below `minScore`. `TryEagerCommit_LeadingRecognisedLeftover_StillReturnsNone` pins that boundary."* **It does not.** That test never reaches the new leading-token check.

On `Tok("hello cease fire")` against `Cmd("cease_fire", P("cease","fire"))` + `Cmd("greet", P("hello"), P("hello","there"))`, `IsBetterCandidate` gives earliest start absolute priority (`Runtime/Commands/VoxrCommandParser.cs:803`), so `greet`/`P("hello")` wins at index 0 with score 1.0 — `P("cease","fire")` at index 0 scores `(-0.5 + -0.5)/2 = -0.5` and is discarded at `:799`. Winner: `bestStartIdx = 0`, `bestEndIdx = 1`, and `firstRecognisedIdx = 0` too — so `bestStartIdx != firstRecognisedIdx` is **false**, and the `None` comes solely from the *unchanged* `bestEndIdx != tokens.Length` clause. The test asserts a trailing-leftover case while its comment describes a leading one.

Verified mutant-survival, the stronger half: across all 17 `TryEagerCommit_*` cases, 14 have `bestStartIdx == 0`, two exit early at `bestCommandIdx < 0`, and the three leading-`[unk]` cases satisfy `bestStartIdx == firstRecognisedIdx`. The recogniser-level eager tests (`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:273-410`) all inject utterances matching from token 0. **Deleting `bestStartIdx != firstRecognisedIdx` from `Runtime/Commands/VoxrCommandParser.cs:1281` leaves the entire suite green** — i.e. the suite does not distinguish this PR's gate from exactly the widening the PR argues is unsafe.

The shipped behaviour is right; the regression net under it is empty. Minimal fix, verified by re-derivation: **drop the `greet` command from that test's grammar** and keep the input. With `Commands(Cmd("cease_fire", P("cease","fire")))` alone, nothing scores > 0 at index 0, so the winner is `cease_fire` at `bestStartIdx = 1`, `bestEndIdx = 3 == tokens.Length`, `firstRecognisedIdx = 0` — the new clause becomes the sole producer of `None` and the mutant fails. (A closer-to-production variant keeps a second command so the leading word is genuinely in-grammar but starts no positive-scoring match — e.g. slot `target = ["hotel one"]`, `Cmd("fire_at", P("fire","at","{target}"))`, input `"hotel cease fire"`.)

### 2. `minor` · CONFIRMED (2 angles) — the new prose states the trailing rule as an absolute the gate does not enforce

`Documentation~/command-recognition.md:276` and the new `CHANGELOG.md` bullet

The added bullet says *"anything left over at the end -- recognised or `[unk]` -- is treated as an in-progress tail and keeps waiting"*; the CHANGELOG says *"recognised or `[unk]`, a tail is an in-progress utterance"*. The `[unk]` half is not true in general. `TryMatchScored`'s skip loop runs before **every** element and never rewinds when that element then matches nothing (`Runtime/Commands/VoxrCommandParser.cs:846`, with `EndIdx = tokenIdx` at `:925`), so a trailing `[unk]` is absorbed into `EndIdx` whenever the pattern's last element misses.

Worked counterexample: `P("set","course","for","home")` on `set course for [unk]` scores `(1+1+1-0.5)/4 = 0.625 ≥ minScore 0.6`, `EndIdx = 4 == tokens.Length`, pattern is terminal → **`Commit` with an `[unk]` tail still present**. Not contrived — the shipped demo grammar has the same shape: `Samples~/CommandRecognition/CommandDemo.cs:95` `{"close","distance","{range}","target","{target}"}` on `close distance cqb target [unk]` scores `(1+1+1+1-1)/5 = 0.6`, hits the gate, and fires `close_distance` with the target slot unfilled.

This behaviour is **pre-existing and unchanged by this PR** — the defect is that the PR newly writes it down as an unqualified guarantee in user-facing docs. One correction to the finding as first reported: the *code comment* is not new prose here — the base comment already said "anything left over (including trailing `[unk]`)"; this PR only inserted "at the END". Suggested wording: keep the absolute for *recognised* tails, and qualify the `[unk]` half as "except where the winning pattern's own trailing element matched nothing, which the skip absorbs".

Worth a separate issue on its own merits: the gate can `Commit` a pattern whose final required element matched nothing, at exactly `minScore`. That is a live pre-existing hazard, not something this PR should fix.

### 3. `nit` · CONFIRMED — `TryEagerCommit_AllUnkTokens_ReturnsNone` cannot reach the new code

An all-`[unk]` buffer is `continue`d at every start index (`Runtime/Commands/VoxrCommandParser.cs:1234`), so control returns at `bestCommandIdx < 0` (`:1262`) before `firstRecognisedIdx` is ever computed. Valid black-box regression coverage; not coverage of the new loop, though the PR lists it among the five new-behaviour cases. Related and worth knowing: the loop's own `firstRecognisedIdx < tokens.Length` bound is **provably dead** — reaching `:1277` requires some non-`[unk]` token, which halts the walk below `tokens.Length`. Harmless defence-in-depth; just not testable.

### 4. `nit` · CONFIRMED — ticked box slightly overstates the `Parse` cross-check

*"leading `[unk]` (single and multi-token run) commits and agrees with `Parse`"* — the `Parse` cross-check exists only for the single-`[unk]` buffer (`Tests~/Runtime/VoxrEagerCommitTests.cs:573-575`); the two-`[unk]` buffer is asserted through the eager path alone. Immaterial in substance: `ParseInternal` skips `[unk]` start indices outright (`:533-534`), `CountRecognisedTokens` excludes `[unk]` (`:1055-1062`), so run length cannot change the result — and `SkippedWord_UnkFillerNotCharged` (`Tests~/Runtime/VoxrCommandParserTests.cs:1024-1035`) already pins `Parse("[unk] [unk] disengage")`. Wording, not risk.

---

### Considered and cleared (nothing refuted, but these were actively hunted)

- **`PIT` returned clean.** `==` on `string` is ordinal and null-safe, consistent with the six pre-existing comparisons against `UnkToken`; no allocation added to a per-VOSK-result path; results arrive main-thread (`VoxrSpeechRecogniser.cs:413` via `Update()`); test attributes/assembly placement correct.
- **`GONE` returned clean on code.** The one behavioural removal (`bestStartIdx != 0`) is fully compensated: leading recognised leftovers still blocked structurally, skipped-word-penalty parity intact (`skipped` is 0, not merely small), `ComputeConfidence` bit-identical since it skips `[unk]` internally, degenerate inputs still guarded.
- **Eager/flush parity (the PR's central safety argument) verified independently by three angles** — `ParseInternal:591-597` charges only `CountRecognisedTokens(tokens, 0, bestStartIdx)`, which is 0 for a pure-`[unk]` lead, so the eager score equals the flush-time score exactly as claimed.
- **Strict-widening property verified:** `bestStartIdx == 0` implies `firstRecognisedIdx == 0`, so no previously-committing buffer can regress to `None`.
- **Other docs are not stale:** `Documentation~/api/command-recogniser.md:16-17` describes extendability only; `KNOWN_LIMITATIONS.md` entries stay accurate (including the `-1.00`-confidence entry whose repro is literally `[unk] cease fire`); Inspector tooltips unaffected.
- Issue #43's proposed fix and scope note are delivered in full, with nothing in the reported problem left unaddressed.

- **On-device (Quest) behaviour** — human-only per the project bindings. The PR's rationale for needing none is sound on audit: no capture, lifecycle, or native-bridge code is touched.
- **The Unity suite results were not re-run by this review.** The 116/116 EditMode and 319/319 PlayMode figures are the author's claim; this review read the test *source* rather than re-running Unity. Test-source auditing is exactly what produced finding 1.
- **"Only a code comment and two doc lines changed after the green run"** is not checkable from the branch — it is a single commit. Indirect corroboration only: the host project's `Jinwoo1601.VoXR.Tests.Runtime.dll` was rebuilt inside the window between branch creation and commit.

---

Nothing here was fixed on the branch: no finding is a blocker, so per the issue lane they are posted for the human to rule on, unfixed. Finding 1 has a verified one-line correction ready if wanted.

PR #60, opened 2026-08-11
Degrade eager flush to a hold on un-analysable grammars, and warn at authoring time (#44)

### LINKED ISSUE #44: Grammar past MaxOptionalExpansion silently disables eager flush set-wide — degrade to HoldExtendable and surface in validation
## Summary

`ComputeCanCommitEarly` guards against pathological grammars: if **any** pattern in the set has more than `MaxOptionalExpansion` (12) optional elements, the precompute returns `null` and eager flush is disabled for the **entire command set**, with only a runtime `Debug.LogWarning`. The rationale in the code is sound — an un-expanded pattern used as the "longer" side of the prefix analysis could unsoundly let a real prefix commit early and fire the wrong command — but the failure mode is a silent, set-wide feature loss triggered by a single over-complex pattern, discoverable only if someone happens to be watching the console.

## Why revisit now

Before #32, the only alternatives were "analysed and possibly instant" or "full `bufferWindow`" — so a whole-set disable was the natural conservative stance. `prefixHoldSeconds` changes the economics: there is now a *cheap* conservative state between "commit instantly" and "wait the full window."

## Proposed changes

1. **Degrade to `HoldExtendable` instead of `None`.** When the grammar exceeds the expansion cap, treat every complete whole-buffer match as `HoldExtendable` rather than reporting no verdict. That is sound (nothing ever commits early that the analysis hasn't vetted — everything holds), and with `prefixHoldSeconds` configured the cost of the un-analysable grammar drops from the full `bufferWindow` to the short hold. Grammars without `prefixHoldSeconds` set keep exactly today's behavior (hold = full window).
2. **Surface the condition at authoring time, not just runtime.** The parser already runs `RunValidationWarnings` on construction; a pattern exceeding the optional-expansion cap is knowable from the assets alone and belongs there (and ideally in the Inspector on the offending `VoxrCommandAsset`), so the author learns which pattern disabled eager analysis *before* a play session, with the specific pattern named — rather than a set-wide console warning after the fact.

## Scope note

Not currently biting our grammar (nothing near 12 optionals per pattern) — filing preventively because the failure is silent and the fix surface is small: the `null` return in `ComputeCanCommitEarly`, the `_canCommitEarly == null` branch in `TryEagerCommit`, and the validation pass.

jinwoo1601:
Closes #44

## What

Two changes to the eager-flush path in `VoxrCommandParser`, both scoped to the case where a single pattern carries more than `MaxOptionalExpansion` (12) optional elements:

1. **`TryEagerCommit` degrades to `HoldExtendable` instead of `None`.** When `_canCommitEarly` is null (analysis abandoned), the verdict is now the held one. It sits *below* every existing gate — score, whole-buffer span from the first recognised token, confidence — so it only ever fires for a buffer that already parses as one complete, confident command. Nothing commits early, which is exactly the verdict the missing analysis would have had to vet; what changes is the wait, from the full `bufferWindow` down to `prefixHoldSeconds` where that is configured. `prefixHoldSeconds = 0` (the default) keeps today's timing exactly.
2. **The condition is reported at construction.** New `WarnOnExcessiveOptionalExpansion(commands)`, called from the ctor alongside `RunValidationWarnings` / `WarnOnDroppableRequiredLiteral`, names the offending pattern text, its intent, its optional count, and the set-wide consequence. It reports every over-limit pattern, not just the first. The old lazy `Debug.LogWarning` inside `ComputeCanCommitEarly` is removed — it fired mid-session on the first eager probe, named only the intent, stopped at the first offender, and never fired at all for an author who had not enabled `eagerFlushOnCompleteMatch`.

Docs: `Documentation~/command-recognition.md` (the prefix-hold bullet claiming too-complex grammars keep the full window was now false, plus a new paragraph on the analysis limit under Eager flush), the `prefixHoldSeconds` row in `Documentation~/api/command-recogniser.md`, and a `### Changed` entry in `CHANGELOG.md`.

**Not done:** the issue's parenthetical "and ideally in the Inspector on the offending `VoxrCommandAsset`". `VoxrCommandAsset` has no `OnValidate` and no custom editor today, so that would mean a new editor surface with no precedent in this package — the sibling authoring warning from #42 is console-only as well. Flagging it rather than expanding scope; say the word and it can be a follow-up issue.

## Why

The guard itself is sound: `ExpandOptionals` enumerates 2^optionals forms, and a partially analysed set could use an un-expanded pattern as the "longer" side of the prefix check and let a real prefix commit early, firing the wrong command. So the refusal has to be whole-set. But the *consequence* was heavier than it needed to be, for two reasons the issue identifies:

- Before #32 the only states were "instant" or "full `bufferWindow`", so a whole-set disable was the natural conservative stance. `prefixHoldSeconds` added a cheap middle state, and the held verdict is exactly as conservative — it just costs less.
- The failure was silent and set-wide: one over-complex pattern quietly removed the feature from every command, discoverable only if someone happened to be watching the console during a play session with eager flush on. The condition is knowable from the assets alone, so it belongs with the other authoring-time validation.

## Validation

- [x] Unity **EditMode 116/116** and **PlayMode 325/325**, `failed="0"` on both (host project `VoXR TestGround`, Unity 6000.4.7f1, run against this working tree). Both suites were run twice — once after the implementation and again after a readability refactor of the new test code — green both times.
- [x] Tests updated/added. Two existing tests asserted the old behaviour: `TryEagerCommit_UnanalysableGrammar_ReturnsNone` is now `..._ReturnsHoldExtendable`, and both over-limit tests needed their `LogAssert.Expect` moved ahead of construction now that the warning is emitted from the ctor. New: the warning names the pattern with no eager probe at all; an at-the-limit (12) grammar stays silent and keeps its analysis; incomplete and unmatched speech still return `None` on an un-analysable grammar; and three recogniser-level tests asserting the effective window is 0.6s with the hold set, 2.0s without it, and 2.0s for half-spoken speech.
- [x] `NativeBridge~/` untouched, so the arm64 build and desktop harness gates do not apply.
- [ ] On-device (Quest) check **not required and not claimed**: the change is confined to `Runtime/Commands/VoxrCommandParser.cs`, with no mic capture, no `OnApplicationPause`, and no native-bridge surface involved.


jinwoo1601 (commented):
## Independent review — PR #60

**Depth profile:** slim (issue lane) — angles run: `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Head reviewed:** `9e3cb0cb4bfb8800bfafae69334bdffe77a3c53b` — **base:** merge-base `ff6dc24eb1a9c8707142a5466ad5596efe914696`.
**Method:** six finder agents, each with fresh context and no knowledge of the session that wrote the code, working from a brief that transcribes the PR's claims and the linked issue verbatim. Every surviving finding then went to a separate verifier prompted to refute it.

**Verdict: no blockers.** The core change holds up under line-by-line re-derivation. The degrade at `VoxrCommandParser.cs:1334-1335` sits below all four pre-existing gates (score, whole-buffer span from the first recognised token, confidence), `CanCommitEarly` still short-circuits on a null table so no `Commit` is reachable, and `EffectiveBufferWindow`'s `prefixHoldSeconds > 0f && prefixHoldSeconds < bufferWindow` guard preserves default timing exactly. The `GONE` angle enumerated all 18 removed parser lines and found nothing dropped that the replacement does not cover.

Four minor findings survived verification, plus one refuted. **None were fixed** — per the issue lane's rule only a confirmed blocker is this run's own work to finish; everything below that is yours to rule on.

---

### 1. Product doc claims an authoring-time surface the package doesn't have — `Documentation~/command-recognition.md:280` (minor, CONFIRMED)

The new paragraph ends: *"…in a warning at construction, so the condition surfaces when the grammar is authored rather than mid-session."* "At construction" is accurate; the clause after it is not. No edit-time hook in this package builds a parser — a repo-wide grep for `OnValidate` / `CustomEditor` / `ExecuteAlways` / `PropertyDrawer` across `Runtime/` and `Editor/` returns zero hits, and the production build sites (`VoxrCommandRecogniser.cs:172`, `:274`) both hang off `Awake()` on a MonoBehaviour with no `[ExecuteAlways]`. Editing a `VoxrCommandAsset` emits nothing; the warning lands at session start, and again on every `NotifySlotChanged` rebuild.

The verifier corrected the finding on one point in the PR's favour and it survived anyway: an edit-time path *does* exist — `VoxrBatchTestRunner.cs:27`/`:61` construct a parser, driven from `Editor/VoxrBatchTestWindow.cs:327` via **Window ▸ VoXR ▸ Batch Test Runner** — but it fires only if the author independently opens that unrelated window and runs a wired-up suite asset, which is not "when the grammar is authored."

Worth noting this sentence is the outlier, not house style: the sibling #42 wording at `:116` (*"logs a validation warning at construction naming the literal and the slot at risk"*) and `:181` name the mechanism without asserting it precedes a session. The `CHANGELOG` entry was checked separately and does not overreach. What the change actually buys — reported whenever the parser is built, unconditionally on eager flush, instead of on the first eager probe — is defensible as written; the "rather than mid-session" clause is not.

### 2. The at-the-limit test doesn't assert the silence it is named for — `Tests~/Runtime/VoxrEagerCommitTests.cs:370` (minor, CONFIRMED)

`OptionalElementsAtTheLimit_DoNotWarn` contains no assertion that observes a log. Three independent angles filed this; the verifier confirmed it from the test framework's own source in the host project's package cache: `LogScope.IsFailingLog` returns true only for `Assert`/`Error`/`Exception`, so an unexpected **warning** never fails a test, while `LogAssert.NoUnexpectedReceived()` throws on any unhandled log of any type.

Concretely: mutate `optionals <= MaxOptionalExpansion` to `<` at `VoxrCommandParser.cs:475` and a 12-optional pattern starts warning — but `ComputeCanCommitEarly`'s independent `> MaxOptionalExpansion` at `:1358` still builds the table, so `CanCommitEarly(1, 0)` stays true and this test stays green. The verifier checked the rest of the suite: no other test uses a 12-optional pattern, so that mutation survives the entire suite. This repo already has the idiom that closes it — `NonHazardPatternShapes_DoNotWarn` ends with `LogAssert.NoUnexpectedReceived()` at `Tests~/Runtime/VoxrCommandParserTests.cs:1362` (also `:1334`, `:1436`).

The verifier also confirmed the remedy is viable today: the at-limit grammar (12 optionals + `cease fire`) trips no other warning, because `WarningForms` compares a >6-optional pattern raw and neither pattern is an element-prefix of the other.

### 3. `HoldExtendable`'s meaning widened in one rendering but not the other three — `VoxrCommandRecogniser.cs:57`, `:131`; `VoxrCommandParser.cs:18-22` (minor, CONFIRMED)

The PR set its own bar by giving `Documentation~/api/command-recogniser.md:17` a third case — *"or any complete match in a grammar too complex for the eligibility analysis"* — and not carrying it to the other places describing the same contract. The Inspector tooltip for that very field still reads *"one that is a prefix of a longer command, or ends in a slot whose value could grow"* (`VoxrCommandRecogniser.cs:57-59`), the `_eagerHoldArmed` comment at `:131-132` says the same, and — inside a file this PR *did* edit — the `EagerCommitVerdict.HoldExtendable` enum doc at `VoxrCommandParser.cs:18-22` still reads *"more speech could still extend it into a longer one … (issue #32)"*. That last one is the authoritative definition of the thing this PR redefined; three other comments in the same file were updated to carry #44.

The widening is real, not theoretical, and this PR's own test proves it: `"cease fire"` is a standalone terminal pattern nothing can extend, yet `UnanalysableGrammar_HoldsInsteadOfPayingTheFullWindow` asserts an effective window of 0.6s. The strongest defence — that a two-item appositive in a length-limited tooltip may reasonably elide a pathological edge case — is why this stays minor rather than being dismissed.

### 4. "Reports every over-limit pattern, not just the first" is claimed but untested — `Tests~/Runtime/VoxrEagerCommitTests.cs` (minor, CONFIRMED)

The behaviour is implemented — `WarnOnExcessiveOptionalExpansion` uses `continue`, with no `break`/`return` in the loop — but a whole-tree scan found exactly two over-limit fixtures, each a single 13-optional pattern in a single command (`VoxrEagerCommitTests.cs:316`, `VoxrCommandRecogniserInjectionTests.cs:659`), and no test anywhere in the package expects two warnings from one construction. The property is asserted only in the PR body; the code comment at `VoxrCommandParser.cs:460-466` says *"naming the pattern responsible"* — singular — so a future short-circuit of that scan would read as a legitimate cost optimisation and pass the suite.

Counterweight, from the verifier: this suite has never tested multi-offender enumeration for **any** constructor warning — `WarnOnDroppableRequiredLiteral`'s six tests all assert a single warning, and `RunValidationWarnings` has no message assertions at all. The finding proposes a standard the repo does not hold elsewhere, which is a fair reason to close it as accepted-as-is.

---

- **Ctor warning re-fires on every parser rebuild (`NotifySlotChanged` → `RebuildParser`), with no dedup.** REFUTED. The property is the established repo pattern, documented in prose 40 lines above the new function (`VoxrCommandParser.cs:411-413`: *"this scan runs unconditionally in the ctor — and so on every parser rebuild"*), and no dedup mechanism exists anywhere in the package. The premise that the removed lazy warning was cheaper is also false: `_canCommitEarlyComputed` is an *instance* field on a parser `RebuildParser` replaces, so base fired at the same per-rebuild rate. `RunValidationWarnings` is arguably worse, since it re-scans the dynamic slot values on each rebuild. The only real deltas — every offender rather than the first, and firing without an eager probe — are issue #44's stated requirements.

- **The Unity suite results.** This is a bare UPM package; `Tests~` runs only inside the host project. The agents read the test sources and reasoned about them but executed nothing, so the PR's `EditMode 116/116` / `PlayMode 325/325` boxes are reported, not independently reproduced, here.
- **On-device (Quest) behaviour.** Nothing in this diff touches mic capture, `OnApplicationPause`, or the native bridge, so the PR's deferral of it is consistent with the change — but it cannot be checked from this environment regardless.

*Nothing in this review approves or blocks the merge — that ruling is the human's.*

PR #61, opened 2026-08-11
Emit phrase-chunked grammar entries instead of a flat word list (#45)

### LINKED ISSUE #45: Emit phrase-chunked grammar entries instead of a flat word list — re-impose sequence constraints at the decoder
## Summary

`GenerateGrammarJson` emits the grammar as a **flat list of single words**. VOSK's grammar format also accepts **multi-word phrase entries**, and a flat word list discards all sequence information: the decoder is free to assemble any in-grammar words in any order, so acoustically-near neighbors substitute for each other with no structural resistance. Emitting contiguous literal runs and slot surface forms as phrase entries ("close to", "on track", "hard burn", "kill the burn") would re-impose sequence constraints at the decoder itself. The command parser is unchanged — recognizer output is still a word sequence; only the hypothesis space the decoder searches gets tighter.

In our month of in-headset testing this was the single most-recommended change from the recognition-accuracy evaluation — "root cause / highest-leverage package change" — because it attacks a whole substitution class at once rather than one confusion pair at a time.

## The substitution class it targets

All observed on Quest 3 with `vosk-model-small-en-us-0.15`, constrained-grammar mode:

- **Destructive substitution risk — close↔coast.** /kloʊs/ vs /koʊst/, both in-grammar. "close to six clicks on track …" decoding as "coast …" doesn't just fail — it fires a *coast* command that kills the standing order it was meant to adjust. Flagged HIGH-risk; a scripted probe (2 close-to orders + 5 bare "coast", 2026-07-26) did not reproduce it that session, but nothing structural prevents it: with a word-list grammar, "close" and "coast" are equally-legal continuations everywhere. With phrase chunks, "close to" exists as a unit and a lone "coast" no longer sits one phoneme away from the start of a range order.
- **Observed wrong-command substitution — "thrusters port" → "thrusters report".** An out-of-vocabulary word snapped to the nearest in-grammar one and (pre-#31) executed an unrelated command. The v1.4.0 `skippedWordPenalty` hardens the *matcher* against the fallout; this issue is about making the *decoder* less likely to produce the bad transcript in the first place. Matcher-layer defenses can't help when a substitution yields a complete, high-scoring command (the close↔coast shape).
- **Digit substitutions** into `NumberSequence` slots, and **clicks↔six** rhyme corruption of range values — score math shows single function-word drops on ≥4-element patterns still clear `minScore`, so a substitution *into a slot* survives scoring and executes with a wrong value.
- **four↔full** ("full stop" → "four stop", 3× in one walk; "all ahead full" → "all ahead four") — same mechanism: with no sequence constraint, the number vocabulary is a legal continuation everywhere, including mid-idiom.

## Proposed change

In `GenerateGrammarJson`, emit as grammar entries:

1. **Contiguous literal runs** from each pattern (split at slot boundaries): `close to`, `on track`, `orient to heading`, `time to target`.
2. **Multi-word slot surface forms** (values and aliases): `hard burn`, `kill the burn`, `maximum burn`.
3. Single words only where a run is genuinely one word.

Keeping `[unk]` and the existing per-set assembly as-is. The parser, scoring, and all consumer-facing APIs stay untouched.

## Open questions / considerations

- Whether single-word entries must *also* remain for words inside phrases, so VAD splits mid-phrase (the `bufferWindow` merge case) can still decode the fragments — worth testing both ways; if fragments need the words listed anyway, the phrase entries still bias the search even when the words remain individually legal.
- Grammar-size growth is bounded (entries ≈ literal runs + slot surface forms, both already enumerable from the assets).
- Dynamic slots (`RegisterSlotValueProvider`) rebuild the grammar already; phrase emission slots into the same rebuild path.
- Interaction with the model's phone-level behavior is empirical — this needs an A/B in-headset pass against the documented confusion pairs (close/coast, full/four, clicks/six) to confirm the gain; the 1.3.0 session logs make that measurable now.

jinwoo1601:
Closes #45

## What

`VoxrCommandParser.GenerateGrammarJson` emitted the grammar as a flat list of single words. It now also emits **multi-word phrase entries**:

1. **Contiguous runs of required pattern literals**, split at slot boundaries — `close distance`, `switch to navigation`, `orient heading`.
2. **Multi-word slot surface forms** — values and alias keys: `hotel one`, `safe range`.

An **optional literal ends a run** just like a slot does: it may not be spoken, so the words either side of it are not reliably contiguous (`launch ?a {?quantity} …` never yields `launch a`).

The individual words are **deliberately kept** alongside every phrase, so the emitted grammar is a strict superset of the previous one. Nothing that decoded before stops decoding, and an utterance the VAD splits mid-phrase still resolves as fragments. The sequence constraint is therefore a bias, not a hard rule.

Parser, scoring, selection and every consumer-facing API are untouched — only the hypothesis space the decoder searches changes. Demo grammar: 98 → 133 entries (35 phrases).

## Why

A word-list grammar discards every sequence constraint the patterns declare: the decoder may assemble any in-grammar words in any order, so acoustically-near in-grammar neighbours substitute for one another with nothing structural resisting. VOSK charges one language-model transition per grammar entry, so a *k*-word phrase costs one transition where the same words cost *k* — the declared order becomes the cheaper path through the search.

## Validation

- [x] **Unity tests, both platforms** (host project `VoXR TestGround`, Unity 6000.4.7f1, working tree under test): **EditMode 116/116**, **PlayMode 330/330**, `failed="0"` on both NUnit roots. Includes the 5 new grammar tests and `VoxrWavReplayTests.Replay_ManifestCase("audio/tts/switch_to_navigation.wav")`, which now passes against the re-baselined `mode_navigation` expectation.
- [x] **Tier C desktop harness**: 16/16 against the updated baseline (`exit 0`).
- [ ] **On-device (Quest) check needed** — the in-headset A/B the issue calls for, against the documented confusion pairs. Not demonstrated here; see *Not covered*.

**Empirical A/B against the real decoder** (Tier C desktop harness, libvosk 0.3.45, same model, same −18 dB AGC gain, committed 16-fixture corpus):

| | result |
|---|---|
| Before | 16/16 matched the old baseline |
| After | 15 transcripts byte-identical, **1 corrected** |

The corrected one is the substitution class this issue targets:

```
switch_to_navigation.wav
  before: "switch two navigation"   → matched no command
  after : "switch to navigation"    → fires mode_navigation
```

That fixture was pinned in the corpus as a **negative homophone baseline** with the note *"if recognition or the parser improves, this fires and the baseline must be consciously updated."* This PR performs that conscious re-baseline, in both `Tests~/Fixtures/audio/manifest.json` (now expects `mode_navigation`) and `NativeBridge~/harness/expectations.json` (transcript + the grammar pin).

**The issue's open question — must single words also remain? — was measured, not guessed.** A phrases-only variant (dropping the 38 single words that occur only inside phrases) produced *identical* transcripts across all 16 fixtures. It buys nothing measurable while losing fragment decodability, so the words stay.

The grammar generator used for the A/B compiles the **real** `GenerateGrammarJson` source (not a re-implementation); it was validated by reproducing the committed 98-entry grammar pin byte-for-byte before the change was applied.

## Not covered

- Confirm/cancel follow-up vocabulary is flattened to words by its caller (`GetFollowUpGrammarWords` → `AddPhraseWords`) before it reaches the emitter, so multi-word follow-ups (`never mind`, `go ahead`) are **not** phrase-chunked. Same failure class, different code path, left alone to keep this diff scoped — worth its own issue.
- The corpus is TTS speech. The gain against the human-speech confusion pairs the issue documents (close/coast, full/four, clicks/six) is **not** demonstrated here.


jinwoo1601 (commented):
## Independent review — PR #61

**Depth profile:** slim (issue lane) — angles `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. All six ran; none dropped. `GONE` and `XFILE` were terminated mid-run by an API session limit on the first attempt and were **re-run to completion** — findings below come from the completed runs.

**Head reviewed:** `47c8ffd66606fbcbb4ac0ef92e956a32d371bfc8` · **merge-base:** `5f3e9d330a14cdf7b0c320e758e84f0c75fabc2d`

Finder agents inherited no context from the session that wrote the change. Every finding below was then adversarially verified by a separate agent instructed to default to REFUTED. **10 findings survived dedup; 7 confirmed, 3 refuted.** No blockers.

---

### What the review positively confirmed

Worth stating, because it is the PR's central safety claim and two independent angles established it rather than taking it on trust:

- **Invariant "strict superset" holds unconditionally at code level, not just for the demo grammar.** `GONE` traced every base emission path — pattern literals via `run`→`AddPhrase`, slot values/aliases via `AddSurfaceForm`, optional literals, digit vocabulary, `additionalWords`, `[unk]`, ordinal sort, `StringBuilder` emission — and found no input shape where a word the base emitted is dropped. Degenerate elements (`""`, `"?"`, multi-space literals) behave as before.
- **The 133-entry pin is exactly right.** `CLAIM`, `LINE` and `XFILE` each independently re-derived `GenerateGrammarJson` over `DemoGrammar.cs` and reproduced the committed pin: 98 base entries all retained, exactly 35 phrases and no others, zero duplicates, correct `StringComparer.Ordinal` positions. The claimed 38-word phrases-only delta also checks out.
- **No consumer assumes single-word entries.** `XFILE` traced `GrammarManager` → `VoxrSpeechRecogniser.SetGrammar` → `vosk_bridge_set_grammar` / `EditorMicBackend` → `vosk_recognizer_new_grm`: all pass-through, nothing splits on whitespace, counts words, or sizes a buffer by entry count.
- **The re-baselined fixture drives the assertions the PR says it does,** and asserts strictly *more* than the negative pin it replaced (intent **and** transcript, vs. only `Assert.IsEmpty(commands)`).
- **`PIT` returned zero findings** — no aliasing on the shared `run` list, no culture sensitivity, no new JSON-escaping exposure, no allocation reaching the zero-alloc poll path.

---

### Confirmed findings

#### F1 · moderate · `KNOWN_LIMITATIONS.md` historicises a live device repro on evidence that does not cover it
Found independently by `CLAIM` and `GONE`. The entry now reads **"Status: substantially mitigated"** and relabels the repro **"(historical, pre-#45)"**. That repro is *human speech on Quest* (`v2.5 test matrix Phase 4.5`), while the only new evidence in this diff is one **TTS** fixture for a **different** utterance — `switch_to_weapons.wav` was already pinned positive at base and is untouched here. The PR body itself defers exactly this: *"On-device (Quest) check needed"*, *"the gain against the human-speech confusion pairs … is not demonstrated here."*

The verifier killed the obvious defence: the retained caveat says the bias can still surface *"wherever the competing words are not separated by a phrase entry"* — but `switch to weapons` **is** now a phrase entry in the committed pin, so the caveat affirmatively places the historicised repro *outside* the residual-risk set rather than covering it.

Severity moderated from `major` (the unchanged Workaround still steers readers off the broken phrasing, so the harm is misdiagnosis latency, not a functional defect) — but this is the file `CLAUDE.md` designates as the *first* stop when triaging a suspected bug.

**Suggested wording:** keep the repro present-tense and attribute the change only to what actually moved — e.g. *"VOSK has transcribed `switch two weapons` on Quest (v2.5 Phase 4.5, human speech); not re-tested since #45. The TTS fixture for 'switch to navigation', previously pinned negative, now decodes correctly and has been re-baselined."* And drop or qualify the "not separated by a phrase entry" narrowing.

#### F2 · minor · `GrammarJson_SlotBoundaryEndsThePhrase` cannot fail
`Tests~/Runtime/VoxrCommandParserTests.cs` — found independently by `LINE` and `GAP`; the verifier compiled a verbatim port of the emitter and ran the regression to settle it. Both assertions are vacuous:

- `"close distance safe range"` — a slot **value** never enters a pattern run (values only reach `AddSurfaceForm`), so this string is unreachable by construction.
- `"distance target"` — under the regression the entry is `close distance target`; the char before `distance` is a space, not `"`, so the leading-quote anchor never matches.

Delete the slot-branch `AddPhrase(...); run.Clear();` and this test stays green on both assertions.

**The finding's impact claim was refuted, though:** invariant 2 *is* guarded — by the sibling assertion `Assert.IsTrue(json.Contains("\"close distance\""))`, which fails under that regression because the entry becomes `close distance target`. The Tier C pin is a second independent guard.

**Residual real gap the verifier found:** keep `AddPhrase` at the slot branch but drop only `run.Clear()`. That emits *both* `"close distance"` and `"close distance target"` — every assertion in the suite passes and a slot-spanning phrase ships. `Assert.IsFalse(json.Contains("\"close distance target\""))` would catch both variants.

#### F3 · minor · The `DemoGrammar` sync obligation does not name the harness grammar pin
`Tests~/Runtime/DemoGrammar.cs` records *"If the demo grammar changes, update this class AND the fixture manifest together"* — two artifacts. This change makes `NativeBridge~/harness/expectations.json`'s `grammar` pin sensitive to **pattern shape**, not just vocabulary, and nothing under `Tests~` compares the emitter's output to that pin.

Concrete live scenario: applying the #42 authoring remedy (`by` → `?by`) leaves the word set identical — valid under the base emitter — while silently deleting a phrase entry under the new one. The verifier added the amplifier: **`--write-baseline` rewrites only `expectedFinals`, never the grammar pin**, so re-baselining under a stale pin re-records transcripts decoded against the *stale* grammar and entrenches the divergence instead of exposing it.

Severity kept at `minor`: the obligation is pre-existing and this PR widens its trigger surface rather than creating it. Note `DemoGrammar.cs` is **not** in this diff — the change makes untouched text less complete.

#### F4 · nit · Stale sibling description in the fixture manifest
`Tests~/Fixtures/audio/manifest.json` — `navigation_mode.wav` still reads *"(its 'switch to navigation' phrasing is the negative homophone fixture)"*, which this same commit made false. Found by three angles. Telling detail: the author *did* sweep the analogous text in `expectations.json` (`homophone-negative` → `silence`), so this is a miss inside an otherwise intentional sweep. The description string is echoed into PlayMode failure messages. Adjacent: `"Positive counterpart: switch_to_weapons.wav"` is now vacuous, since both are positive.

#### F5 · nit · The decoder mechanism is stated as established fact
`Documentation~/command-recognition.md` and `KNOWN_LIMITATIONS.md` assert *"VOSK charges one language-model transition per entry, so a three-word entry costs one transition where the same three words cost three."* Nothing in the repo derives this; the vendored `vosk_api.h` documents only *"the list of phrases to recognize as JSON array of strings"* — establishing that multi-word entries are accepted, and nothing about cost. The claim appears in four files, all this PR's own.

The verifier's substantive point: Vosk estimates an **n-gram LM over the entries treated as sentences**, and the decoding graph still traverses **one arc per word** — so "a three-word entry costs one transition" is at best a metaphor for a boundary/backoff cost difference, not a description of what the decoder does. The observed effect is measured and real; the stated explanation may be literally wrong. Note the CHANGELOG labels the fixture delta *"Measured against the real decoder rather than assumed"* immediately after asserting the unmeasured mechanism.

**Suggested fix (framing only):** lead with the observed effect, demote the transition-cost model to an offered explanation. The actionable guidance ("keep function words inside a run of required literals") stands on the effect alone.

#### F6 · nit · The ordinal-sort half of the determinism invariant is untested
Swapping `sorted.Sort(StringComparer.Ordinal)` for `sorted.Sort()` is invisible to the entire unit suite — the only `AreEqual` on a grammar string is the one-element `["[unk]"]` case, which pins no order. Only the manually-run, vendor-gated Tier C pin would catch it.

*Rationale partly refuted:* the finding argued the sort became load-bearing only now, citing `torpedo range` vs `torpedoes`. The verifier measured that pair on .NET 10/ICU (`ord=-1, invCulture=-1, curCulture=-1` — no reversal), and showed the base pin was **already** comparer-dependent (`"[unk]"` sorts ahead of `"a"` only because `[` = 0x5B < `a` = 0x61). Real coverage gap, but not a new one.

#### F7 · nit · The multi-word required-literal element path is unexercised
No pattern in any test grammar, nor `DemoGrammar`, nor the shipped sample contains a multi-word literal *element*, so the `element.Split(' ')` merge path has no coverage.

*Rationale refuted:* the finding claimed the issue's motivating examples (`close to`, `on track`) are that shape — they are not; those are *emitted run entries*, produced from single-word elements, and that path is well covered. The verifier also showed the asset-authoring path **cannot** produce a multi-word element (`VoxrAssetConversionTests` pins `"fire weapons"` → `{"fire","weapons"}`); it is reachable only from hand-written C# patterns, and the base emitter split identically.

---

### Refuted findings

Named so you can see what was considered and killed:

- **CHANGELOG "nothing that decoded before stops decoding" is contradictory** — REFUTED. The clause is scoped by its own premise and re-hedged in the same sentence (*"the sequence constraint is a bias rather than a hard rule"*), with the measured counter-example one sentence later. `switch two navigation` remains legal and decodable; it stopped *winning*.
- **New docs section misleads about confirm/cancel phrases** — REFUTED. The new section is scoped by its own enumeration (pattern literal runs, slot values/aliases) and never mentions follow-up vocabulary; the sentence it was paired with is pre-existing and untouched. At base the page said nothing about the grammar's unit, so the addition *reduces* the false-premise surface.
- **Re-baselining deleted the corpus's only non-silence negative case** — REFUTED. That class is pinned at parser level by three existing tests (notably the #31 `"target disengage"` → below-`minScore` case, verbatim the alleged failure scenario). `VoxrFixtureCorpusTests` documents that homophone traps are negative *only while the decoder loses* — a recorded defect, not a designed coverage slot — so retention was impossible without keeping the bug. The new pin is strictly stronger. Survives only as F4.

---

Stated so none of these reads as a passing check:

- **On-device (Quest) behaviour** — the in-headset A/B against close/coast, full/four, clicks/six. The PR defers this explicitly; **F1 is precisely where the docs outrun it.**
- **The ticked Unity counts** (EditMode 116/116, PlayMode 330/330) were not re-run here — treated as claims. No existing assertion in the repo contradicts the change. One angle noted the PlayMode transcript assertion runs against the Windows VOSK build while the pinned transcript came from Linux libvosk 0.3.45.
- **The Tier C 16/16 result** was not re-run (needs the gitignored `vendor/` libvosk + model) — but the committed pin was independently re-derived from source by three agents and is internally consistent.
- **Actual VOSK decoder semantics** — see F5.
- No CI runs the Unity suites (`.github/workflows/` holds only `release.yml`), so every guard discussed above is local-manual.

---

Per the issue lane: no confirmed blocker, so **nothing here has been fixed or ruled on** — all seven findings are left for your decision.

jinwoo1601:
Pushed `c97540d`, addressing **F1** and **F2** from the review above. The remaining five findings (F3–F7) are untouched and still awaiting your ruling.

**F1** — `KNOWN_LIMITATIONS.md` no longer historicises the repro. Status is now "still open", stating exactly what was measured (one TTS fixture, re-baselined) and marking the human-speech repro present-tense and not re-tested on device. The caveat that scoped residual risk to "wherever the competing words are not separated by a phrase entry" is gone — as the verifier pointed out, `switch to weapons` *is* now a phrase entry, so that wording disclaimed the very repro it should have covered.

**F2** — `GrammarJson_SlotBoundaryEndsThePhrase` now asserts absence of `"close distance target"` / `"set distance target"` — the welds the boundary actually prevents — plus presence of the runs either side.

I verified the fix discriminates rather than assuming it, by generating the test grammar from the real emitter under both defect variants:

| | old assertions fail? | new assertions fail? |
|---|---|---|
| correct code | no | no |
| variant A — drop slot-branch flush **and** reset | **no** | **yes** |
| variant B — drop only the reset | **no** | **yes** |

Variant B is the residual gap the verifier identified: it still emits `"close distance"`, so the sibling assertion stays green and it previously escaped the entire suite. Both variants are now caught.

Unity re-run after the fixes: **EditMode 116/116, PlayMode 330/330**, all five phrase tests passing. No production code changed, and the counts match the earlier run exactly.

PR #62, opened 2026-08-12
Document the NumberSequence slot contract at every point of use (#46)

### LINKED ISSUE #46: NumberSequence GetSlot contract (spoken words, not digits) is undocumented at every point of use
## Summary

For a `NumberSequence` slot, `VoxrCommand.GetSlot()` returns the **raw spoken digit words** ("two seven zero"), never a numeric string — every consumer must run the value through `VoxrNumberParser` themselves. This contract is currently stated nowhere a consumer actually looks when wiring a slot: not on `GetSlot`/`GetSlot`'s XML docs, not on `VoxrSlotType.NumberSequence`, not in the command-recognition guide's slot section, and not demonstrated in the CommandRecognition sample's handler code. `Documentation~/api/number-parser.md` documents the parser itself, but nothing routes a `GetSlot` caller to it.

Until it's documented at the point of use, it's tribal knowledge — and it bites.

## What it cost in practice (one consumer project, one month)

- A handler did the obvious thing — `int.TryParse(cmd.GetSlot("track"))` — which silently fails on every word-string. Result: **every numeric voice command in the scene was dead** ("Say again" on all of orient/intercept/adjust_range) while number-free commands worked, shipping through review and surfacing only in-headset. The failure mode is invisible to headless/EditMode tests because test doubles construct slot values directly and never reproduce the word-string shape.
- The digit/cardinal conversion was **independently reimplemented twice** in the same project (two different scripts, with *different* digit-first/cardinal-first fallback orderings) before anyone discovered the package already ships `VoxrNumberParser` — the utility was found by reading `Library/PackageCache` source, not docs.
- Both `ParseDigitSequence` and `ParseCardinal` throw `FormatException`, so every consumer call site grows nested try/catch; the docs should show the intended fallback chain (digit-sequence first, cardinal second) as the canonical pattern, including the known ambiguity that "two seventy" parses as 72 via the cardinal path, not 270.

## Suggested doc changes

1. **XML doc comment on `GetSlot`/`HasSlot`** (and `VoxrSlotMatch.Value`): one sentence — "NumberSequence slot values are the spoken digit words (`\"two seven zero\"`), not a numeric string; convert with `VoxrNumberParser`."
2. **`VoxrSlotType.NumberSequence` enum doc + slot section of the command-recognition guide**: state the contract and show the canonical conversion snippet (digit-sequence → cardinal fallback, with the try/catch shape it currently requires).
3. **`api/data-types.md`**: the `VoxrCommand` slot table should carry the same note.
4. **CommandRecognition sample**: have the sample handler actually convert a NumberSequence slot via `VoxrNumberParser`, so copy-paste consumers inherit the correct pattern instead of `int.TryParse`.
5. Cross-link `api/number-parser.md` from all of the above.

A typed accessor on `VoxrCommand` (e.g. `TryGetSlotInt`) would remove the footgun outright and is worth considering separately; this issue is only about making the existing contract discoverable, which is cheap and immediately closes the trap for new consumers.

jinwoo1601:
Closes #46

## What

Makes the `NumberSequence` slot contract — `GetSlot` returns the **words as spoken**, never a numeric string — visible everywhere a consumer meets it.

- **XML doc comments** (new) on `VoxrCommand.GetSlot`, `VoxrCommand.HasSlot`, `VoxrSlotMatch.Value`, and both `VoxrSlotType` members, so the contract reaches the IDE tooltip at the call site rather than living only in a doc page.
- **Corrected three docs that taught the wrong shape** by writing the *converted* value as if it were the slot value: `// "heading two seven zero" -> heading="270"` in the command-recognition guide, the same claim in `api/command-definitions.md`, and `heading = 270` in the sample's "Try saying" table. The root `README.md` feature bullet is disambiguated the same way.
- **Canonical conversion snippet** in the guide's *NumberSequence Slots* section: `ParseDigitSequence` first, `ParseCardinal` as fallback, both in `try`/`catch` since they throw `FormatException` rather than returning a sentinel — plus why the order is load-bearing (`"two seven zero"` is `270` on the digit path but `9` on the cardinal one) and the `"two seventy"` → `72` trap.
- **Cross-links to `api/number-parser.md`** from `api/data-types.md`, `api/command-definitions.md`, the command-recognition guide, and the `CommandRecognition` sample README.
- **The `CommandRecognition` sample now handles its own `set_heading` intent** through that exact helper (`CommandDemo.TryParseNumberSlot`), instead of leaving the one slot type that needs conversion undemonstrated — so copy-paste consumers inherit the correct pattern.
- `api/command-definitions.md`'s `VoxrSlotType` table also had the accepted vocabulary wrong (`"zero"` through `"nine"`); it is the full `DigitVocabulary` (zero–nineteen, the tens, `hundred`, `thousand`), which is what decides where a slot value ends.

Documentation, XML comments, and the sample only — **no runtime behaviour changes**. The typed accessor the issue floats (`TryGetSlotInt`) is deliberately not included; the issue scopes itself to discoverability.

## Why

`int.TryParse(cmd.GetSlot("track"))` is the obvious thing to write and fails on every word-string — silently, since `TryParse` does not throw, so it yields `0`. That presents as every numeric voice command being dead while number-free ones work, and it is invisible to headless tests because test doubles construct slot values directly. Per the issue, one consumer project lost a month to this and reimplemented the digit/cardinal conversion twice before finding that `VoxrNumberParser` ships with the package.

## Validation

- [x] **Unity EditMode: 116/116 passed**, 0 failed / 0 inconclusive / 0 skipped (`start-time 2026-08-12 04:07:31Z`).
- [x] **Unity PlayMode: 330/330 passed**, 0 failed / 0 inconclusive / 0 skipped (`start-time 2026-08-12 04:08:10Z`). Host project `VoXR TestGround`, Unity 6000.4.7f1, run against this branch's tree.
- [x] Standalone `dotnet` compile of the edited runtime files plus the edited sample (with Unity/recogniser stubs), documentation generation on: **0 errors**, and no `CS1570`/`CS1574` — every new `<see cref="...">` resolves. The sample is not compiled by any Unity check here, since `Samples~` is only built once a user imports it.
- [x] Parser behaviour claims in the new prose verified against `Runtime/Commands/VoxrNumberParser.cs`: `ParseCardinal("two seven zero") == 9`, `ParseCardinal("two seventy") == 72`, `ParseDigitSequence` throws for any token above `nine`.
- [x] Independent review run (`review-pr`, slim profile, 6 angles) — 10 findings verified, 6 CONFIRMED / 4 REFUTED. Five confirmed findings were statements this PR introduced and are fixed in `d1bdec9`; see the posted review. The Unity suite numbers above were taken at `4bbd128` and still hold, since the fixes add no executable change (the `Runtime/` diff is `///` lines only).
- [ ] **Not needed:** no on-device (Quest) check — nothing that runs on device changed.


jinwoo1601 (commented):
## Independent review — PR #62

**Head reviewed:** `4bbd12878081ed25f567574e6bd290529e130e45` (base merge-base `c14690a`).
**Method:** six finder agents against the diff with no context from the session that wrote it, deduped, then one adversarial verifier per surviving finding. **10 findings verified: 6 CONFIRMED, 4 REFUTED.**

Five of the six confirmed findings were statements *this PR introduced* — a documentation PR shipping false documentation is self-defeating, so they were fixed on this branch in `d1bdec9`. The sixth is left for you to rule on.

---

### CONFIRMED — fixed in `d1bdec9`

**F1 · `"as spoken"` was written as an absolute, and it is false for `Enumerated` slots** — *moderate*
`Runtime/Commands/VoxrCommand.cs:16` said "The matched value, exactly as spoken", `:63` "The value is the words as spoken", and `Documentation~/api/data-types.md:50,64` the same. But `VoxrCommandParser.cs:1041` returns `entry.CanonicalValue`, so a spoken alias arrives resolved — and the repo has a test named exactly `Alias_ResolvesToCanonicalValue` (`Tests~/Runtime/VoxrCommandParserTests.cs:424-432`) asserting `GetSlot("weapon") == "jackal"` for spoken *"shoot jackals"*.

Two things made this worse than a loose phrase. The `data-types.md:64` row attached "as spoken" to an **Enumerated** example (`"missiles"`) in the `weapon` slot that carries the `jackals → jackal` alias in both the sample and the test grammar; and the base text there was neutral ("Matched value (e.g. `"missiles"`)"), so the diff *introduced* the falsehood. This PR exists to stop consumers writing comparisons that silently never match — shipping a new false claim about the same accessor would have reproduced the bug class in the fix. Now scoped to `NumberSequence`, with the `Enumerated`/canonical behaviour stated.

**F2 · the cross-link claimed for "the sample READMEs" did not exist** — *minor*
Both the PR body and the shipped `CHANGELOG.md` entry claimed `api/number-parser.md` was cross-linked from "the sample READMEs". `grep -rn "number-parser" Samples~/` returned nothing; the added mentions were backticked identifiers, not links, and the sample README's See Also was untouched. The plural was also unearned — only one of four sample READMEs was touched. The verifier specifically tested whether the CHANGELOG's grammar could distribute "state it" over four subjects but "cross-link" over three, and concluded it cannot. Fixed by adding the link and naming the one README.

**F3 · `command-definitions.md` kept "digit words" after this PR deleted the term's only definition** — *minor, narrowed*
The diff corrected the `VoxrSlotType` row at `:79` to the full `DigitVocabulary`, which meant deleting the file's own gloss ("zero" through "nine") — while `:49`, `:52`, `:53` still said "digit words" 30 lines above. That is materially wrong, not just loose: `VoxrCommandParser.cs:1058-1063` increments once per `DigitVocabulary` token, so `NumberSequence("code", minWords: 3, maxWords: 3)` equally accepts `"seventeen twelve nineteen"`. Fixed, and the `MinWords`/`MaxWords` rows now say they count vocabulary words rather than digits.
*Narrowed by the verifier:* `Documentation~/index.md:26` was refuted as a site (it already carries the link, and describes the parser class rather than the slot contract).

**F4 · "throw rather than returning a sentinel" was unqualified, and untrue for the empty string** — *minor*
Both parsers open with `if (string.IsNullOrWhiteSpace(words)) return 0;` (`VoxrNumberParser.cs:31,50`) — which the linked page states two rows above the claim ("Returns `0` for null/empty"), so the new prose contradicted its own cross-link target. It matters here: `GetSlot` returns `""` for an unmatched slot with no warning, and the shipped sample's optional `{?elevation}` hits that path on every "orient heading X" utterance, so a reader who took "always throws" literally and dropped the guard would log `mark 0` for speech containing no elevation. Fixed in the guide, the sample comment and the CHANGELOG; the guide now says why the empty-string guard is load-bearing.

**F5 · non-compiling `int.TryParse` fragment** — *nit*
`data-types.md:53` wrote `int.TryParse(cmd.GetSlot("heading"))` — no single-argument overload exists in any .NET version (the verifier compiled it: `CS1501`). Fixed to match the phrasing used at the other three sites.

---

### CONFIRMED — left for your ruling

**F8 · the fallback-order rationale is pinned by no test** — *minor, durability*
The new prose makes two specific readings load-bearing across four documents: `"two seven zero"` is `270` on the digit path but `9` on the cardinal one (the stated reason digit-first is correct), and `"two seventy"` comes back `72`. Both are **numerically correct today** — verified against source — so there is no live defect. But `grep -rn "seventy" Tests~/` returns nothing, and the nearest tests (`ParseCardinal("forty two") == 42`, `ParseCardinal("three hundred fifteen") == 315`) exercise different shapes: an edit that concatenated adjacent unit-digit words — the natural "make cardinal handle digit dictation" change — would leave all 16 parser tests green while falsifying the rationale in four documents.

Two assertions in `Tests~/Runtime/VoxrNumberParserTests.cs` would close it. I did not add them unprompted: it is additive scope beyond issue #46, and it would move the PlayMode count off the 330 recorded below, requiring a re-run. Your call.

---

### REFUTED — considered and killed

- **`Depends: (none)` headers vs the new `<see cref>` references.** Surveyed all 39 `Depends:` headers under `Runtime/`: every entry resolves to a real code token, and `Runtime/Native/BridgeNative.cs` names a sibling type in a comment while keeping `Depends: (none)`. The field tracks code dependencies; doc crefs are not one.
- **`GetSlot`'s "logs a warning" is unconditional prose over a `#if DEBUG` diagnostic.** True, but byte-identical at base and nothing the diff added contradicts it — pre-existing, out of scope here. Clean standalone issue candidate; it is the only occurrence in `Documentation~/`.
- **`VoxrSlotDefinition.NumberSequence()` factory lacks an XML doc.** The factory is a *declaration* site; the documented contract is about what `GetSlot` *returns*, and every read site now carries it. The factory also appears in none of issue #46's five enumerated suggestions, and both doc pages already annotate the factory example inline.
- **`VoxrSlotAsset` fields lack `[Tooltip]`.** The finding's premise ("every neighbouring field has one") is false — `AliasEntry.variant`/`canonical` have none. A `[Tooltip]` on `slotType` also renders regardless of the selected enum value, so it would not produce the claimed effect, and the file is not in the diff.

### Not verifiable from here

- **On-device (Quest):** not needed, and I agree with the PR's box — `git diff` over `Runtime/` is `///` lines only, zero IL delta, so there is nothing for a device to exercise.
- **Re-running the Unity suites:** not re-run after `d1bdec9`. The recorded **EditMode 116/116** and **PlayMode 330/330** (both `0 failed / 0 inconclusive / 0 skipped`, fresh `start-time` stamps of `2026-08-12 04:07:31Z` / `04:08:10Z`) were taken against `4bbd128`, whose tree is byte-identical to `main` plus this PR. The fixes in `d1bdec9` add no executable change — verified: filtering the `Runtime/` diff for non-`///` lines returns nothing — so the counts stand. Independently, a finder reconciled both totals against the suite source (EditMode 116 = 107 `[Test]` + 9 `[TestCase]`).
- **IDE tooltip rendering:** cref resolvability was checked by compiling with documentation generation on (0 errors, no `CS1570`/`CS1574`, re-run after the fixes); the rendered tooltip itself was not seen.

### Worth noting

The contract this PR documents is already test-locked — `VoxrCommandParserTests.cs:605` asserts `GetSlot("heading") == "two seven zero"`, and the fixture manifest pins the same value — so a *code* regression is caught. A *doc* regression is not; the repo has no docs lint or link check. Out of scope for this lane, recorded rather than filed.

Two surfaces the sweep did not reach, both outside issue #46's five enumerated suggestions and both flagged by verifiers as separate follow-up candidates rather than defects here: `Documentation~/inspector-authoring.md:28,31` and `Documentation~/api/scriptable-objects.md:14,17,18` still say "digit words" and carry neither the value-shape note nor a `number-parser.md` link. That is the zero-code authoring path, where no IDE tooltip reaches the author.

*Comment review only — not an approval. I do not merge.*

PR #63, opened 2026-08-12
Document the matching and scoring model (#47)

### LINKED ISSUE #47: Document the matching/scoring model — score formula, penalties, selection order, gates, eager-flush verdicts

jinwoo1601:
Closes #47

## What

Adds **`Documentation~/scoring.md`**, a reference for the model that decides whether a spoken command fires:

1. **The score formula** — the per-element contribution table (matched required literal/slot `+1.0`, matched optional literal `+0.5`, missed required literal `−0.5`, missed required slot `−1.0`) and the dynamic denominator, with the consequence spelled out: a miss is charged twice (withholds credit *and* subtracts a penalty while still occupying the denominator), so one dropped function word scores `1.5 / 3` = `0.50` on a three-element pattern and fails the default gate, but `5.5 / 7` = `0.79` on a seven-element one and passes.
2. **The skipped-word penalty** and its four rules — applied *after* selection (so it filters via `minScore` without changing which pattern wins), `[unk]` never charged, counting restarts per extracted command, proportional.
3. **Selection and tie-break order** — earliest start → score → consumed span → literal count → registration order, plus why the span term sits above literal count and that `TryEagerCommit` mirrors the same ordering.
4. **Sequential extraction** and what it means for patterns that prefix others.
5. **The two gates** — that confidence is the *minimum* per-word value over the matched span, not an average (a `1.00` match rejected at `0.39`), the ~0.5 ceiling on "two", the `-1` "no data" sentinel, joint-tuning guidance, and the post-gate filters (debounce, `requiresConfirmation`, `allowPartialMatch`) with their exact `rejectReason` strings.
6. **Eager-flush verdicts** — `Commit` / `HoldExtendable` / `None`, the four conditions a verdict above `None` requires, terminality, and the `MaxOptionalExpansion` guard.
7. **Three worked examples** traced candidate scoring → selection → gates, each ending in the session-log entry it produces, followed by a field-to-rule map and a symptom table for the three most common rejections.

**Cross-linked** from `index.md`, `command-recognition.md` (a pointer at the head of *Scored Matching* plus See Also), `api/command-recogniser.md` (the `minScore`/`minConfidence` rows), `troubleshooting.md` (two new symptom entries: a command that fires with a slot silently missing, and the `~0.50` short-pattern rejection), and the session log's own `readme` field.

**One code edit**, the `readme` string in `Editor/VoxrDebugSessionLog.cs`: it gains the pointer to `scoring.md`, and its description of an `attempts` entry is corrected. It said "each attempt within it is one command pattern evaluated against that utterance" — but `attempts` holds one entry per *extraction round*, reporting the pattern that won selection; losing candidates are never recorded. A reader would have concluded the opposite of the truth from the old wording. `editor-testing.md` carried the same error in two places and is corrected too. `schemaVersion` is unchanged, because the JSON shape is not.

## Why

The scoring model existed only as `internal const`s in `VoxrCommandParser` and rules spread through its selection loop, and each release since 1.2.1 added another interacting knob. Since 1.3.0 the auto-exported session log surfaces per-attempt `score`, `minScore`, `aggregateConfidence`, and `rejectReason` for every utterance — exactly the right diagnostic data, but uninterpretable without the formula behind it. A consumer diagnosing a rejection, tuning the thresholds, or authoring patterns that must not shadow each other had to reverse-engineer `TryMatchScored` to do it.

## Validation

- [x] **Every number in the doc was measured, not derived by hand.** Built a scratch rig (per the established grammar-A/B pattern) that compiles the *real* parser sources with `dotnet` against a `UnityEngine.Debug` stub, then drove `TryMatchScored`, `Parse`, and `TryEagerCommit` directly. Confirmed: `1.5 / 3` = `0.50` and `5.5 / 7` = `0.7857` for one dropped required literal; `thrusters report` → `1.00` unpenalised → `0.50` final; `[unk] [unk] report` → `1.00` (filler free); `launch launch …` → `4 / (4 + 1)` = `0.80`; confidence = `0.39` (the minimum, from "two") on an otherwise-perfect `1.00` match, and `-1` with no word data; the span tie-break electing the tailed pattern over a bare sibling **declared first**; sequential extraction yielding two commands at `1.00` each with no cross-penalty; and all five eager verdicts in the §6 table (`fire` → `HoldExtendable`, `fire at` → `None`, `fire at hotel one` → `Commit`, leading `[unk]` → `Commit`, trailing `[unk]` → `None`).
- [x] All 40+ intra-doc links and heading anchors across `Documentation~/` + `KNOWN_LIMITATIONS.md` resolve (scripted check: 0 broken).
- [x] The edited `readme` string literal compiles and renders as intended (extracted and built standalone).
- [ ] **Unity test suite not run.** No runtime or parser logic changed — the only code edit is an Editor-only `const string` with no test asserting its content — so the two-platform host-project run would exercise nothing this PR touches. Say the word if you want it run anyway.
- [ ] No on-device (Quest) check applicable: documentation plus one Editor-only string.


jinwoo1601 (commented):
## Independent review — PR #63

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Reviewed at:** `b12ad41dd3cf3da6f74fe53c24f40ea42afd70d0`, against merge-base `66b1855`.
**Method:** six finder agents with no context from the session that wrote the code, deduped to 15 findings, then one adversarial verifier per finding. **13 CONFIRMED, 2 REFUTED.**

Because this is a documentation PR, the brief instructed every angle to treat a prose claim that contradicts the source as a correctness defect at the same severity as a wrong line of code, and to re-derive every number from `VoxrCommandParser.cs` / `VoxrCommandRecogniser.cs` rather than check the document against itself.

**The PR's own claims held up.** All seven of issue #47's enumerated coverage items land; all three named cross-link sites exist and resolve; the "every number was measured" box survived independent re-derivation of the whole §1/§2/§3/§6 arithmetic; the 0-broken-links box reproduced. The deferral of the Unity suite is accurate — the only code edit really is an Editor-only `const string`, `Editor/Jinwoo1601.VoXR.Editor.asmdef` declares `"includePlatforms": ["Editor"]`, and no test under `Tests~/` asserts that string's content.

**What the review nevertheless found is that the new page asserted several things the source does not do.** Nine of those are fixed on this branch in `49647c9`; four remain open for your ruling.

---

### Fixed on this branch (`49647c9`)

**F1 · major · CONFIRMED — `OnUnrecognisedSpeech` was documented backwards.**
`scoring.md:174` said the event "fires only when nothing matched at all — not when a match was filtered by confidence or debounce." `anyThresholdFiltered = true` is set at exactly two sites, `VoxrCommandRecogniser.cs:701` (confidence) and `:713` (debounce); the `minScore` rejection at `:668-696` and both pending diversions `continue` without touching it, so `:755`'s `if (!anyThresholdFiltered)` **fires** for them. A committed test already pinned this — `Tests~/Runtime/VoxrPendingCommandTests.cs:73-87` asserts `unrecognised` is non-null for a sub-`minScore` match. Replaced with a five-row table. Note two *pre-existing* pages stated the mirror-image error (`command-recognition.md:426`, `api/command-recogniser.md:42`) — **no page in the repo stated the actual rule**; both are corrected here.

**F2 · major · CONFIRMED — the correction sweep missed the highest-traffic instance.**
`troubleshooting.md:48` still read "the full match breakdown: which patterns were tried, what score each received" — the exact misconception this PR corrects in `editor-testing.md` and the `readme` const — sitting *two lines above the new bullet this PR appends to the same list*. `VoxrCommandParser.cs:671` records `ParseDiagnosticEntry` only for `best*` after the winner is chosen, and `Editor/VoxrDebugWindow.cs:365` renders that same array; there is no second data source. (The verifier refuted the finder's extension of this to line 49, which is fine as written.)

**F3 · major · CONFIRMED — "one entry per extraction round" replaced one over-general rule with another.**
Four paths short-circuit before the parse and publish a synthetic attempt that is not a round and names no pattern: `"no match"` (`VoxrCommandRecogniser.cs:645`, pinned by `Tests~/Editor/VoxrDebugSessionLogTests.cs:76-87`), confirm/cancel (`:577`), follow-up slot-fill (`:626`), and pending timeout (`:450`, which is not an utterance at all — its `inputText` is the *original* command's transcript). All reach the log: the `LastMatchDiagnostics` setter fires `DiagnosticsPublished` unconditionally (`:118`) and `OnDiagnosticsPublished` guards only on test-run/not-playing. Now tabulated, with the tell that an empty `pattern` is what distinguishes them. The `"no match"` entry also carries `aggregateConfidence: 0`, **not** the `-1` sentinel — worth knowing before you write a log-parsing script.

**F4 · major · CONFIRMED — the page contradicted itself about "two".**
Worked example C gave "two" a confidence of `0.39` and cited the ≈0.50 ceiling as the explanation — but a 0.50 floor explains a *pass* at the 0.4 gate, never a fail, and `KNOWN_LIMITATIONS.md:131` says in terms that "the default `minConfidence=0.4` accommodates this". Issue #47 reports the real observation as a weak *function word* at 0.39 and the ~0.5 ceiling for "two" as two separate facts; the page had fused them. The weak word is now an ordinary literal, with "two" left at 0.50 and passing — which sharpens the lesson, since the culprit turns out not to be the word you would suspect.

**F5 + F7 · minor · CONFIRMED — `-1` was attributed to an unreachable cause.**
"A span that is entirely `[unk]`" cannot happen: both selection loops skip `[unk]` start positions (`VoxrCommandParser.cs:570`, `:1309`), so a winning span always contains a matched non-`[unk]` token. The reachable second cause is a matched span with no word data of its own — `UtteranceBuffer.Append` appends text unconditionally but words only when a result supplies them (`UtteranceBuffer.cs:22-32`), so a buffer merging a spoken result with an injected one can match on the half that has none, **with `words[]` populated**. That also falsifies the `readme`'s retained promise that `-1` implies an empty `words` array. Both corrected. The same wrong cause was in pre-existing `troubleshooting.md:76` and `command-recognition.md:143`; corrected too.

**F8 · major · CONFIRMED (severity raised by the verifier) — the confidence definition was incomplete in a way that flips the gate.**
`InstanceBuildWordConfidence` keys the table by word **text**, first occurrence winning (`VoxrCommandParser.cs:1123`), and `ComputeConfidence` looks each span token up by text (`:1153`) — never by position. So a repeated word is scored at its first occurrence *everywhere*, including when that occurrence lies outside the matched span. Measured on the real parser: `orient heading two two zero` with per-word `.94 .88 .85 .39 .97` reports **0.85**, not the 0.39 that is genuinely inside the span — it fires at the default gate where the documented rule predicts rejection. The inverse also holds: a weak word *before* the match can drag the reported value down. Documented, with the `NumberSequence` case flagged as where it bites.

**F11 · nit · CONFIRMED — `rejectReason` strings are locale-formatted.**
`$"score {cmd.Score:F2} < minScore {minScore:F2}"` and friends use `CurrentCulture`; `grep -rn "InvariantCulture\|CultureInfo" Runtime/ Editor/` returns nothing, and the verifier confirmed the Unity Editor installs no culture override either. On a comma-decimal locale the field reads `score 0,50 < minScore 0,60`, so the page's advice to match the quoted literal fails. Noted in the page and in the `readme`. (The numeric JSON *fields* are unaffected — `JsonUtility` writes numbers invariantly.)

**F12 · nit · CONFIRMED — the symptom table was labelled "rejections"; its third row is an acceptance** (`accepted: true` with an empty `slots` array — precisely §7 B's point that this failure produces no `rejectReason` at all). The reader who most needs that row was being told to look elsewhere. Also corrected in `troubleshooting.md:50`, which called all three worked examples "rejections" when example A ends in acceptance.

---

### Open for your ruling — not fixed

**F6 · minor · CONFIRMED — `KNOWN_LIMITATIONS.md:382-383` says the parser "cannot compute a meaningful average".**
It computes a minimum and always has (`ComputeConfidence` has never contained a sum or a division, back to the commit that introduced it). This PR makes "minimum, not average" a headline claim and links into that file twice from the same section, so the two now contradict. One word, in a file this PR does not otherwise touch — left for you rather than widened into.

**F9 · nit · CONFIRMED (downgraded from minor) — `scoring.md` is absent from the README's documentation list**, where it is now the only `index.md` Guides page missing. Weakened by the README list being a demonstrable curated subset (`native-bridge.md` has never appeared in it) and by "For full documentation, see the documentation index" sitting directly above. Strengthened by commit `7cb2ca5`, which edited that list purely to keep it level with `index.md`. Your call whether the list is a mirror or a curation.

**F10 · minor · CONFIRMED — nothing in the repo pins the arithmetic this page now advertises.**
The verifier's mutation argument is the decisive part: set `RequiredLiteralMissPenalty` to `0f` and the doc's headline example scores `2/3 = 0.667`, **passing** the default gate, so both the "rejected" row and the new troubleshooting entry become false — yet the whole suite still passes, because every existing score assertion is `1.0`, `0.5`, or `5/6`, none of which involves a missed required literal. The repo has a precedent for pinning documented tradeoffs (`Tests~/Runtime/VoxrCommandParserTests.cs:1531-1536`, "Pinned so the tradeoff is a known, tested consequence rather than a surprise"). The proportionate remedy is one assertion pinning `1.5/3 = 0.50` with a comment naming `scoring.md`, or at minimum a back-pointer comment at the constants (`VoxrCommandParser.cs:32-35`, where `RequiredLiteralMissPenalty` carries no comment at all while its neighbours carry rationale + issue numbers). **Not done here because a new test cannot be validated without the host-project Unity run this PR defers** — adding an unrun test would be worse than naming the gap.

**F13 · nit · CONFIRMED — fixed in `49647c9`** (the CHANGELOG said the page gives "the four conditions each [verdict] requires"; four gate any verdict above `None`, and `Commit` needs a fifth). Listed here only so the count reconciles.

---

- **`api/data-types.md` should link to the new page.** REFUTED: `data-types.md:95` already links "Command Recognition — how matching works", and this PR planted the pointer at the top of that page's matching section, so the route from the `VoxrCommand.Score` definition site is one hop. The issue named three cross-link sites; all three were delivered.
- **The `readme` correction belongs under `### Fixed` rather than `### Added`.** REFUTED on this file's own practice: `#46` was documentation-only, corrected three places that "actively taught the wrong shape", and sits under `### Changed`; every issue number in the file appears exactly once, so splitting one issue across two sections would be unprecedented here.

---

- **The Unity test suite was not run**, for this review or the PR. The deferral's stated reason checks out, and no existing test asserts the changed string. But note F1 and F3 were each independently pinned by tests that already exist and pass — the errors were in prose, which no suite can catch.
- **On-device (Quest) behaviour** — not applicable to this diff, named so nothing reads as passing.
- **The locale claim (F11)** rests on the absence of any culture override plus .NET formatting semantics; the verifier could not launch the Editor under a comma-decimal locale from WSL.
- **Rendered Markdown appearance on GitHub.** Structural validity was checked (0 broken links or anchors across `Documentation~/`, `KNOWN_LIMITATIONS.md`, `README.md`, `CHANGELOG.md`, both before and after the fixes).

Every numeric claim changed by the fixes was re-measured by driving the real parser sources: example C now reports `aggregateConfidence 0.39` with a slot confidence of `0.50`, and the repeated-word case reports `0.85`.

Nothing here has been merged, approved, or closed — F6, F9 and F10 are yours to rule on.

jinwoo1601:
Pushed `49647c9` addressing F1, F2, F3, F4, F5, F7, F8, F11, F12, F13 from the review above. F6, F9 and F10 are left open for your ruling.

jinwoo1601:
Pushed `1a57955` addressing **F6** and **F9**.

**F6** — `KNOWN_LIMITATIONS.md`'s `-1.00` entry carried two errors, not one, and they were interleaved in the same two sentences, so both are corrected together:

- "the parser cannot compute a meaningful average" → the aggregate is the **minimum**, which `ComputeConfidence` has returned in every revision since it was introduced. When no per-word confidence exists for the matched span there is nothing to take the minimum of, which is what the sentinel means.
- The **Repro** derived the sentinel from a leading `[unk]` run ("okay cease fire" → `[unk] cease fire`), which per F5 cannot produce it — selection never starts a match on `[unk]`, so a winning span always holds at least one matched word. Replaced with a repro that does reproduce (a result carrying no per-word data), and the entry now states explicitly that an `[unk]` run is not a cause, with a cross-link to the second, less obvious route.

The **Where seen** history and the workaround are untouched.

**F9** — `scoring.md` added to the README documentation list, in the same position it occupies in `index.md`. This settles that list as a mirror rather than a curation; `native-bridge.md`'s long-standing absence from it is still unresolved either way, and I have left it alone.

Links and anchors re-checked across `Documentation~/`, `KNOWN_LIMITATIONS.md`, `README.md` and `CHANGELOG.md`: 0 broken.

**F10 remains open** — nothing pins the documented arithmetic, and adding the test needs the host-project Unity run this PR defers.

PR #64, opened 2026-08-12
Enforce single-owner access to the process-global native bridge (#57)

### LINKED ISSUE #57: Two VoxrSpeechRecogniser components silently share one process-global bridge; either OnDestroy frees both
Two `VoxrSpeechRecogniser` components in one scene silently share a single process-global native bridge. Neither component is aware of the other, and whichever one is destroyed first tears down the recognizer and model out from under the survivor.

This is the product bug behind the test-side fix in #55 / PR #56 — that PR excluded the PlayMode test assembly from Android players precisely so the on-device player could not hit this. #55 and #56 both promised this follow-up issue; filing it now. It was independently confirmed by the review pass posted on PR #56.

## Mechanism

The bridge is process-global C++ state, not per-instance:

- `NativeBridge~/src/vosk_bridge.cpp:246-272` — `vosk_bridge_destroy()` operates on file-scope globals (`g_recognizer`, `g_model`, `g_initialised`, `g_recognition_thread`). There is no handle or instance id in the ABI; there is exactly one bridge per process.

On the managed side each component believes it owns that state:

- `Runtime/VoxrSpeechRecogniser.cs:68-88` — `IsInitialised` is not backed by an instance field. It calls `BridgeNative.vosk_bridge_is_initialised()`, i.e. it reports whether *the process* has an initialised bridge.
- `Runtime/VoxrSpeechRecogniser.cs:130-133` — `InitialiseAsync()` early-returns on `if (IsInitialised) return;`. So the second component's `InitialiseAsync` returns immediately as a no-op and never applies its own model path, sample rate, or gain settings — it silently inherits the first component's configuration.
- `Runtime/VoxrSpeechRecogniser.cs:188-207` — `ReleaseNativeResources()` calls the unconditional `vosk_bridge_destroy()`.
- `Runtime/VoxrSpeechRecogniser.cs:515-518` — `OnDestroy()` calls `ReleaseNativeResources()`.

## Repro

1. Put two `VoxrSpeechRecogniser` components in a scene (two GameObjects, or two additively-loaded scenes each carrying one).
2. `await` `InitialiseAsync()` on component A, then on component B.
3. Observe B's configuration was never applied — B reports `IsInitialised == true` having initialised nothing.
4. Destroy either component (or unload either scene).
5. The survivor's `IsInitialised` now reports `false` and every subsequent bridge call operates on a freed recognizer/model.

## Impact

- Silent misconfiguration: a second recogniser's model path / sample rate / AGC target are discarded with no error or warning.
- Cross-instance teardown: one component's `OnDestroy` invalidates another's native state. In a scene-additive VR app this happens on ordinary scene unload, not just on quit.
- It blocks running more than one recogniser fixture per player process — which is why the Android exclusion in #56 was needed on the test side.

## Possible directions (not decided)

- Refcount initialisation in the bridge, so `destroy` only tears down on the last release.
- Give the ABI a handle so instances are genuinely independent.
- Or make the constraint explicit and enforced: document single-instance-per-process, and have a second component fail loudly (error + disable) instead of silently sharing.

Related: #55, #56.

jinwoo1601:
Closes #57

## What

Gives the process-global native bridge a **single explicit owner**, so a second `VoxrSpeechRecogniser` can no longer inherit the first's configuration or free its recognizer.

A static `s_bridgeOwner` records which component initialised the bridge. The claim is taken **synchronously before the first `await`** in `InitialiseAsync` — Unity runs this on the main thread, so a check-and-set with no await between its halves cannot interleave with a second component's call; claiming after the model-extraction await would let both instances past. It is released on `ReleaseNativeResources()` / `OnDestroy`, and released again in a `finally` when the attempt fails, so one component's bad model path cannot lock out a sibling. A destroyed owner's claim lapses via Unity's `== null` overload even if its `OnDestroy` never ran (a subclass can hide the base's private `OnDestroy` — `VoxrPushToTalkPauseTests` does exactly that).

Every other instance becomes inert towards the bridge:

| Member | Behaviour when another component owns the bridge |
|---|---|
| `IsInitialised`, `IsRecognising` | `false` — they described the *process*, not the instance |
| `InitialiseAsync()`, `SetGrammar()`, `ResetRecogniser()` | Rejected: `Debug.LogError` **and** `OnError`(`AlreadyInitialised`), naming the owning GameObject |
| `StopRecognition()`, `ReleaseNativeResources()` | Quiet no-ops, so a duplicate's `OnDestroy` cannot free the owner's state |

`StartRecognitionAsync` needed no guard of its own — it already gates on `IsInitialised`, which now reports honestly.

Loud on both channels deliberately: the silent sharing this replaces was invisible to a developer who had not subscribed to `OnError`, which is exactly who it stranded. Quiet on stop/release because those are teardown, are called on every push-to-talk release, and the loud report already happened at initialise.

## Why

`vosk_bridge.cpp` keeps `g_model` / `g_recognizer` / `g_initialised` at file scope and its C ABI carries no handle, so there is one bridge per process — but the managed side never modelled that. `IsInitialised` was not backed by an instance field; it called `vosk_bridge_is_initialised()`. So `InitialiseAsync` early-returned on *another* component's success and discarded its own model path, sample rate and AGC target with no error, and `ReleaseNativeResources()` called the unconditional `vosk_bridge_destroy()` — reached from `OnDestroy()` — so either component's destruction freed the other's recognizer and model. In an additive-scene VR app that fires on ordinary scene unload, not just on quit. It is also what forced the PlayMode test assembly to be excluded from Android players in #55.

### Scope

Issue #57 listed three undecided directions. The other two — refcounting the bridge, or giving the ABI a per-instance handle so two recognisers genuinely coexist — are native-side redesigns requiring on-device verification, so they stay open; this PR takes the third and makes the constraint explicit and enforced, managed-side only. **The native ABI is unchanged.**

The guards key off *"another live component owns the bridge"* rather than *"am I the owner"*, so **a lone recogniser sees no behaviour change whatsoever** — the new code path only engages once a second live instance exists. That is what keeps all 453 existing tests semantically untouched, including the case of an uninitialised single recogniser calling `ResetRecogniser()`.

Documented in `api/speech-recogniser.md` (new *One recogniser per process* section), the getting-started lifecycle section, and a `KNOWN_LIMITATIONS.md` entry carrying the persistent-recogniser pattern to use instead of one per scene.

## Validation

- [x] **Unity EditMode — 116/116 passed, 0 failed** (host project `VoXR TestGround`, Unity 6000.4.7f1, against this working tree)
- [x] **Unity PlayMode — 337/337 passed, 0 failed**
- [x] 7 new PlayMode tests in `VoxrSpeechRecogniserOwnershipTests` confirmed present in the result XML and passing: second-initialise rejected while owner keeps the bridge; second destroyed leaves owner initialised (the headline regression — and the second component is deliberately left *uninitialised*, since that is the worst form of the bug: it needed no init, only an `OnDestroy`); second `ReleaseNativeResources` leaves owner initialised; `SetGrammar` and `ResetRecogniser` rejected loudly; owner destroyed **and** owner released each free the claim for the next recogniser.
- [x] The two suites most likely to be perturbed — both initialise a real recogniser and would break first if a claim leaked across SetUp/TearDown — are unaffected: `VoxrPlaybackSeamTests` 8/8, `VoxrWavReplayTests` 17/17.
- [x] No `NativeBridge~/` change, so the arm64 build and desktop harness gates do not apply and were not run.
- [ ] **On-device (Quest) check needed.** This touches `VoxrSpeechRecogniser` lifecycle behaviour, and the new tests are wrapped in `#if UNITY_EDITOR_WIN` — establishing a real owner needs a backend that actually initialises, which off-device means `EditorMicBackend`. So the Android branch of these guards (`vosk_bridge_destroy` / `vosk_bridge_stop` / `vosk_bridge_set_grammar` suppression for a non-owner) has no automated coverage. Worth confirming a normal single-recogniser scene still initialises, recognises and tears down unchanged on device.


jinwoo1601 (commented):
## Independent review — PR #64

**Depth profile:** slim (issue lane, head branch `issue/57-*`) — angles `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`. No angles dropped.
**Head reviewed:** `23b5240bcb8f6d42c6803d5f76f34f4e78bf1427` (base merge-base `052b33f`).
**Method:** six finder agents ran against the diff with no context from the session that wrote it; the `CLAIM` angle audited this PR's own body and every ticked validation box. All ten deduplicated findings then went to independent verifiers prompted to *refute*. **8 confirmed, 2 refuted.**

Since the same session wrote the code, the two findings that land hardest on that session's own claims (F1, F7) are stated below at full strength rather than softened.

---

### Fixed on this branch — pushed as `5e3e701`

#### F1 — `InitialiseAsync` never re-checked ownership after its awaits · CONFIRMED (LINE, PIT, GAP, GONE — four angles independently)

The claim was taken synchronously before the first `await`, but an `async` continuation outlives the component that started it. A destroy or a `ReleaseNativeResources()` landing during model extraction hands the claim away mid-flight; the success path never re-checked, so it finished initialising anyway and left the bridge initialised with `s_bridgeOwner == null` — **the one state in which every guard in this PR is inert**, since they all keyed off *"another live component owns it"*. From there the next component hit the unchanged `if (IsInitialised) return;`, inherited the owner's configuration, and freed it on destroy: issue #57 verbatim.

Verification corrected the severity in **both** directions, and the correction matters:

- The unowned-bridge end state is **equally reachable at the merge base**, where `InitialiseAsync` had no claim at all. So that half is the new invariant *under-delivering in one window*, **not a regression**. Against base parity it is not blocker-level.
- The strict regression this PR *did* introduce is the narrower second half: a non-owner's `ReleaseNativeResources` early-returned past `_editorBackend?.Release()` — the only caller of that method anywhere in the repo, and what frees `vosk_model_free` / `vosk_recognizer_free` / `Microphone.End`. Base ran it unconditionally. Reachable when a component's init completes after its own release, then a sibling claims.
- The in-repo `EditorMicBackend._released` flag exists precisely because continuations survive teardown here — but it does not cover this, because in the failing trace the backend is constructed *after* the release.

**Fix.** Re-check `OwnsBridge` after each await and abandon the initialise if the claim is gone, releasing an editor backend that came up orphaned. Split `ReleaseNativeResources` by what the resource actually *is*: the editor backend is instance-owned and is always released; only the handle-less `vosk_bridge_destroy()` is suppressed for a non-owner. And back `IsInitialised` / `IsRecognising` on ownership rather than on "someone else owns it" — which is what issue #57 asked for in the first place ("`IsInitialised` is not backed by an instance field"). An unowned-but-initialised bridge now produces a loud native `ALREADY_INITIALISED` on the next initialise instead of a silent hand-over.

#### F3 — docs and the rejection message treated `Destroy` and `ReleaseNativeResources()` as interchangeable · CONFIRMED (PIT)

Only `ReleaseNativeResources()` frees the claim synchronously. Unity defers `Object.Destroy` — and therefore `OnDestroy` — to end of frame, and the fake-null lapse flips at the same moment, so **both** release paths are unavailable until then. The handover this PR's own docs recommended (`Destroy(outgoing); incoming.Initialise();` in one frame, or an `UnloadSceneAsync` overlapping the incoming scene's `Start()`) was therefore rejected with an error. The shipped error string said it too: *"Release or destroy the existing recogniser first."*

The finding's escalation was **refuted**: `StartRecognitionAsync` retries, so under the default push-to-talk wiring the next press recovers, costing only the pre-warm. It is sticky only in `Continuous` listening mode, which has no retry. Docs and the error message now say which verb is synchronous and what retries.

#### F8 — the failed-initialise claim release had no test · CONFIRMED (GAP)

`if (!claimHeld && OwnsBridge) s_bridgeOwner = null;` is the only *automatic* path that frees a claim stranded by a failed initialise — without it one bad model path bricks the bridge process-wide. No test in the repo exercised it. Added `FailedInitialise_ReleasesClaimForNextRecogniser`, which sets `modelRelativePath` by reflection to a bogus leaf name (the extractor caches on `Path.GetFileName`, so a bogus path sharing the real leaf would succeed off the existing extraction) and asserts a sibling can still take the bridge.

**Re-verified after the fix — Unity 6000.4.7f1, host project `VoXR TestGround`: EditMode 116/116, PlayMode 338/338, zero failures.** The +1 is exactly the new test; every other fixture held its prior count.

---

### Left for your ruling — not fixed

#### F2 — the Editor backend was genuinely per-instance at base; the rationale text is untrue on that branch · CONFIRMED-as-clarity (GONE) · **the one that wants a decision**

The guards sit *above* the `#if UNITY_EDITOR_WIN` split, so they govern the editor branch too — but `EditorMicBackend` makes zero `vosk_bridge_*` calls and holds instance `IntPtr _model` / `_recognizer`. At base, two recognisers in the Windows Editor each loaded their own model with their own settings, and destroying one freed only its own. **None of #57's three symptoms existed on that branch.**

Verification narrowed this a long way and it is *not* a code-scoping defect: the lost capability is unreachable by any package consumer (`EditorBackend`/`StartPlayback` are `internal`), no in-repo caller does it, base could not have run two live captures anyway (`Microphone.Start(null, …)` keys on the shared default device), and the uniform policy is deliberate and declared in the new test file's own header comment.

What survives is that a Windows-Editor developer who trips the guard is handed a reason that is **demonstrably untrue of their platform** — in the error string, in `api/speech-recogniser.md` ("a process has exactly one recogniser and one model no matter how many components reference it"), and in `KNOWN_LIMITATIONS.md` ("Nothing on the managed side can make two components genuinely independent without changing that ABI" — on that branch, it already had).

I did not rewrite this myself: whether to qualify the text or to scope enforcement to the native branch is the framing of the whole change, and that is your call, not mine.

#### F7 — four of seven original tests are green at base, including both "headline regression" tests · CONFIRMED and *strengthened* (CLAIM, GAP)

Verification found it worse than reported — four, not two. `SecondDestroyed_LeavesOwnerInitialised` and `SecondReleaseNativeResources_LeavesOwnerInitialised` drive a `_second` deliberately left uninitialised, so its `_editorBackend` is null, and base's editor-branch release was already a null no-op. The destructive `vosk_bridge_destroy()` they are named for is in the `#else` arm, which cannot compile in the only configuration that runs them.

**Decisively: delete the entire non-owner teardown guard and both tests stay green.** No executable check in this PR fails if that guard is removed, and `StopRecognitionCore`'s guard has no test at all.

The original validation box called these "the headline regression … confirmed … passing". **It should not have.** They are smoke tests; invariant 2 rests entirely on the deferred on-device check. The other three (`SecondInitialise`, `SecondSetGrammar`, `SecondResetRecogniser`) genuinely do fail at base, and the two claim-release tests are head-mutation-sensitive.

This is also why the F1 fix above ships with only *one* new test rather than one per change: `EditorMicBackend._released` already covers the editor-branch race, and `ExtractModelAsync` returns synchronously on a warm cache, so the F1 paths are not reachable from a Windows-editor test. Adding one that looked like it pinned them would have repeated exactly this mistake.

#### F4 — the mic-permission coroutine re-enters `StartRecognitionInternal` with no re-check · CONFIRMED, minor (CLAIM, GAP)

`RequestMicPermissionThenStart` yields unboundedly on the first-run Android dialog, then calls `StartRecognitionInternal()` with no ownership or `IsInitialised` re-check, arming `_isRecognising` and its ungated `Update()` drain against a bridge another component may now own — dispatching the owner's results to the wrong subscribers. **Not a regression**: that block is byte-identical to base, so it is an *uncovered* path for the new invariant. `#if UNITY_ANDROID && !UNITY_EDITOR`, so no test here can reach it. The "taken over" leg was refuted (a destroyed component's coroutines stop); the surviving leg is an explicit `ReleaseNativeResources()` overlapping the prompt — which is the handover the docs now recommend.

Correspondingly, the body's *"`StartRecognitionAsync` needed no guard of its own"* is true of the synchronous path and **over-broad**, not false; *"every other instance becomes inert towards the bridge"* is the line this path does not honour.

#### F6 — docs omit that `Initialise()` / `StartRecognition()` inherit the loud rejection · CONFIRMED (a) (XFILE)

Those are the members `VoxrPushToTalkController` actually calls, and on a non-owner each press fires one `Debug.LogError` **and** one public `OnError(AlreadyInitialised)`. No doc says so — the new section, the Methods table (`StartRecognition()` still reads "Calls `Initialise()` if needed"; `Initialise()` still reads "No-op if already initialised", which is false for a non-owner), `getting-started.md`, and `error-codes.md` all attribute that error to `InitialiseAsync`/`SetGrammar`/`ResetRecogniser` only.

The per-press volume itself was **refuted** as a defect — the loud init path is load-bearing for invariant 3, and it is *why* the stop guard can be quiet. But nothing records that asymmetry, so the intent is not legible from the artifact.

#### F9 — "all 453 existing tests" · CONFIRMED, nit (CLAIM)

446 pre-date this change; 453 was the post-change total, double-counting the 7 tests added here — which are precisely the ones that instantiate a *second* recogniser, i.e. outside the claim's own reasoning. Independently confirmed by counting the base tree (116 EditMode; 291 `[Test]` + 14 `[UnityTest]` + 10 `[TestCase]` + 15 `[ValueSource]` expansions = 330 PlayMode). **Now 116 + 338 = 454, of which 446 pre-date this PR.**

---

- **F5 — `AlreadyInitialised` reused for a managed-only condition contradicts `error-codes.md`.** REFUTED. The "native bridge only" framing was already false at base in a sharper form: `MarkBridgeUnavailable` fires `ModelLoadFailed` from managed code *when no native bridge exists*, and `EditorMicBackend` / `ModelExtractor` / the permission callback fire four more. Five of eight codes already carried managed-side uses. The description and message read as redundant, not contradictory, and a dedicated code would violate the PR's own no-new-enum-values scope. Residue: `error-codes.md`'s scope line has been over-narrow since well before this PR.
- **F10 — `Closes #57` retires the two directions the PR says stay open.** REFUTED on repo precedent: PR #61 (`Closes #45`) and PR #60 (`Closes #44`) both closed with declared, unfiled residue. The issue framed its three directions as alternatives ("**Or** make the constraint explicit and enforced") and the reported defect no longer reproduces. The #56 sub-claim was materially wrong — that impact bullet ends "on the test side", and the Android exclusion's motivating consumer (Tier D) was abandoned 2026-08-11.
  - *Unrelated hygiene note it surfaced:* `KNOWN_LIMITATIONS.md:306` claims a residue was "Filed as a low-priority follow-up" and no such open issue exists.

---

### Not verifiable from this environment — stated, not claimed

- **On-device Quest / Android arm64.** Every `#else` (native) branch of every guarded method is compile-only from WSL. Given F7, this is where invariant 2 actually rests — the cross-instance teardown guard has **no** automated coverage in any configuration, and F4 is Android-only by construction.
- **The published-package view.** No package build is runnable here.
- The host-project Unity runs *were* executed for this branch (both before and after the fix) and are reported above with counts read from freshly-written NUnit XML, start-times checked against the run to rule out the stale XMLs in that tree.

<sub>Review by Claude Code (`review-pr`, slim profile). Findings are evidence for your ruling, not a ruling — nothing here is approved, merged, or closed by me.</sub>

jinwoo1601:
Pushed `5e3e701`, which addresses **F1** (ownership not re-checked across `InitialiseAsync`'s awaits — including the strict regression where a non-owner's `ReleaseNativeResources` skipped `_editorBackend.Release()`), **F3** (`Destroy` vs `ReleaseNativeResources()` synchronicity, in the docs and the rejection message), and **F8** (added `FailedInitialise_ReleasesClaimForNextRecogniser`). Re-verified: EditMode 116/116, PlayMode 338/338.

Two corrections to this PR's own body, both raised by the review against the session that wrote it:

- **The validation box overclaimed.** `SecondDestroyed_LeavesOwnerInitialised` and `SecondReleaseNativeResources_LeavesOwnerInitialised` were described as confirming "the headline regression". They do not — they pass against `main` too, and deleting the non-owner teardown guard outright leaves them green. They are smoke tests; the cross-instance teardown guard has no automated coverage in any configuration, and invariant 2 rests entirely on the deferred on-device check (F7).
- **"all 453 existing tests" should read 446** — 453 was the post-change total, double-counting the tests this PR adds. Current totals are 116 + 338 = 454, of which 446 pre-date this branch (F9).

F2, F4, F6 and F7 remain open on the PR, unfixed and unruled — F2 in particular wants your decision, since it is about how the change is framed rather than a bug.

jinwoo1601:
Pushed `d38e30c`, resolving **F2** per your ruling: keep uniform enforcement, correct the rationale.

The enforcement is unchanged — what changed is the reason it gives. The rejection message, the API doc's *One recogniser per process* section, `KNOWN_LIMITATIONS.md`'s root cause, and the CHANGELOG entry all previously told a Windows-Editor developer that the managed side cannot make two components independent without an ABI change. That is false on the branch they are running: `EditorMicBackend` holds instance `IntPtr _model`/`_recognizer`, loads its own model, and never calls `vosk_bridge_*`. Each now scopes the ABI argument to device and says the Editor is held to the same rule for parity — so a scene that works in the Editor cannot fail on device — noting that enforcing on the Editor branch is also what makes the constraint testable at all, since the coverage runs in EditMode/PlayMode rather than on device.

Two smaller things folded in:

- **F6's documentation half.** The non-owner list presented itself as exhaustive but omitted `Initialise()` and `StartRecognition()`, which route through `InitialiseAsync()` and inherit the rejection once per press. F6's other half stays as-is — verification found the loud init path is load-bearing for the no-silent-misconfiguration invariant, and is precisely why the stop guard can afford to be quiet.
- `KNOWN_LIMITATIONS.md` now states plainly that the two native-side directions are unfiled and that the entry is their only record.

Re-verified: **EditMode 116/116, PlayMode 338/338**, no count drift. The rewording moved a string that three `LogAssert.Expect` tests match on, so the regex moved with it; all three pass.

Still open and unruled: **F4** (Android mic-permission coroutine re-enters `StartRecognitionInternal` without re-checking ownership — minor, not a regression, unreachable by any test here) and **F7** (the cross-instance teardown guard has no automated coverage in any configuration; invariant 2 rests on the deferred on-device check).

Issue #65, opened 2026-08-12
Scoring model: pattern length determines drop fragility, and coverage never enters sibling comparison


jinwoo1601:
## Summary

Carried forward from #42, which was closed by #58 having delivered its validation warning and authoring guidance. Two things #58 explicitly left open remain, and they are the same underlying question: **the scoring model normalises every candidate to a 1.0 ceiling, so a pattern's length determines how much a single dropped word costs it, and how much of the utterance a candidate actually explains never enters the comparison.**

## Symptom 1 — length determines fragility

Observed in-headset during the #42 test cycle: a dropped function word pushed a 3-word pattern to **0.50**, under the 0.60 default `minScore`, so the command did not fire at all. A 7-word pattern absorbed the *identical* single-word drop at **0.79** and fired cleanly.

Short, specific patterns are structurally the most fragile to exactly the drop VOSK makes most often — short unstressed function words ("by", "to", "the"). Compare "time to target" → "time target" from the same cycle. There is no bare sibling involved here, so neither #58's detector nor its authoring rule reaches this case: an author following the guidance perfectly still gets it.

## Symptom 2 — coverage is not part of sibling comparison

The original #42 case: `decelerate` scores 1.0 (1/1) while `decelerate by {burn_level}`, charged `RequiredLiteralMissPenalty` for the dropped "by", scores 0.5 (1.5/3). The bare pattern wins and the heard "hard burn" is silently discarded. #58's authoring remedy (`?by`) fixes the shape where an author can restructure the grammar; the comparison itself still ignores that one candidate explains recognized, in-grammar slot content the other leaves unconsumed.

## Why this needs a design pass, not an issue-lane fix

From #58's analysis, recorded so it is not re-derived:

- Because every score normalises to a 1.0 ceiling, **no penalty or threshold tuning reorders the #42 pair.** This is not a tie-break problem (#41's consumed-span tie-break does not reach it either).
- A new ordering key **above** `Score` in `IsBetterCandidate` flips selection but leaves the winner at 0.5, under the `minScore` gate — the command stops firing instead of firing with the slot filled. Strictly worse than today for that utterance.
[...]

Issue #33, opened 2026-07-26
TokensCompatible treats slot-vs-literal as always compatible — lone-slot patterns can never eager-flush


jinwoo1601:
Companion to #32, but where #32's prefix hold is *genuine*, this one is a false positive.

Eager commit requires the matched pattern to be no compatible prefix of any longer pattern. `TokensCompatible` is deliberately conservative: any position involving a **slot** is treated as compatible with anything. A bare single-slot pattern (`{burn_level}` in The Long Burn's grammar) is therefore judged a potential prefix of **every** ≥2-token pattern in the grammar (slot-vs-"close", slot-vs-"intercept", …), so `CanCommitEarly` is structurally false for it — every bare "coast" / "hard burn" waits the full buffer window, despite zero real ambiguity (no surface form of the slot begins any other pattern; "coast" is no one's first word).

Request: **value-aware compatibility** — treat slot-vs-literal as compatible only when some surface form in the slot's vocabulary has its *first word* equal to that literal. This makes lone-slot patterns eager-committable with no grammar change, keeps genuine holds (e.g. `decelerate` vs `decelerate {burn_level}`) intact, and fixes the class rather than the instance.

PR #67, opened 2026-08-12
Refuse an eager commit when a required slot is unfilled (#66)

### LINKED ISSUE #66: Eager flush can commit a command with an unfilled required slot
## Summary

`TryEagerCommit` can return a verdict above `None` for a winning pattern whose **required slot was never filled** — firing a command with a missing argument, and truncating the utterance immediately before the word that would have supplied it.

The eager gate never checks that required elements actually matched. It relies on the score arithmetic to hold such candidates below `minScore`, which is emergent rather than enforced — and at five elements the arithmetic lands on exactly `0.60`, which passes the default gate.

## Repro (against the shipped demo grammar)

Pattern: `["launch", "{?quantity}", "{weapon}", "target", "{target}"]` (`Samples~/CommandRecognition/CommandDemo.cs`, mirrored in `Tests~/Runtime/DemoGrammar.cs`).

Speak **"launch all missiles target hotel one"** with `eagerFlushOnCompleteMatch` enabled. Mid-utterance the buffer is `launch all missiles target`, and that buffer scores:

| Element | Outcome | Raw | Denom |
|---|---|---|---|
| `launch` | matched | +1.0 | +1.0 |
| `{?quantity}` → "all" | matched optional slot | +1.0 | +1.0 |
| `{weapon}` → "missiles" | matched | +1.0 | +1.0 |
| `target` | matched | +1.0 | +1.0 |
| `{target}` | **missed required slot** | **−1.0** | +1.0 |

`score = 3.0 / 5 = 0.60`.

Now walk the gate in `Runtime/Commands/VoxrCommandParser.cs`:

1. `:1337` — `bestScore < minScore` is `0.60 < 0.60` → **false**, so it passes.
2. `:1356` — `bestStartIdx != firstRecognisedIdx || bestEndIdx != tokens.Length`. The match starts at token 0, and **a missed slot consumes nothing, so it never advances `tokenIdx`** — `EndIdx` stays at 4, which equals `tokens.Length`. The match therefore "reaches the end of the buffer" **without having filled the slot**, and this check passes too.
3. Confidence passes on the four matched words.

Result: a `Commit`/`HoldExtendable` verdict that flushes `launch_weapon` with `target` unfilled — right before "hotel one" arrives to fill it.

## Root cause

The gate's four conditions (documented at `Documentation~/scoring.md:209-216`) assert *score*, *start position*, *end position*, and *confidence*. None asserts **completeness**. The end-position condition looks like it would catch an unfilled trailing slot, but it compares `EndIdx`, which a miss does not advance — so an unmatched required slot is invisible to it.

`RequiredSlotMissPenalty = -1.0` (`:34`) happens to sink most such candidates below `minScore`, which is why this has not been hit often. Five elements with exactly one missed slot is the case where it does not.

## Proposed fix

Add an explicit completeness condition: **no verdict above `None` when the winning candidate has a missed required slot.**

Missed required *literals* must remain acceptable — the command is still fully determined and all its arguments are present, so a dropped function word should not block an eager commit.

`MatchResult` already carries `SlotCount` (`:999`); the check needs the pattern's *required* slot count to compare against, or an explicit "missed a required element" flag set in `TryMatchScored` at `:954`.

## Tests

- Buffer `launch all missiles target` → verdict `None` (the case above).
- Buffer `launch all missiles target hotel one` → still `Commit`. Guards against over-correcting.
- A pattern with a missed required *literal* but all slots filled → verdict unchanged, proving literals are not caught by the new condition.

Existing eager tests live in `Tests~/Runtime/VoxrEagerCommitTests.cs`.

## Provenance and sequencing

Carved out of the #65 scoring-model design pass as an independently valuable fix — it is a live bug on `main` today, not a consequence of that redesign.

**It should land before the #65 work begins.** That design lowers the required-literal miss cost, which widens this hole slightly: a pattern of **8+ elements** with a missed slot *and* a dropped literal would newly cross `0.60` (`(6−1)/8 = 0.625`, currently `(6−1−0.5)/8 = 0.5625`). Fixing the missing condition first means that change cannot introduce a new instance of this bug.

## Note on verification

The arithmetic and the two gate conditions above are derived by reading `TryEagerCommit` and `TryMatchScored` directly; the repro has **not** been executed in the editor. The first task on picking this up is to turn the case above into a failing test and confirm it fails for the stated reason.

jinwoo1601:
Closes #66

## What

`TryEagerCommit` gains a fifth gate condition: **no verdict above `None` when the winning candidate has a missed required slot.**

- `MatchResult` carries a new `MissedRequiredSlot` flag, set in `TryMatchScored` on the required-slot miss branch.
- `TryEagerCommit` tracks it for the winner and returns `None` when set.
- Required *literals* are deliberately exempt — a dropped function word leaves every slot filled and the command fully determined, so it must not block an eager commit.
- `ParseInternal` ignores the flag entirely: a partial match is still a legitimate parse result there and still reaches the pending-command path after the full `bufferWindow`.

Docs: `scoring.md` §6 goes from four gate conditions to five; `command-recognition.md` gains the matching "keeps waiting" bullet; `CHANGELOG.md` under `[Unreleased] → Fixed`.

## Why

The gate asserted score, start position, end position, and confidence — nothing asserted **completeness**. It relied on `RequiredSlotMissPenalty = -1.0` to sink incomplete candidates below `minScore`, which is emergent rather than enforced. At five elements with exactly one missed slot the arithmetic lands on `3.0 / 5` = **exactly `0.60`**, which passes the default gate.

The end-of-buffer condition looks like it would catch an unfilled trailing slot, but cannot: a miss consumes no tokens, so it never advances `EndIdx`. The match therefore "reaches the end of the buffer" without having filled the slot.

Note that `api/command-recogniser.md` already documented `eagerFlushOnCompleteMatch` as firing when the buffer "can't be extended **or completed** by more words" — the contract was written correctly; the parser never enforced the "completed" half.

## Validation

**The repro was executed, not just derived.** The issue flagged that its arithmetic came from reading the source and had not been run. I compiled the real parser sources with `dotnet` against a `UnityEngine.Debug` stub (the established WSL rig pattern) and drove `TryEagerCommit` directly against the shipped demo grammar.

Before the fix:

| Buffer | Verdict | Parse |
|---|---|---|
| `launch all missiles target` | **`Commit`** | `launch_weapon` score `0.60`, slots `[quantity=all, weapon=missiles]` — **no `target`** |
| `fire all missiles at` | **`Commit`** | `launch_weapon` score `0.60`, slots `[quantity=all, weapon=missiles]` — **no `target`** |

The issue named the `launch` pattern; `fire {?quantity} {weapon} at {target}` is a **second live instance** in the same shipped grammar, found by the rig.

After the fix both return `None`, while `launch all missiles target hotel one` and `fire all missiles at hotel one` still `Commit`, and `launch all missiles hotel one` (dropped `target` literal, all slots filled, score `0.70`) still `Commit`. Nine regression probes across the demo grammar are unchanged.

- [x] **Unity PlayMode — 342/342 passed, `failed="0"`.** Runtime tests are PlayMode, so this is the run that covers the four new tests; all four passed. `VoxrEagerCommitTests` reports 44/44 — the 40 pre-existing eager tests still pass alongside the additions.
- [x] **Unity EditMode — 116/116 passed, `failed="0"`.**
- [x] Four tests added to `Tests~/Runtime/VoxrEagerCommitTests.cs`: the repro → `None`; the filled case → `Commit` (guards against over-correcting); an unfilled *optional* slot → `Commit`; a missed required *literal* with all slots filled → `Commit` (proves literals are not caught).
- [x] `NativeBridge~/` untouched — bridge build and desktop harness correctly not run.
- [ ] **On-device (Quest) verification: deferred, not performed.** This is pure parser logic with no bridge, mic-capture, or MonoBehaviour-lifecycle surface, so it does not touch what that binding exists to cover.

## Sequencing

Per the issue, this should land **before** the #65 scoring-model work. That design lowers the required-literal miss cost, which widens this hole: an 8+ element pattern with a missed slot *and* a dropped literal would newly cross `0.60` (`(6−1)/8 = 0.625`, currently `(6−1−0.5)/8 = 0.5625`). Fixing the missing condition first means that change cannot introduce a new instance of this bug.


jinwoo1601 (commented):
## Independent review — PR #67

**Head reviewed:** `d57c76431237d17a8a05d5631b88c127174029b1` · merge-base `d081e11`
**Method:** six finder agents ran against the diff with no context from the session that wrote it, including a `CLAIM` angle auditing this PR's own body and every ticked validation box. Seven surviving findings each went to a separate adversarial verifier instructed to default to REFUTED.

**Result: 3 confirmed (2 nits, 1 minor), 4 refuted. No blockers.** Nothing was fixed — per the issue lane, only a confirmed blocker is fixed in-run; everything below stays here for you to rule on.

---

### Confirmed findings

#### 1. `LINE-1` — the new comment's stated *reason* is over-general — **nit**, high confidence

`Runtime/Commands/VoxrCommandParser.cs:1355` justifies the new condition with "a miss consumes no tokens and so never advances `EndIdx`". That is not true in the `[unk]` case. The `[unk]` skip at `:928` runs before **every** element, including one that then matches nothing, and `EndIdx = tokenIdx` is returned unconditionally at `:1009`. For `[launch, all, missiles, target, [unk]]`, element 4's skip advances `tokenIdx` 4 → 5, `TryMatchSlot` then returns null, and the required-slot miss has moved `EndIdx`.

The same file already says so 500 lines earlier, in the field's own doc comment at `:845` — "including any `[unk]` skipped ahead of a trailing element that matched nothing" — and that comment is pre-existing, so the new text was written against a field whose documentation already recorded the exception.

**No behavioural consequence.** The `[unk]` advance makes `EndIdx == tokens.Length` *more* likely, so the whole-buffer gate passes a fortiori and the new completeness condition is still the only thing that catches the case. The conclusion holds; only the mechanism is overstated. Correct phrasing would scope it to *recognised* tokens.

Footprint is three artifacts, all repeating the same claim: `VoxrCommandParser.cs:1355`, `Documentation~/scoring.md:214`, `CHANGELOG.md:30`. The `command-recognition.md` bullet is unaffected — "an unspoken slot consumes no *words*" is accurate.

#### 2. `GAP-2` — the `HoldExtendable → None` half of the change is untested — **minor**, high confidence

No test anywhere reaches `:1364` with `bestMissedRequiredSlot == true` on a **non-committable** winner. Every existing missed-required-slot case either dies earlier at the score gate (`VoxrEagerCommitTests.cs:410-422`, score `0.5 < 0.6`) or wins with a terminal pattern (the four new tests). Both un-analysable-grammar degrade tests are structurally blind to it: `OverLimitCommand()` at `:315` is 13 optional *literals* with **zero slots**, and both callers build it with an empty `Slots()`.

The verifier established this is **not** merely a refactor guard — the missing test fails on base and passes on head, making it a genuine regression test for behaviour this PR changes. Under the #44 degrade with `prefixHoldSeconds = 0.6f`, base shortens the window from 2.0s to 0.6s for a `launch_weapon` with no `target`: the exact harm #66 exists to stop, via the one path no test covers. `scoring.md:209` also contracts it explicitly — condition 4 gates *any* verdict above `None`, and `HoldExtendable` is above `None`.

A single case would close it: an over-limit command plus the launch pattern, buffer `launch all missiles target`, asserting `None`.

#### 3. `XFILE-1` — new docs bullet omits *which* wait — **nit**, medium confidence

`Documentation~/command-recognition.md:300` ends "keeps waiting", where its immediate neighbour at `:299` says "keeps waiting **the full window**". The precision is inverted: `:299`'s case yields `HoldExtendable` and may wait only `prefixHoldSeconds`, while `:300`'s case yields `None` and always waits the full `bufferWindow`. One phrase closes it.

Two escalations were checked and **refuted**: the "second bullet above" cross-reference at `:310` survived the insert (the new bullet landed at position 3), and `:320` in the same file already resolves the ambiguity rather than compounding it.

---

### Refuted findings

Named so you can see what was considered and killed.

- **`GAP-1`** — "nothing asserts `ParseInternal` still handles the repro buffer". **Refuted:** `VoxrPendingCommandTests.cs:99` already injects `launch missiles target` — a missed-required-slot buffer — and asserts the pending path, with nine further call sites; `VoxrCommandParserTests.cs:1227` pins selection where a missed-slot candidate loses. The claimed in-file convention also doesn't exist (3 of 44 tests call `Parse()`). Structurally, the flag is write-only outside `TryEagerCommit`, so `ParseInternal` cannot read it.
- **`GAP-3`** — "the end-to-end split-utterance payoff is untested". **Refuted:** the recogniser branches on the verdict only (`VoxrCommandRecogniser.cs:497-500`) and cannot see *why* `None` was returned, so a new test would exercise no new path. The existing split test's second injection merges to a 1.0 buffer that already executes the new line with the flag clear.
- **`GAP-4`** — "the gate silently changes `allowPartialMatch` behaviour". **Refuted, premise inverted:** the pending branch at `VoxrCommandRecogniser.cs:668` requires `score < minScore`, the exact complement of the eager gate's `>= minScore`. Pre-change, such a buffer did not route into the follow-up flow — it fired the command **with the argument missing**. The change is a strict improvement there, and the finding's proposed test would have encoded a false expectation. No shipped sample sets `allowPartialMatch`.
- **`COV-1`** — "the `fire …` second instance has no test". **Refuted:** `DemoGrammar` configures a *recogniser*, while `TryEagerCommit` is reachable only via a private `_parser` field and private `ProbeEagerCommit()` — driving the eager gate from the shipped grammar is architecturally impossible (0 of 44 tests in that file do). The two patterns are numerically identical (`[req-literal, opt-slot, req-slot, req-literal, req-slot]`), and the `fire` pattern already has independent shipped-grammar coverage via `fire_two_torpedoes_at_bravo_two.wav` in the WAV-replay suite, which runs with eager flush enabled.

---

### What the review positively confirmed

- **The bug was real and this fixes it.** Two angles independently traced base `d081e11` returning `Commit` for `launch all missiles target`, and confirmed `3.0f/5.0f` and the literal `0.6f` are the same float — so `0.6f < 0.6f` is genuinely false and the boundary is not an artifact. The committed repro test fails on base for the stated reason, not incidentally.
- **Every arithmetic claim in the PR body checks out**, including the #65 sequencing figures, which match `Planning~/design-docs/scoring-model.md:113,174` verbatim.
- **Both ticked Unity counts reconcile independently**: `Tests~/Editor` carries exactly 116 test attributes; `Tests~/Runtime` carries 327, one of which is a 16-fixture `[ValueSource]` expansion — `327 − 1 + 16 = 342`.
- **No scoring number moved.** The parser diff is 27 insertions, 0 deletions; the flag is set inside the pre-existing miss branch without touching `rawScore`, `denominator`, `tokenIdx`, `consumedEndIdx`, `literalCount`, or `slotCount`.
- **All seven prior issue-numbered behaviours in this function survive** (#25, #31, #32, #33, #41, #43, #44), each hand-traced. Notably, #44's own comment claimed "everything above has already established one complete, confident match" — that claim was **false on `main`** and is only now true, which makes the new check's placement above it required rather than incidental.
- **No pre-existing test changes outcome.** Every eager test asserting a fire uses a fully-slotted buffer; the two that assert no-fire were already below `minScore`.
- **No platform traps.** The flag cannot desync from the winner (assigned in the same `IsBetterCandidate` block as every other `best*` field); `default(MatchResult)` can never win selection (`Score <= 0` short-circuit at `:880`); the struct grows 28 → 32 bytes, immaterial against the dictionary lookups already in the same call; the test asmdef's Android exclusion does not affect the Editor PlayMode run.

### Scope note (not a defect in this PR)

The fix removes the *early* incomplete fire. It does not remove the eventual one: at or above `minScore`, a missed-slot buffer that the speaker never completes still fires incomplete on the timer path after `bufferWindow`, because the `allowPartialMatch` pending branch requires `score < minScore`. That is pre-existing behaviour and explicitly #65's territory — the CHANGELOG entry states it plainly ("wait the window they would have waited before eager flush was enabled") — but it bounds what merging this changes.

- **On-device (Quest) behaviour** — correctly declared deferred by the author. This is pure parser logic with no bridge, mic-capture, or MonoBehaviour-lifecycle surface.
- **Re-running the Unity suites** — agents were barred from running them (host project, editor-closed, multi-minute batch mode). The 342/342 and 116/116 counts are the author's claims; the review corroborated them by reconciling test-attribute counts, which is short of re-execution.

jinwoo1601:
Pushed `acce532`, addressing all three confirmed findings from the review above.

| Finding | Fix |
|---|---|
| `LINE-1` (nit) | The condition's comment justified itself with "a miss consumes no tokens and so never advances `EndIdx`" — false for a trailing `[unk]` run, since the skip at `:928` runs before every element including one that then matches nothing. Now scoped to *recognised* tokens, with the `[unk]` case named explicitly as making the end-of-buffer condition pass more readily. Corrected in all three artifacts the verifier identified: `VoxrCommandParser.cs:1355`, `Documentation~/scoring.md:214`, `CHANGELOG.md:30`. |
| `GAP-2` (minor) | Added `TryEagerCommit_UnfilledRequiredSlot_UnanalysableGrammar_StillReturnsNone`, pairing `OverLimitCommand()` with the launch pattern. `LaunchParser()` was split into `LaunchSlots()` / `LaunchCommand()` / `LaunchParser()` so both tests share one grammar definition. |
| `XFILE-1` (nit) | `command-recognition.md:300` now says "keeps waiting **the full window**", matching its neighbour, and states explicitly that this case never qualifies for the shortened `prefixHoldSeconds` hold. |

**The `GAP-2` test is a real regression test, not a refactor guard** — confirmed by executing both revisions against the same grammar:

| Revision | `TryEagerCommit("launch all missiles target")` on the degrade grammar |
|---|---|
| base `d081e11` | **`HoldExtendable`** — an incomplete `launch_weapon` arming the shortened hold |
| head `acce532` | **`None`** |

The complete buffer returns `HoldExtendable` on both, confirming the #44 degrade is genuinely active in that grammar rather than the test passing for an unrelated reason. Construction emits exactly one warning, matching the `LogAssert.Expect` regex.

**Re-verified:** Unity PlayMode **343/343** (`failed="0"`, up from 342 — the one new test), EditMode **116/116** unchanged. `VoxrEagerCommitTests` fixture 44 → **45**, all five issue-#66 tests passing.

One note from the re-run worth recording: the green proves the expected warning *did* arrive (an unmet `LogAssert.Expect` fails at teardown), but not that no *additional* warning was emitted — UTF fails on unexpected errors, not unexpected warnings. Pinning that would need `LogAssert.NoUnexpectedReceived()`. Left alone as a pre-existing convention across this suite rather than changed unruled here.

The four refuted findings remain unfixed and unchanged, as reported.

PR #69, opened 2026-08-12
Pin the absence of extra logs in warning-asserting tests (#68)

### LINKED ISSUE #68: Warning-assertion tests pin that a warning fires, but not that no others do
## Summary

Every test that asserts an authoring-time warning uses `LogAssert.Expect(...)` and stops there. `Expect` asserts the named message **arrived**; it says nothing about *other* messages. Because Unity's Test Framework only auto-fails on `Error`/`Assert`/`Exception` — not on `Warning` or `Log` — an **additional, unrelated warning fires completely undetected** in all 16 of those tests.

Surfaced during PR #67 review; not a defect in that PR, which is why it is filed separately.

## The two halves of `LogAssert`

From the framework's own docs (`com.unity.test-framework/UnityEngine.TestRunner/Assertions/LogAssert.cs`):

> A test fails if Unity logs a message other than a regular log or warning message.

So for a `LogType.Warning` expectation, `Expect`'s suppression half is a **no-op** — the warning would never have failed the test anyway. Only its arrival assertion is load-bearing. `NoUnexpectedReceived()` (`LogScope.cs:234`) is the complement: it scans every captured log for the first one not matched by an `Expect` and throws `UnhandledLogMessageException`, **regardless of log type**.

## Current state

The repo already uses both, but in a clean split that leaves one side uncovered. All three `NoUnexpectedReceived()` calls are in *negative* tests asserting silence — none pairs with an `Expect`:

| Test | File:line |
|---|---|
| `OptionalLiteralBeforeSlot_KeepsTheSlotWhenTheLiteralIsDropped` | `Tests~/Runtime/VoxrCommandParserTests.cs:1423` |
| `NonHazardPatternShapes_DoNotWarn` | `Tests~/Runtime/VoxrCommandParserTests.cs:1451` |
| `OptionalLiteralBeforeNumberSequence_RecoversTheElidedLiteral` | `Tests~/Runtime/VoxrCommandParserTests.cs:1525` |

The 16 *positive* tests assert a warning fires, and none pins the absence of others:

| File | Tests |
|---|---|
| `Tests~/Runtime/VoxrCommandParserTests.cs` | `RequiredLiteralBeforeSlot_WarnsAtConstruction` (`:1367`), `RequiredLiteralDropped_BarePatternWinsAndDiscardsTheSpokenSlot` (`:1383`), `OptionalSlotAfterTheLiteral_AlsoWarns` (`:1460`), `MultipleRequiredLiteralsBeforeTheSlot_Warn` (`:1554`), `HazardSplitAcrossTwoIntents_Warns` (`:1582`), `BareFormReachableOnlyByOmittingAnOptional_Warns` (`:1611`) |
| `Tests~/Runtime/VoxrEagerCommitTests.cs` | `ManyOptionalElements_DisablesEagerCommitForWholeParser` (`:331`), `ManyOptionalElements_WarnsAtConstruction_NamingThePattern` (`:356`), `TryEagerCommit_UnanalysableGrammar_ReturnsHoldExtendable` (`:486`), `TryEagerCommit_UnanalysableGrammar_IncompleteSpeech_StillReturnsNone` (`:506`), `TryEagerCommit_UnfilledRequiredSlot_UnanalysableGrammar_StillReturnsNone` (`:817`) |
| `Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs` | `InjectText_BeforeConfigure_LogsWarningAndDoesNotThrow` (`:72`), `PrefixHold_ClearedOnFlush` (`:672`) |
| `Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.cs` | `SecondInitialise_RejectedLoudly_OwnerKeepsBridge` (`:77`), `SecondSetGrammar_RejectedLoudly` (`:131`), `SecondResetRecogniser_RejectedLoudly` (`:146`) |

The last three expect `LogType.Error`, where the suppression half *is* load-bearing — but a stray **warning** in those tests is still invisible.

## Why it is worth closing

The authoring-time warnings are a real, growing product surface (#42, #44), and the `#42` droppable-required-literal check deliberately compares patterns **across intents**, not just within one command's pattern set. So adding a second command to any test grammar can newly trip it. That is not hypothetical: it was an open question during PR #67, whose new test pairs an over-limit command with the launch pattern. It was resolved by driving the real parser outside Unity and observing exactly one warning — but the test itself does not pin that, so a future broadening of the `#42` check would silently start emitting a second warning in these grammars with every suite still green.

## Options

1. **Per-test** — append `LogAssert.NoUnexpectedReceived()` to the 16 tests. Explicit and local; matches the three existing uses; no blast radius. Verbose, and new tests do not inherit it.
2. **`[TestMustExpectAllLogs]` on the affected fixtures** — one attribute per fixture, inherited by new tests in them. Documented as "the same as calling `LogAssert.NoUnexpectedReceived()` at the bottom of every affected test".
3. **`[assembly: TestMustExpectAllLogs]` with `MustExpect=false` opt-outs** — the framework's own documented migration path. Strongest and future-proof, but it applies to **every** test in `Jinwoo1601.VoXR.Tests.Runtime` (343 at time of writing) and promotes stray `Log` messages to failures too.

**Suggested:** determine first how many tests in the Runtime assembly emit incidental `Debug.Log`/`Debug.LogWarning` — that number decides between (2) and (3). If it is small, (3) is the better end state; if it is large, (2) gets the coverage without a broad migration. (1) is the fallback if both prove disruptive.

## Acceptance

- Any test asserting a warning also fails if an *additional* unexpected log is emitted.
- EditMode and PlayMode suites stay green (currently 116/116 and 343/343).
- No change to runtime code — tests only.

## Note on verification

The `LogAssert`/`LogScope`/`TestMustExpectAllLogsAttribute` semantics above were read from the installed `com.unity.test-framework` package source in the host project, not from memory. The call-site inventory was generated by mapping every `LogAssert.*` line to its enclosing test method; re-derive it before acting, since line numbers move.

jinwoo1601:
Closes #68

## What

Adds `LogAssert.NoUnexpectedReceived()` to every test that asserts an authoring-time warning — **option 1** of the three the issue offered — matching the three uses already in the suite. One test also gained a missing second expectation (below).

Tests only; no runtime code touched. No `CHANGELOG` entry: `Tests~` is excluded from the shipped package, so nothing here is user-facing.

## Why

`LogAssert.Expect` has two halves. It asserts the named message **arrived**, and it suppresses that message from failing the test. For a `LogType.Warning` the suppression half is a no-op — the Test Framework only auto-fails on `Error`/`Assert`/`Exception` — so only the arrival assertion was load-bearing, and an additional unrelated warning fired completely undetected.

## Which option, and why the issue's decision tree does not survive contact

The issue said to measure first: count the tests emitting incidental logs, and let that number pick between (2) fixture-level and (3) assembly-level `TestMustExpectAllLogs`.

Measured with `[assembly: TestMustExpectAllLogs]` on the runtime test assembly: **81 of 343 PlayMode tests fail**, all `UnhandledLogMessageException`, all on `[Warning]`-level logs, zero other breakage. Not small — but the shape matters more than the count:

| Fixture | Failures | Contains warning-asserting tests? |
|---|---|---|
| `VoxrCommandParserTests` | 66 | yes (6) |
| `VoxrCommandRecogniserInjectionTests` | 11 | yes (4) |
| `VoxrEagerCommitTests` | 3 | yes (5) |
| `VoxrAssetConversionTests` | 1 | no |

The issue's rule was "if large, (2) gets the coverage without a broad migration." It does not: the incidental warnings sit **inside exactly the fixtures option 2 would annotate**, so option 2 costs the same ~80-test migration as option 3. And the noise is not per-test — it comes from shared grammar helpers, with a single-character alias `"a"` in one slot driving 56 of the 81. Those cannot be cleaned up without changing what the tests exercise; `Alias_QuantityA_ResolvesToOne` exists precisely to test that alias.

So the decision tree runs off the end and lands on its own stated fallback, (1).

## Two corrections to the issue's inventory

Both surfaced by re-deriving the call-site map, as the issue advised.

**1. It is 18 warning-asserting tests, not 16.** The expectation the issue attributed to `PrefixHold_ClearedOnFlush` is actually in the shared helper `ConfigureForUnanalysableGrammar`, which serves three `UnanalysableGrammar_*` tests. `PrefixHold_ClearedOnFlush` has no `Expect` of its own — it is an ordinary incidental-warning test, not one of the targets.

**2. The measurement caught a live instance of the defect.** Those three tests expected only `"more than the 12"`, while their grammar emits a **second** warning first: the #42 droppable-required-literal check fires because `"status"` is a bare form of `"status report {target}"` — which is the very shape the prefix-hold fixture exists to exercise, so it is correct and cannot be designed away. Both warnings are now expected.

Order is load-bearing there: `LogScope.ExpectedLogs` is a **queue**, matched against logs in emission order (`LogScope.cs:206-227`), and the parser ctor runs the #42 scan (`VoxrCommandParser.cs:229`) before the optional-expansion one (`:230`). The new expectation had to go *before* the existing one. A comment in the helper records this so a future edit does not reorder them.

## Validation

- [x] Unity **EditMode** — 116/116 passed, 0 failed, exit 0 (baseline match)
- [x] Unity **PlayMode** — 343/343 passed, 0 failed, exit 0 (baseline match)
- [x] Both runs verified fresh, not stale XML — a tests-only change reproduces baseline counts exactly, so stale artifacts were moved aside first and each result's UTC `start-time` confirmed after the cutoff
- [x] Zero `Unhandled log message` failures in the PlayMode log
- [x] No compile errors in either assembly
- [x] Working tree reverted after the run: `Tests` → `Tests~`, all 58 generated `.meta` orphans plus root `Tests.meta` deleted, host manifest untouched
- N/A NativeBridge gates — `NativeBridge~/` is unchanged by this branch
- N/A On-device (Quest) check — no runtime code changed; nothing here can affect device behaviour

## Note for whoever picks up the follow-up

Unity's own doc comment for `TestMustExpectAllLogsAttribute` — and the issue's option 3, which quotes it — shows the opt-out as `[TestMustExpectAllLogs(MustExpect=false)]`. **That does not compile.** `MustExpect` is a get-only property (`TestMustExpectAllLogsAttribute.cs:33`); the opt-out has to be positional, `[TestMustExpectAllLogs(false)]`. Worth knowing before anyone attempts the broader migration.

Also worth recording while the data is fresh: the 262 PlayMode tests that already pass under assembly-wide strictness could be locked in cheaply, since only 4 fixtures are dirty. That is a larger behavioural change than this issue asked for, so it is left as a decision rather than folded in here.

### DIFF
--- a/Tests~/Runtime/VoxrCommandParserTests.cs
+++ b/Tests~/Runtime/VoxrCommandParserTests.cs
@@ -1372,6 +1372,7 @@ namespace VoXR.Tests.Runtime
                 parser,
                 "the shape is a warning, not an error — construction still succeeds"
             );
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -1394,6 +1395,7 @@ namespace VoXR.Tests.Runtime
                 result.Command.HasSlot("burn_level"),
                 "the spoken burn level is discarded with nothing to signal it"
             );
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -1471,6 +1473,7 @@ namespace VoXR.Tests.Runtime
             );
 
             Assert.IsNotNull(parser);
+            LogAssert.NoUnexpectedReceived();
         }
 
         // Mirrors the shipped set_heading grammar (DemoGrammar / CommandDemo /
@@ -1572,6 +1575,7 @@ namespace VoXR.Tests.Runtime
             );
 
             Assert.IsNotNull(parser);
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -1600,6 +1604,7 @@ namespace VoXR.Tests.Runtime
                 "the bare intent wins and the spoken burn level is stranded"
             );
             Assert.IsFalse(result.Command.HasSlot("burn_level"));
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -1637,6 +1642,7 @@ namespace VoXR.Tests.Runtime
                 "the bare form wins at 1.0 and the target is stranded"
             );
             Assert.IsFalse(result.Command.HasSlot("target"));
+            LogAssert.NoUnexpectedReceived();
         }
     }
 }
--- a/Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs
+++ b/Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs
@@ -72,6 +72,7 @@ namespace VoXR.Tests.Runtime
             LogAssert.Expect(LogType.Warning, new Regex("InjectText called before parser is ready"));
 
             Assert.DoesNotThrow(() => _recogniser.InjectText("anything"));
+            LogAssert.NoUnexpectedReceived();
         }
 
         [TestCase(null)]
@@ -668,7 +669,17 @@ namespace VoXR.Tests.Runtime
 
         void ConfigureForUnanalysableGrammar(float prefixHold)
         {
-            // Construction is where the over-limit pattern is reported now.
+            // Construction is where the over-limit pattern is reported now. Two warnings land
+            // here, not one: the prefix-hold grammar this builds on is itself a droppable-
+            // required-literal shape ("status" is a bare form of "status report {target}"),
+            // which is the very thing the hold exists to exercise, so the #42 check fires too.
+            // Expectations are matched in queue order against the logs in emission order, and
+            // the ctor runs the #42 scan before the optional-expansion one — so this pair must
+            // stay in this order.
+            LogAssert.Expect(
+                LogType.Warning,
+                new Regex(@"""status"" \(intent 'status'\) is a bare form of")
+            );
             LogAssert.Expect(LogType.Warning, new Regex("more than the 12"));
 
             _recogniser.Configure(PrefixHoldSlots(), UnanalysableCommands());
@@ -697,6 +708,7 @@ namespace VoXR.Tests.Runtime
                 "nothing commits early on a grammar the eligibility analysis never vetted"
             );
             Assert.AreEqual(0.6f, _recogniser.TestEffectiveBufferWindow, 1e-4f);
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -708,6 +720,7 @@ namespace VoXR.Tests.Runtime
             _recogniser.InjectText("cease fire");
 
             Assert.AreEqual(2.0f, _recogniser.TestEffectiveBufferWindow, 1e-4f);
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -720,6 +733,7 @@ namespace VoXR.Tests.Runtime
             _recogniser.InjectText("cease");
 
             Assert.AreEqual(2.0f, _recogniser.TestEffectiveBufferWindow, 1e-4f);
+            LogAssert.NoUnexpectedReceived();
         }
     }
 }
--- a/Tests~/Runtime/VoxrEagerCommitTests.cs
+++ b/Tests~/Runtime/VoxrEagerCommitTests.cs
@@ -345,6 +345,7 @@ namespace VoXR.Tests.Runtime
             Assert.IsFalse(overLimit, "the over-limit pattern is disabled");
             Assert.IsFalse(normal,
                 "a normal command in the same set is also disabled (whole-parser, never partial)");
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -364,6 +365,7 @@ namespace VoXR.Tests.Runtime
             );
 
             Assert.IsNotNull(parser);
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -495,6 +497,7 @@ namespace VoXR.Tests.Runtime
                 parser.TryEagerCommit(Tok("cease fire"), null, 0.6f, 0.4f),
                 "an un-analysable grammar holds — it never commits early"
             );
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -519,6 +522,7 @@ namespace VoXR.Tests.Runtime
                 parser.TryEagerCommit(Tok("banana"), null, 0.6f, 0.4f),
                 "speech matching nothing must not arm the shortened hold"
             );
+            LogAssert.NoUnexpectedReceived();
         }
 
         // ---------- Tie-break parity with ParseInternal (Review MF-5) ----------
@@ -826,6 +830,7 @@ namespace VoXR.Tests.Runtime
                 parser.TryEagerCommit(Tok("launch all missiles target"), null, 0.6f, 0.4f),
                 "an incomplete command must not reach the degrade's hold either"
             );
+            LogAssert.NoUnexpectedReceived();
         }
     }
 }
--- a/Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.cs
+++ b/Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.cs
@@ -99,6 +99,7 @@ namespace VoXR.Tests.Runtime
                 _ownerErrors,
                 $"Owner must see no error: [{string.Join("; ", _ownerErrors)}]"
             );
+            LogAssert.NoUnexpectedReceived();
         }
 
         [UnityTest]
@@ -138,6 +139,7 @@ namespace VoXR.Tests.Runtime
                 $"Expected exactly one rejection: [{string.Join("; ", _secondErrors)}]"
             );
             Assert.IsTrue(_owner.IsInitialised, "Owner must still hold the bridge.");
+            LogAssert.NoUnexpectedReceived();
         }
 
         [Test]
@@ -153,6 +155,7 @@ namespace VoXR.Tests.Runtime
                 $"Expected exactly one rejection: [{string.Join("; ", _secondErrors)}]"
             );
             Assert.IsTrue(_owner.IsInitialised, "Owner must still hold the bridge.");
+            LogAssert.NoUnexpectedReceived();
         }
 
         [UnityTest]


jinwoo1601 (commented):
## Independent review — PR #69

**Head reviewed:** `896bf8cc86b3ec43f336ff5128b5debe01b2e409`
**Base:** merge-base `5a0359261f2d14cc621c1800e257e86ce19d1431` against `main`

Six finder agents ran with fresh context against the diff, inheriting nothing from the session that wrote it. Six findings survived dedup; each was then handed to an adversarial verifier told to refute it. **Four confirmed — all nits. Two refuted. No blockers, no majors, no minors, and no code defect of any kind.**

Four of the six angles (`LINE`, `GONE`, `XFILE`, `PIT`) returned **zero findings**.

---

### Confirmed findings

All four are documentation/wording. None affects code, tests, or runtime behaviour. None blocks merge.

#### 1. `nit` — PR body says "one test" gained the second expectation; it was one *helper*, affecting three tests

The "## What" section reads *"One test also gained a missing second expectation (below)."* The expectation is not in a test — it is in the shared helper `ConfigureForUnanalysableGrammar` (`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:670`), called by three tests (`:695`, `:718`, `:731`). The body's own Correction 2 says "those three tests", so the document contradicts itself.

The verifier knocked down the finder's "understates the blast radius by 3×" framing: the bullet carries an explicit `(below)` pointer, and a reader who follows it gets the correct picture. **What survives is narrow — a wrong numeral and a misattributed location in a summary sentence that immediately points at its own correction.**

#### 2. `nit` — the body's stated rationale does not describe 3 of its 18 changes

The title says "warning-asserting tests" and the body says "every test that asserts an authoring-time warning", but three of the eighteen assert an **error**: `LogAssert.Expect(LogType.Error, DuplicateOwner)` at `Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.cs:77`, `:132`, `:148`.

The finder originally charged the PR with "silently folding them into a warning framing". **The verifier refuted that narrative outright**: issue #68 uses the warning-only framing in its own title, summary, inventory header, and acceptance criterion, and confines the `LogType.Error` caveat to one footnote. The PR inherited the canon's wording rather than inventing it.

What survives: the body never carries over that caveat, so its stated *Why* — "for a `LogType.Warning` the suppression half is a no-op" — is inapplicable to those three, where suppression **is** load-bearing and only a stray *warning* was ever invisible. The ownership fixture is never named in the body. No behavioural consequence: `NoUnexpectedReceived()` is log-type agnostic.

#### 3. `nit` — `OptionalElementsAtTheLimit_DoNotWarn` asserts silence in its name only

`Tests~/Runtime/VoxrEagerCommitTests.cs:371-385` exists to pin that an at-the-limit pattern **stays silent**, but observes no logs. Since the framework auto-fails only on `Error`/`Assert`/`Exception`, it would still pass if that pattern started warning. Its counterpart `NonHazardPatternShapes_DoNotWarn` (`Tests~/Runtime/VoxrCommandParserTests.cs:1429`) does close with `NoUnexpectedReceived()` at `:1453`.

Verified useful detail: **the one-line fix would pass.** The grammar provably emits nothing — `Slots()` is empty, neither form is an element-prefix of the other, and `AtLimitCommand()` has exactly 12 optionals against `MaxOptionalExpansion = 12`.

Downgraded minor → nit on two grounds. It is **out of issue #68's acceptance** (which covers tests that *assert* a warning; this one executes no `Expect`), and the PR did not touch it — the file's five hunks straddle the method without entering it. Also, the regression class is largely already caught: the warn threshold and the analysis-abandon threshold share one constant, so a change to it fails the existing `Assert.IsTrue` at `:381` anyway. Only a warn-site-only drift would slip through.

#### 4. `nit` — the helper documents its ordering constraint but not the obligation it places on callers

`ConfigureForUnanalysableGrammar` (`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:670`) holds both `Expect` calls; the matching `NoUnexpectedReceived()` lives in each of the three callers. The new comment block (`:672-678`) documents the queue-ordering constraint meticulously and says nothing about why the complement sits at the call sites. A fourth caller added later would inherit the expectations and silently not inherit the check — and gets no lexical prompt, since the word `LogAssert` never appears at the invocation.

The verifier established that **the obvious remedy is wrong**, which is worth recording: moving `NoUnexpectedReceived()` into the helper would scan only the `Configure`-time logs and be blind to anything the test body emits afterwards — `VoxrCommandRecogniser.cs:229` can log during `InjectText` — manufacturing exactly the half-check this issue exists to kill. A fixture-level `[TestMustExpectAllLogs]` is also unavailable: it applies to all 34 tests in the fixture, 11 of which emit incidental warnings. **The only in-scope remedy is one more sentence in the existing comment.**

---

### Refuted — recorded so you can see what was considered and killed

- **"The convention this PR establishes is nowhere recorded or enforced."** Premise inverted: the convention **pre-exists at the merge-base** — three bare `NoUnexpectedReceived()` calls, equally uncommented, at base `:1423/:1451/:1525`. The PR propagated house style rather than inventing an undocumented one. The finding also restated a tradeoff issue #68 had already accepted in writing ("Verbose, and new tests do not inherit it"), and its proposed remedy *was* option 2 — the option the PR's own measurement ruled out. The finder's tally of "15 added sites" was also wrong; it is 18.

- **"An asymmetry was left inside the otherwise-strict #42 region."** `OptionalLiteralBeforeNumberSequence_AlsoClaimsAnUnmarkedStrayDigit` lacks the call its twin has — but **the twin's call is pre-existing at the merge-base (`:1525`)**, and the PR touched neither half of the pair. No asymmetry was introduced. The ask is also redundant: both build the identical `HeadingParser()`, so the twin's existing assertion already pins that grammar's silence.

---

### Claims audited and found supported

Recorded because a review that only lists defects hides how much was actually checked.

- **All 18 warning-asserting tests are covered, and 18 is the right number.** Independently re-derived by three separate agents. `Tests~/Editor/` contains **zero** `LogAssert` call sites, so no EditMode test qualifies — that question is closed, not merely unexamined.
- **The corrected inventory holds.** `PrefixHold_ClearedOnFlush` genuinely has no `Expect` of its own; the expectation issue #68 attributed to it is in the shared helper.
- **The new regex matches the emitted text.** Reconstructed character-by-character from `BuildDroppableLiteralWarning` (`Runtime/Commands/VoxrCommandParser.cs:446`) against the test's actual grammar, including that the cross-intent branch is taken and the bare/extended roles are the way round the regex assumes. Escaping is correct for both the C# verbatim string and `Regex` metacharacters.
- **The ordering claim is exact, and better guarded than the PR body implies.** `VoxrCommandParser.cs:229` does precede `:230`. Beyond the comment, a reorder is caught *mechanically*: `EvaluateLogScope(true)` runs after every test regardless of `MustExpect` and throws on any expectation left queued, so a swap fails all three tests twice over.
- **No test emits an uncovered log.** Every one of the 18 had its complete log set derived against all reachable warning sites.
- **No fragility from Unity lifecycle or setup noise.** `UnityLogCheckDelegatingCommand` wraps the test method *innermost*, so `[SetUp]`/`[UnitySetUp]`/`[UnityTearDown]` logs land in a separate scope — the ownership fixture's heavyweight async setup cannot leak in. The one `[UnityTest]` completes synchronously (`InitialiseAsync` returns before its first `await`), so its check is reached. No newly-strict test depends on a `#if DEBUG` or `Debug.isDebugBuild` gate.
- **Strictly additive.** One line removed repo-wide, a comment, fully subsumed by its replacement. No existing `Expect`, assertion, or test was weakened, broadened, reordered, or deleted; the three pre-existing `NoUnexpectedReceived()` calls are intact. No stray `[assembly: TestMustExpectAllLogs]` was left behind from the measurement.
- **`MustExpect` really is get-only** (`TestMustExpectAllLogsAttribute.cs:33`), so the PR's warning about `[TestMustExpectAllLogs(MustExpect=false)]` not compiling is correct.
- **The measurement numbers check out.** 81/343, the per-fixture split 66/11/3/1, "all `[Warning]`", and "56 from the single-character alias" were each corroborated against the run artifacts. Caveat for the record: **those artifacts live outside the repo, so a future reader cannot re-derive them from the PR alone.**

---

- **The ticked EditMode/PlayMode boxes (116/116, 343/343).** Reviewing agents were instructed not to run Unity; every analysis above is static. The environment *can* produce those runs (host project per the project bindings) and the artifacts are consistent with the claim, but this review does not independently re-execute them.
- **On-device (Quest) behaviour** — human-only, always. The PR marks it N/A on the grounds that no runtime code changed, which the diff supports.

---

*Slim profile, issue lane. No blocker surfaced, so nothing was fixed on the branch in response to this review — all four nits are left unfixed and unruled for you. Not approved and not merged; both are yours.*

PR #71, opened 2026-08-12
Refuse an eager commit when a required tail is unmatched (#70)

### LINKED ISSUE #70: Eager flush can commit a pattern whose terminal required literal was never spoken
## Summary

The eager-flush gate refuses to commit when a required **slot** is unfilled (#66). The same hole is open one element class over: a required **literal** that was never spoken also consumes no token, so the gate's end-of-buffer condition is satisfied vacuously and the command commits while the speaker is still mid-utterance — sometimes as the *wrong* command.

## Symptom

The demo grammar registers `["switch","to","weapons"]` (`Tests~/Runtime/DemoGrammar.cs:136`) and `["switch","to","navigation"]` (`:143`). Say *"switch to navigation"*. At the moment the buffer holds `switch to`, the first pattern matches "switch" and "to" and misses "weapons", scoring `(1 + 1 − 0.5) / 3 = 0.50` — under the default `minScore` of `0.60`, so the verdict is `None`. It is saved by the arithmetic alone.

At four elements the arithmetic stops saving it. A pattern `[a, b, c, d]` of required literals, with `d` unspoken, scores `(3 − 0.5) / 4 = 0.625` — over the gate — and every other eager condition passes:

- `MissedRequiredSlot` is false (no slots involved), so #66's guard at `Runtime/Commands/VoxrCommandParser.cs:1367` never fires.
- `bestEndIdx == tokens.Length` holds, because the miss consumed no token and `EndIdx` never advanced past what actually matched (`:1386`).
- The start-at-first-recognised-token check, the confidence check, and `CanCommitEarly` all pass.

So the gate commits a command the speaker has not finished saying. Where a sibling pattern shares the prefix, the committed intent can be the **wrong one** — the tie is broken by registration order at `:884-891`.

## Root cause

Neither `EndIdx` nor `ConsumedEndIdx` can distinguish *"the pattern ended exactly at the buffer end"* from *"the pattern ran out of buffer with elements still owing"*, because a miss consumes nothing and leaves both indices sitting at the buffer end.

The comment at `:1363-1366` exempts literals deliberately, reasoning that the command "is still fully determined and its arguments are all present". That is true when the dropped literal is **medial**. It is false when the dropped literal is the **terminal, discriminating** word — which is exactly the word the speaker is about to say next.

## Fix

Add a completeness condition alongside #66's: **no eager verdict above `None` when any required pattern element sits after the last element that actually matched.**

Medial drops keep committing early — all slots filled, pattern tail landed at the buffer end — so only the "buffer ended where the pattern still owes a word" case is blocked. #66's slot rule stays exactly as it is: it is strictly stronger for slots, since a medial missed slot means an absent argument and must block wherever it occurs.

## Notes

- Existing eager coverage cannot catch this: every fixture in `Tests~/Runtime/VoxrEagerCommitTests.cs` uses two-element all-literal patterns or slot-terminated ones, so the two-element case (`"cease"` → `0.5`) passes by one element's margin.
- This is a prerequisite for the first backlog item of #65. Zeroing `RequiredLiteralMissPenalty` widens this hole from `N≥4` to `N≥3` (`2/3 = 0.667`), which is where two-word-prefix grammars live — the `switch to` case above becomes live.

jinwoo1601:
Closes #70

## What

A verdict above `None` from `TryEagerCommit` now additionally requires that **no required element sit after the last element that actually matched**.

- `MatchResult` gains `HasUnmatchedRequiredTail`.
- `TryMatchScored` tracks `requiredAfterLastMatch` — reset by every element that matches, incremented by a missed required literal *or* slot. Non-zero at the end means the pattern's **tail** is what went unmatched; a medial miss is followed by a match and clears it.
- `TryEagerCommit` refuses on it, placed immediately after the #66 slot guard so it sits **above** the #44 degrade — the same ordering requirement #66 had, for the same reason.
- The comment claiming required literals are "deliberately exempt" is corrected to the medial-vs-tail distinction it should always have been.
- `Documentation~/scoring.md` §6 gains the tail rule as condition 5 (confidence renumbers to 6), and condition 4 stops claiming required literals are exempt outright — they are exempt from *that* condition only. `Documentation~/command-recognition.md` gains the matching user-facing bullet. Both are the sites #66 updated in the same commit as its parser change.

Nine new tests: **seven** in a `// ---------- Unmatched required tail (issue #70) ----------` region of `VoxrEagerCommitTests.cs`, **one** added to that file's issue #66 region covering a *medial* missed required slot (which the tail condition deliberately cannot see), and **one** recogniser-level test in `VoxrCommandRecogniserInjectionTests.cs` pinning that the right command fires once the last word lands. Plus one corrected comment on the existing medial-miss test.

## Why

The whole-buffer condition is satisfied **vacuously** by a pattern still owing words. A miss consumes no token, so `EndIdx` — and `ConsumedEndIdx` — sit at the buffer end exactly as they do for a pattern that genuinely finished there. Neither index can tell the two apart. This is the hole #66 closed for slots, one element class over.

The shipped demo grammar carries the shape: `switch to weapons` and `switch to navigation` diverge only on their last word, so a speaker two words into "switch to navigation" presents a buffer both patterns match at `(1 + 1 - 0.5) / 3 = 0.50`, tying, with registration order picking `mode_weapons`. The **wrong** command, not merely an early one.

At three elements the default `minScore` of `0.60` refuses that today — by arithmetic, not by rule. At **four** elements the arithmetic stops helping: `(3 - 0.5) / 4 = 0.625` clears the gate with every other eager condition satisfied. The bug is live on `main` today.

A medial miss is deliberately still allowed, which is why this is a tail test rather than a ban on missed literals: "launch all missiles hotel one" drops the "target" function word yet fills every slot and still lands its final element on the buffer's last token, so nothing the speaker says next was owed to it.

Scoring, selection and `ParseInternal` are untouched — these buffers remain ordinary parse results and simply wait the `bufferWindow` they would have waited before eager flush was enabled.

## Validation

- [x] Unity **EditMode 116/116** and **PlayMode 352/352** — 468 tests, zero failures, run on the final tree via the host project (`VoXR TestGround`, Unity 6000.4.7f1) per the project's verification bindings.
- [x] `TryEagerCommit_UnmatchedTerminalLiteral_LiveAtDefaultThreshold` pins the four-element `0.625` case at the **default** `minScore`, so the fix is proven against the hole as it exists on `main`.
- [x] The headline demo-grammar test lowers `minScore` to `0.4` deliberately: at the `0.6` default that three-element buffer is caught by the score gate today, which would make the test green **without** the fix.
- [x] `TryEagerCommit_MissedRequiredLiteral_AllSlotsFilled_StillCommits` (pre-existing, #66) still passes — the regression guard on this not over-reaching into medial misses.
- [x] No `NativeBridge~/` change, so the arm64 build and desktop harness are out of scope and were not run.
- On-device (Quest) check **not** required: this touches neither capture, the bridge, nor MonoBehaviour lifecycle.

## Context

Ruled as DR-6 of the scoring-model design and routed here on the #66 precedent, as a live bug on `main` rather than part of that redesign. It is a prerequisite for the first backlog item of #65 — zeroing `RequiredLiteralMissPenalty` widens this hole from `N>=4` to `N>=3` (`2/3 = 0.667`), which is where two-word-prefix grammars live.

## Review

An independent `review-pr` pass (slim profile) ran against `6df391a` and posted its findings as a comment review. It returned **no blockers** — `LINE`, `GONE` and `PIT` each found nothing in the fix itself — and eight confirmed findings below blocker. **All eight are addressed in `f609c4c`**; see the follow-up comment for the mapping.


jinwoo1601 (commented):
## Independent review — PR #71

**Head reviewed:** `6df391a9da976cc91433559bf0067ff6640be8a3` · **merge-base:** `daeb94bdafb65bc77cc85e57fccb9085f1fb010c`

Finder agents ran with fresh context and no knowledge of the session that wrote the fix. Every finding below carries a verifier verdict; three were checked by compiling the real parser standalone (`dotnet` + a `UnityEngine.Debug` stub) and running the buffers at **both** head and base, so the before/after claims are measured rather than argued.

**No blockers.** The core change verified clean on three separate angles.

---

### The fix itself — verified correct

- **`LINE`** re-derived all seven branches of the counter and traced eight adversarial pattern/token pairs (every element missing; trailing optional after a required tail miss; optional matching after a required miss; medial miss; trailing `[unk]`; empty pattern) without breaking it. `HasUnmatchedRequiredTail` is bidirectionally equivalent to "a required element sits after the last element that actually matched" — no branch resets without a real match, none increments without a required miss. A zero-consumption slot "match" that would spuriously clear the counter is unreachable (`minWords >= 1` is enforced at construction).
- **`GONE`** established the newly-refused set is exactly *"buffer exhausted, or exhausted across trailing `[unk]`, with trailing required literals unspoken"* — issue #70's target and nothing wider. It re-traced all 14 pre-existing `Commit`/`HoldExtendable` assertions plus the prefix-hold integration tests; every winner carries a false tail flag. Notably it also confirmed a complete candidate can never lose its commit to an incomplete sibling: a complete match normalises to `1.0` and any miss strictly lowers the ratio, so at equal `startIdx` the complete one always wins — the incomplete one wins only at an *earlier* start, which is precisely the case that used to fire the wrong command.
- **`PIT`** cleared struct-construction completeness, the stale-copy hazard across all nine `best*` locals, `in`-parameter defensive copies, and `LogAssert` usage. The struct is **32 bytes before and after** — both bools land in pre-existing padding — so there is no hot-loop cost beyond one extra branch per element.

---

### Findings

#### 1. `Documentation~/scoring.md:214` — stale condition list carrying an affirmatively false clause · **major** · CONFIRMED

§6 opens *"A verdict above `None` requires **all** of:"* and lists five conditions. The tail rule is a sixth — the guard sits **above** the end-of-buffer check, so it is an independent requirement, not an implication of an existing one — and it appears nowhere: `grep` across `Documentation~/` finds no mention of it at head, while `CHANGELOG.md:30` documents it fully. The two product docs this PR touches and skips now disagree with each other.

Worse, condition 4 ends: *"Required *literals* are exempt — a dropped function word leaves the command fully determined."* This PR deleted that exact sentence from the parser comment as wrong, and its own headline case is a dropped function word leaving the command **un**determined.

**Precedent, verified:** `git show d57c764 --stat` shows #66 editing `Documentation~/scoring.md`, `Documentation~/command-recognition.md` and the parser **in one commit**, and `git log -S` confirms that condition list has exactly one author. #66 made structurally the identical edit — inserted its condition 4, renumbered confidence 4→5.

**The product-docs-are-post-G2 exemption was tested and does not apply.** The project CLAUDE.md binds `Documentation~/` and `CHANGELOG.md` together as one product-doc bucket, and this PR updates `CHANGELOG.md`; the deferral is bound to G2, and the issue lane is gate-exempt by that same file; #66 and #68 both ran this lane and shipped their doc edits inline.

**Fix:** add the sixth condition, renumber confidence to 6, and narrow condition 4's closing sentence to *medial* misses.

#### 2. `Documentation~/command-recognition.md:300` — no user-facing bullet, and the neighbour misdirects · **minor** · CONFIRMED

#66 added its case to this list; #70 has none. The adjacent bullet at `:302` explains waiting by *"anything left over at the end — recognised or `[unk]` — is treated as an in-progress tail"* — but that defines a tail as leftover **buffer tokens**, while #70's rule is about leftover **pattern elements**, which the parser's own comment says are indistinguishable by index. A reader asking "why doesn't my command fire early any more?" checks that bullet, sees nothing left over, and concludes it should. It is the only place the doc uses "tail" and it now names the wrong sense.

#### 3. `Tests~/Runtime/VoxrEagerCommitTests.cs:769` — #66's slot rule now has no discriminating test · **minor** · CONFIRMED

Both #66 refusal pins use a pattern whose missed slot is **terminal**, so this PR's tail guard refuses them unaided. **Measured:** with `VoxrCommandParser.cs:1388-1389` deleted, all **31** verdict assertions in the suite stay green. #66's rule remains load-bearing for a *medial* slot miss, which nothing covers.

The original finder's example was **refuted** — `["launch","{weapon}","target","{target}"]` on "launch target hotel one" scores `0.5` and dies at the score gate regardless. A medial slot miss needs five required elements to clear `0.60`: pattern `("launch","{weapon}","target","{target}","now")` on "launch target hotel one now" scores exactly `0.60` with the tail flag false — head `None`, guard-deleted **`Commit`**. That is the test worth adding.

#### 4. `Runtime/Commands/VoxrCommandParser.cs:979` — the slot half of the counter is currently unreachable · **minor** · CONFIRMED

`requiredAfterLastMatch++` sits in the same block that sets `missedRequiredSlot = true`, so every candidate reaching it is already refused one guard earlier. Boolean-difference analysis: removing it can only flip the flag for exactly the population `:1388` already refuses.

**Verifier came down against removal.** Dropping it would make the field's own doc comment false — it is documented as "any REQUIRED element", not "any required literal" — and would couple the tail guard to the continued existence and placement of #66's guard. Keep it as defensive symmetry; add one line saying it is presently dominated.

#### 5. `Tests~/Runtime/VoxrEagerCommitTests.cs:872` — no trailing-`[unk]` tail-miss test · **nit** · CONFIRMED

Measured at both revisions: `"set auto pilot [unk]"` vs `("set","auto","pilot","on")` gives `Score 0.625, EndIdx 4, ConsumedEndIdx 3, HasUnmatchedRequiredTail True` → head `None`, **base `Commit`**. Neither the score gate nor the whole-buffer check refuses it, so only the new guard can. Production code is correct; only coverage is missing.

Two honest qualifications from the verifier: the reachable shape is a *final* result carrying trailing filler, not a VOSK partial (partials only assign `LastPartialResult`; the eager probe runs on final results). And the variant is a weaker discriminator than the existing test except against one specific mis-implementation — a guard qualified with `&& bestConsumedEndIdx == tokens.Length`.

#### 6. `Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs` — the wrong-command harm is pinned nowhere · **nit** · CONFIRMED

`TryEagerCommit` returns an enum, so no parser-level test *can* pin which intent fires — yet "the wrong command fires, not merely an early one" is the issue's headline harm and the parser's own comment.

Two of the finding's three supporting claims were **refuted**: three of the six new tests do assert `Commit`/`HoldExtendable`, not only `None`; and the proposed `"switch to"` → `"navigation"` test would have been **vacuous**, because `minScore` is not exposed by the test-only setter block, so that buffer's `0.50` dies at the score gate pre-fix too. A working version costs ~26 lines using the 4-literal autopilot grammar at the default threshold, where pre-fix behaviour is a genuine wrong-command fire.

#### 7. PR body — "Five new tests"; the region contains six · **nit** · CONFIRMED

Counted directly: six `[Test]` methods. The validation tally corroborates the code, not the prose — `main`'s baseline is 343 PlayMode (PR #69), this PR ticks 349.

#### 8. `Runtime/Commands/VoxrCommandParser.cs` — field comment imprecise · **nit** · CONFIRMED

The comment says a tail miss "leaves **both** of them at the buffer end". False when trailing `[unk]` exists: `EndIdx` passes the filler, `ConsumedEndIdx` does not — as finding 5's measurement shows (4 vs 3). The sentence explaining the mechanism gets the mechanism slightly wrong.

---

### Claims audited and upheld

`CLAIM` verified the guard placement, the counter semantics, the "scoring/selection/`ParseInternal` untouched" claim (the field is written once and read twice, and `IsBetterCandidate` never sees it), and independently re-derived the `0.625` four-element case as `Commit` on base and `None` at head. `XFILE` confirmed the registration-order claim by reading the shipped sample assets — `Set_Common.asset` lists `Cmd_ModeWeapons`' guid before `Cmd_ModeNavigation`'s — and confirmed `MatchResult` never crosses a file boundary, with no editor tooling or recogniser code duplicating the eager conditions.

- **The ticked EditMode/PlayMode counts.** This is a bare UPM package; `Tests~` cannot run standalone, and the run was performed against a host project unreachable from the review environment. The counts were audited for internal consistency (343 + 6 = 349) and found plausible, not re-executed.
- **On-device (Quest) behaviour.** The PR's assertion that no device check is needed is consistent with the diff — nothing touches capture, the bridge, or MonoBehaviour lifecycle.

jinwoo1601:
### Review findings addressed — `f609c4c`

All eight confirmed findings from the review above are fixed. Mapping:

| # | Finding | Addressed by |
|---|---|---|
| 1 | `scoring.md:214` stale list + false clause (**major**) | §6 gains the tail rule as **condition 5**, confidence renumbers to **6**, and condition 4's closing sentence narrows to "exempt from **this** condition … but see condition 5" |
| 2 | `command-recognition.md:300` missing bullet | New "still owing its last word" bullet, with the medial contrast, before the split-command entry |
| 3 | #66's slot rule had no discriminating test | `TryEagerCommit_MedialUnfilledRequiredSlot_ReturnsNone` — 5 required elements, medial `{weapon}` miss, score exactly `0.60`, tail flag false, so the #66 guard is the only thing refusing it |
| 4 | Dominated slot increment unexplained | Comment at the increment: currently dominated by the #66 guard, kept so the field stays true to its name and the tail condition does not depend on that guard's placement |
| 5 | No trailing-`[unk]` tail-miss test | `TryEagerCommit_UnmatchedTerminalLiteral_TrailingUnk_ReturnsNone` — `EndIdx` 4, `ConsumedEndIdx` 3, score `0.625` |
| 6 | Wrong-command harm pinned nowhere | `EagerFlush_PatternStillOwingItsLastWord_FiresTheRightCommand` in `VoxrCommandRecogniserInjectionTests.cs` — "set auto pilot" → 0 fires, then "off" → 1 fire asserting `Intent == "autopilot_off"` |
| 7 | PR body test count | Body corrected. It said five; the honest number is **nine** — seven in the #70 region, one in the #66 region, one recogniser-level |
| 8 | Field comment said "both" indices sit at the buffer end | Rewritten: `EndIdx` stops where a genuinely-finished pattern would, and trailing `[unk]` makes that *harder* to see, not easier |

Two fixes deliberately departed from what the finder proposed, on the verifiers' reasoning:

- **#4 is a comment, not a removal.** Deleting the dominated increment would falsify the field's own documented meaning ("any required **element**", not "any required literal") and would couple the tail condition to the continued existence and placement of the #66 guard.
- **#6 uses a different grammar than proposed.** The suggested "switch to" → "navigation" test would have been green without the fix: `minScore` is not exposed by the test-only setter block, so that buffer's `0.50` dies at the score gate either way. The 4-literal autopilot grammar at the **default** threshold (`0.625`) is genuinely red pre-fix.

Finding #3's proposed example was also refuted during verification — `["launch","{weapon}","target","{target}"]` on "launch target hotel one" scores `0.5` and dies at the score gate regardless — so the committed test uses the 5-element form that actually discriminates.

**Verification:** EditMode **116/116**, PlayMode **352/352** — 468 tests, zero failures, on the final tree.

PR #72, opened 2026-08-13
Stop charging a missed required literal twice (#65 §5.1, DR-7)

jinwoo1601:
Implements §5.1 of the locked scoring-model design (issue #65, backlog item 1), plus **DR-7**, an amendment the build itself forced.

Closes the cases #65 reports under **symptom 1** — its 3-element example fires where it did not before. Length-dependence is *reduced*, not removed (a drop costs `1/N` rather than `1.5/N`); the length-independent option was rejected at G1 as fork F1. Symptom 2 is untouched — see *Explicitly not fixed* below.

## The change

`RequiredLiteralMissPenalty`: `-0.5f` → `0f`.

A missed required literal was already charged once — it keeps its place in the denominator while contributing nothing to the numerator. Subtracting a penalty on top charged it a *second* time, so one dropped word cost `1.5/N` of a ceiling fixed at `1.0` rather than `1/N`. Because the cost is a fraction of pattern length, it fell hardest on the patterns least able to absorb it: `time to target` heard as "time target" scored `0.50` against a `0.60` gate and did not fire at all, while a 7-element pattern shrugged off the identical drop at `0.79`. Pattern length, not pattern quality, decided whether a command survived a dropped function word.

Reduced, not abolished: the cost stays length-proportional, so `cease fire` heard as "fire" still scores `1/2` and is still refused. `RequiredSlotMissPenalty` is untouched at `-1.0` — an absent argument is not a dropped function word.

| pattern | one literal dropped | before | after |
|---|---|---|---|
| `cease fire` (2) | `1/2` | 0.25 | **0.50** — still rejected |
| `time to target` (3) | `2/3` | 0.50 | **0.67** — fires |
| `decelerate by {burn_level}` (3) | `2/3` | 0.50 | **0.67** — fires, slot intact |
| `launch … on my mark` (7) | `6/7` | 0.79 | **0.86** |
| 5-element, two drops | `3/5` | 0.40 | **0.60** — exactly the gate, fires |

## DR-7 — why this PR is larger than one constant

The review cycle proved **Amendment A1 ruling 2's premise false**. That ruling accepted the zero-crossing on the grounds that *"minScore at the recogniser is the real gate — at default settings a user sees nothing new."* It is not, and they do.

`IsBetterCandidate` ranks **earliest start above score**, so a candidate newly lifted over the `<= 0f` floor does not merely trail as a harmless low-scoring tail result — it can win round 1 outright. Three harms followed, every one reproduced by compiling the real parser at both revisions. The first two bite at the **default** `minScore`; the third additionally requires the `AllowPartialMatch` opt-in, which is off by default:

- **A false positive.** The fragment consumes the leading tokens, moving the origin issue #31 charges skipped words from. On this package's own demo grammar, `"alpha one weapons mode"` went from `mode_weapons 0.500` (correctly suppressed) to `mode_weapons 1.000` — **fired**. Control: `"zzz weapons mode"` unchanged at `0.667`, because out-of-grammar preamble cannot be consumed.
- **A true positive lost.** `_resultBuf` holds one slot per *active* command and extraction breaks silently when full, so a fragment evicts a real command: `"hard burn set burn to coast fire"` lost `fire` entirely, spoken and matched at `1.0`. Verified to be eviction rather than score damage — widening the grammar by one command restored it untouched.
- **Pending widened.** The partial-match branch is gated on `Score > 0f`, not `minScore`, and sits *inside* `if (cmd.Score < minScore)`. 14 of 135 partial demo-grammar utterances went from 0 results to pending-eligible, and `EnterPending` cancels a live pending command.

Per the workflow, implementation contradicting a locked decision means reopening the design rather than patching in place. The design was reopened, **Amendment A2** written, and **re-locked at G1** with ruling: state the rule the penalty had been enforcing by accident.

**DR-7 — a candidate that missed more of its required elements than it matched is not a candidate.** Deliberately a *count* over required elements, not a score threshold: no serialized field, nothing to configure, and independent of the coverage term backlog item 2 will move into the score. Optional elements are excluded from both sides, exactly as they are excluded from the denominator.

Measured against the real parser: all six §5.1 arithmetic cases identical, `F9 Commit` / `F10 None` identical, all 10 ablation rescues intact, F7 fixture delta still zero, and **the A1 ruling 2 residue drops from 4 to 0** — so `NumberSequence_RespectsMaxWords` and `HazardSplitAcrossTwoIntents_Warns` are byte-identical to `main` again and needed no re-derived expectations at all.

## Evidence

- **Unity: EditMode 116/116, PlayMode 369/369** — fully green, both platforms, run against `5e5865e` (artifacts under `Planning~/verification-runs/5e5865e/`).
- **Every numeric expectation hand-derived**, then independently confirmed by a `dotnet` harness compiling the shipped parser at both revisions. No assertion is fitted to output (design §7.3).
- **The `3/5` boundary is bit-exact.** `3f/5f` and `0.6f` are both `0x3F19999A`, and the actual accumulation path lands on those bits — so the `>=` gate admits it on a true equality, not a tolerance.
- **A/B measurement.** The committed corpus cannot usefully exercise this change: **15 of the 16** fixtures are exact phrase matches, so `m = 0`. (The 16th, `split_…`, carries two half-command finals; replayed individually they do move, and `target hotel one` crosses the gate `0.50 → 0.667`.) That would have made the design's planned A/B all but vacuous. Replaced — with the maintainer's agreement — by **single-word ablation of those same real transcripts**: 61 variants, nothing invented. **10 newly clear `minScore`; 9 of the 10 recover the correct intent.** The rig was validated first by reproducing the committed 133-entry grammar pin byte-identically.
- **F7: 0 of 16** manifest cases change `expectedIntent`/`expectedSlots`.

## Review

Eleven finder angles + five adversarial verifiers. Three findings confirmed (the DR-7 harms above), one refuted, one downgraded to a doc gap. Also fixed: a **runtime warning string** that told grammar authors the longer pattern is "penalized for the miss" — nothing is penalized now, it loses the credit — plus eight comment/consistency items.

## Explicitly not fixed

- **Symptom 2 stays open.** `decelerate by {burn_level}` with "by" dropped now scores `0.67`, but bare `decelerate` still scores `1.0`, still wins selection, and the spoken "hard burn" is still discarded. That is backlog item 2 (`feat-coverage-in-selection`).
- **Product docs deliberately untouched** — post-G2 by the project's workflow. Eight sites state the old arithmetic; the maintainer rules the fork at G2.
- **One unruled consequence**, carried to G2: where the dropped word is the *discriminator* between sibling patterns, the survivors tie and registration order picks the winner. Pinned by a test. Note it is **already live on `main` at N≥4** — §5.1 widens it to N≥3.

---

## Review applied (`5e5865e`)

The independent review posted on this PR found **no behaviour defect** — every major reduced to an inaccurate explanation of correct code. Two reductions went against the finder: DR-7's "lost true positives" were `main`'s *non-monotone* behaviour (more debris raised a command from suppressed to fired), and the matched-optional veto needs more optional credit than any shipped pattern carries.

Corrected since: DR-7's rationale (it **tightens**, it does not restore the old filter), the slot-miss routing comment, the optional/denominator analogy, and `TryEagerCommit`'s condition block. Three tests added for gaps the review named — the eager-gate inheritance, the optional clause in both directions, and the partial/pending path at recogniser level. Each was measured at head *and* head-minus-DR-7 first, so each fails with the rule removed.


jinwoo1601 (commented):
# Independent review — PR #72

**Profile:** full (feature lane). CLAIM + all 11 roster angles; **no angles dropped**.
**Head reviewed:** `01511a66fa8a3cc880c4c898132509e473aedd9f`, merge-base `5957fdff89a9a9f9eb6c01e1973a22ea8840cff7`.
**Method:** 12 finder angles, then 3 adversarial verifiers on the surviving majors. Several angles compiled the shipped parser standalone at up to **four** revisions (base, head, head-minus-DR-7, head-plus-variant) and replayed real utterances, so most claims below are measured rather than reasoned.

**Bottom line: no code-behaviour defect survived verification.** All three majors reduced to inaccurate *explanations* of correct code — but two of them are inaccurate inside **locked canon** and shipped source, and one records a consequence the maintainer has not yet ruled on. Nothing here changes what the code does; several things change what the record says it does.

---

## Findings

### 1. DR-7's stated rationale is false — it tightens, it does not restore · CONFIRMED-AS-CLARITY · high

DR-7 is justified in three places as restating what `RequiredLiteralMissPenalty = -0.5f` enforced by accident. It does not. The old filter admitted iff `missed < 2 × matched`; DR-7 admits iff `missed <= matched`. Measured on the base build: 2 matched / 3 missed scores `(2 − 1.5)/5 = 0.1` and **is emitted** on `main`.

Sites: `Runtime/Commands/VoxrCommandParser.cs:927-929`; its verbatim twin at `Tests~/Runtime/VoxrCommandParserTests.cs:1809-1810`; and `scoring-model.md` §A2.3 Option B ("restores as a stated rule what the `-0.5` penalty was providing as an arithmetic artifact").

**The regression this implied was refuted.** A finder reported the band `1/3 < f < 0.5` losing true positives — `"cqb hotel two cease fire"` fires `cease_fire 1.000` on `main` and nothing at head. The verifier reproduced it exactly and then probed the class:

```
"hotel two cease fire"      -> cease_fire 0.500   main: SUPPRESSED
"cqb hotel two cease fire"  -> cease_fire 1.000   main: FIRES
```

On `main`, *more* debris raises the same command from suppressed to firing. That is non-monotone incoherence, not a capability; head is monotone (0.500 → 0.400), which is what the #31 comment says the model is. No rule can separate these cases from the ruled-harmful `"alpha one weapons mode"` — they are the same case, and `main` differed only because an invisible fragment happened to clear `1/3`.

**Still owed:** correct the three sites, and record the `1/3 < f < 0.5` band — §A2.3's Option B table has no row for it, because the prototype measured only the newly-*admitted* direction.

### 2. The slot-miss routing comment is false, and one new band reaches the flush path · CONFIRMED · high

`Runtime/Commands/VoxrCommandParser.cs:54-56` ships: *"`RequiredSlotMissPenalty` stays at -1.0: a missing required SLOT means the command's argument is absent … and is already routed to the pending/partial path."* That routing is `if (cmd.Score < minScore)` (`VoxrCommandRecogniser.cs:668`); above the gate nothing is routed — the command fires with the argument absent. And `AllowPartialMatch` defaults **false**, so at stock settings there is no pending path at all: the sentence is false above *and* below the gate.

It also contradicts this feature's own canon, which already records the claim as "**non sequitur, and empirically false**" (`architecture.md:50`). The doc was corrected; the code comment was not.

**Behavioural half is narrow.** Measured at both revisions, this PR newly opens exactly **one** band: N=8 with one missed required slot **plus** one dropped required literal (`4.5/8 = 0.5625` → `5/8 = 0.625`). N=9 already fires on `main` at `0.6111`; N≥5 with no literal drop is identical at both revisions. The shipped demo pattern `launch {?quantity} {weapon} target {target}` on `"launch all missiles target"` scores `0.6000` with `target` **absent on both revisions** — `main` already fires a targetless `launch_weapon` through the flush path. This is design §5.4's known latent bug widened by one band, and **F10 is satisfied as literally worded** ("must still refuse the *eager* commit").

**Still owed:** fix the comment; name the N=8 band at G2 beside the dropped-discriminator consequence.

### 3. The optional-exclusion analogy is false, in shipped source and in locked canon · CONFIRMED-AS-CLARITY · high

`VoxrCommandParser.cs:889-891` and `scoring-model.md` §A2.5 ruling 1 both justify DR-7's optional handling as "excluded from both, **exactly as they are excluded from the denominator**". A *matched* optional is not excluded from the denominator — `+0.5/+0.5` for a literal, `+1.0/+1.0` for a slot (`:1027-1030`, `:1053-1054`), and canon's own §2 table says so. Only *omitted* optionals leave both sides. The rule as implemented is correct; its reason is not.

**The alleged harm was refuted.** A DR-7 veto can only reach a gate-passing candidate when matched-optional credit `P ≥ 2.0`; **no shipped pattern exceeds `P = 1.5`**. Across the 16 committed transcripts and every single-word ablation, **zero** utterances lost a result. Patching matched optionals into `matchedRequired` produces **byte-identical corpus output** — the alternative buys nothing.

---

## Minor / nit findings

**PR body over-claims (4).** "Every fixture is an exact phrase match" — canon says **15 of 16**; the `split_…` half `target hotel one` crosses the gate `0.50 → 0.667`. · "PlayMode 366/366" has **no recorded provenance** — the build record stops at `363/363` on `3367808`, and the head commit records a *red* `365/366`. · "all at the **default** `minScore`" over-reaches for harm 3: `AllowPartialMatch` and `FireAsIs` both default off. · "Closes symptom 1" — #65 titles it "length determines fragility", which this *reduces*, not removes.

**Product docs — the fork list is incomplete in kind, not just in count.** Every enumerated site is arithmetic; **no entry records a DR-7 doc need at all.** `Documentation~/scoring.md:118` enumerates extraction's stopping conditions and DR-7 adds a fourth; `scoring.md:92` ("every candidate with a positive score competes") is now simply false; `command-recognition.md:107,114` is the prose twin of the warning string this PR corrected, and sits outside the listed Scored-Matching range.

**Coverage (4).** DR-7 is a live refusal reason inside `TryEagerCommit` with **no test** and no mention in that method's otherwise-exhaustive refusal-condition comment block. · The optional-exclusion clause has zero coverage — all three `Admission_*` fixtures are optional-free. · The pending harm is asserted at parser level only, never where it lives. · No F-row exists for DR-7, and `architecture.md` still reads `status: halted-pending-g1`.

**Conventions / accuracy (7).** "Halved" survives in two test comments though `1.5/N → 1/N` is a third (four angles flagged it). · A region preamble says "these three" over four tests. · `MatchedRequired` breaks the struct's `Count` suffix. · An F6 derivation divides by 4 without saying why. · The discriminator test re-implements `ModeSwitchParser()` element-for-element. · `requirements.md` F7 still says 17 fixtures where the manifest holds 16. · A parser-level assertion message says "firing on the boundary" where `Parse` applies no threshold.

**Non-functional.** `requirements.md:52` says "**Zero cost** … any measurable parse-time regression is a defect". Measured: `MatchResult` 32 → 40 bytes, +2–9 % parse time (≈0.15–0.8 µs on a 7.7 µs parse), and DR-7 does **not** cut rounds — base and head are byte-identical at 21 rounds. Allocation is unchanged to the tenth of a byte (344.4 B/utterance). That "Must" needs amending or the G2 reader will score it as passing.

---

## Considered and killed

- *DR-7 destroys true positives in the `1/3 < f < 0.5` band* — **refuted**: `main`'s behaviour there is non-monotone; head is the coherent one.
- *DR-7 wrongfully vetoes gate-passing candidates via matched optionals* — **refuted**: needs `P ≥ 2.0`, no shipped pattern exceeds `1.5`, zero corpus impact.
- *`return default` can now slip through admission* — **refuted**: `Score <= 0f` fires one line earlier.
- *Test-count arithmetic doesn't reconcile* — **refuted**: 337 → 351 static, constant +15 offset from other assemblies, so 366 is correct.

## Verified sound

Every numeric expectation independently re-derived **and** confirmed against the real parser at both revisions — no assertion is fitted to output (F6 holds). · All four DR-7 increments sit in the correct branches; optionals touch neither counter. · The `>` comparison must be `>` — `>=` breaks two in-tree cases. · `3f/5f` ≡ `0.6f` ≡ `0x3F19999A` through the *accumulated* path; and since `rawScore`/`denominator` only accumulate exact multiples of 0.5, equal scores are always bit-equal, so the untouched `Score != bestScore` equality is safe despite §5.1 making ties commoner. · The three `Admission_*` tests are **not vacuous** — all three fail with the rule deleted. · F9 clean: 40 000 random grammar/utterance pairs, `Commit ⟹ Parse` yields one result at or above `minScore`, **0 violations** at head and base. · The zero-crossing is **provably** closed — `main raw ≤ 0 ∧ head raw > 0 ∧ L+S ≤ m` is unsatisfiable, so M1/M2/M3 cannot recur through any grammar. · 180 demo-grammar ablations: identical result counts, **0 intent flips**. · `ScoreFollowUp` genuinely untouched; all array-shape consumers guarded; `GenerateGrammarJson` byte-identical so `expectations.json` does not re-baseline. · No `[SerializeField]`, no public API change, no TODOs introduced.

- **The Unity run itself.** Bare UPM package, no `Assets/`; `116/116` and `366/366` were produced in a host project unreachable from here. The counts were checked for *self-consistency* against the diff and reconcile — but see minor finding above: no build record names a run on `01511a6`.
- **On-device Quest behaviour** — not required for this feature (design §7.5); nothing here touches capture, the bridge, or lifecycle.
- **IL2CPP/arm64 codegen.** All timing and bit-pattern work was desktop x64 CoreCLR.

*Posted by an independent review fan-out; no angle inherited context from the session that wrote the code.*

jinwoo1601:
Review fixes pushed as `5e5865e`; the PR body above is corrected for the four claim findings.

**Unity re-verified on `5e5865e`: EditMode 116/116, PlayMode 369/369, fully green** (was 366/366 — the three new tests account for the difference).

Applied from the review:

- **Four false rationales corrected in shipped source.** DR-7 "restores what the penalty enforced" → it **tightens** (old filter: `missed < 2 × matched`; DR-7: `missed ≤ matched`), with the monotonicity evidence that justifies the tightening. Plus the slot-miss routing comment, the optional/denominator analogy, and `TryEagerCommit`'s condition block, which never mentioned that admission runs above all of it.
- **Two dated correction notes inside Amendment A2** (§A2.3, §A2.5 ruling 1) — original text preserved, rulings untouched.
- **Three tests** for the gaps named: eager-gate inheritance of DR-7, the optional clause in *both* directions, and the partial/pending path at recogniser level with a positive control. Each was measured at head and head-minus-DR-7 before being written, so each fails with the rule removed.
- **`requirements.md`**: the "Zero cost" NFR amended against measurement (+2–9 %, `MatchResult` 32→40 bytes, allocation unchanged at 344.4 B/utterance); **F13** added for DR-7; F7's fixture count fixed; and **F4 flagged as falsified** with a new §4.1 — its acceptance check discharges vacuously, since it cannot see candidates that were previously invisible.

**Carried to G2 unruled**, not fixed here:

1. **Product docs — eleven sites, four files, two classes.** The review found the fork list was incomplete *in kind*: every previously-listed site was arithmetic, and none recorded a DR-7 need. `scoring.md:118` enumerates extraction's stopping conditions and DR-7 adds a fourth; `scoring.md:92` ("every candidate with a positive score competes") is now flatly false.
2. **Dropped discriminator** — where the dropped word is the discriminator between siblings, registration order picks the winner. Already live on `main` at N≥4; §5.1 widens it to N≥3. Ruled on by neither A1 nor A2.
3. **N=8 slot-miss band** — a command firing with a required argument absent. Widens a latent bug design §5.4 already documents (`main` already does this at five elements); F10 is satisfied as literally worded, since it says *eager* commit.
4. **Symptom 2 stays open** — item 2's scope.

jinwoo1601:
**G2 accepted** (human ruling in conversation). Product docs pushed as `e5eff7b` — this PR is ready for your merge.

### Product docs — corrected now, scoped to correctness

The review showed the doc problem was in **two classes**, and the second is what made deferral the wrong call:

- **Arithmetic** — worked examples, the fragility table, the `KNOWN_LIMITATIONS` root cause, the troubleshooting heading (which described a case that can no longer occur).
- **The model itself** — `scoring.md` stated *"Every candidate with a positive score competes"* and gave three stopping conditions for sequential extraction. Both are now false. A reader debugging a non-matching command was being told the wrong **rule**, not just a stale number.

`scoring.md` gains an **Admission** subsection stating both filters; §3 and §4 point at it; `troubleshooting.md` gains *"A command reports nothing at all, though it clearly part-matched"*, which is the admission rule's user-visible face. `CHANGELOG.md` carries two behaviour-change entries under `[Unreleased]`, per DR-5. The structural rewrite of `scoring.md` stays backlog item 3's.

### Filed rather than fixed here

- **#73** — the flush path has no completeness condition, so a command can fire with a required slot unfilled. Pre-existing at N≥5 on `main` (the shipped demo grammar does it today at exactly `0.60`); this PR widens it by one band. Issue lane, on the #66/#70 precedent. The rule there is statable, which is why it is a fix rather than a doc entry.
- **#74** — design-lane: route an ambiguous sibling tie to confirmation instead of firing the first-registered pattern. Not a bug — no scorer can recover an intent whose discriminating word is gone — but guessing may not be the best available answer, and the parser already knows when it is guessing.
- **Dropped discriminator** documented in `KNOWN_LIMITATIONS.md` with the authoring shapes to prefer.

### Still open, by design

**#65 stays open.** This closes the cases it reports under symptom 1; symptom 2 (a bare sibling still beating its slot-filled form at 1.00) is backlog item 2's scope and is untouched.

---

Final state: **EditMode 116/116, PlayMode 369/369**; F7 zero fixtures changed; 10 of 61 ablated real transcripts newly recognised, 9 recovering the correct intent. Preflight PASS on all 8 items. **The merge is yours** — I do not merge or tag.

PR #75, opened 2026-08-13
Refuse a command missing a required argument on the flush path (#73)

### LINKED ISSUE #73: Flush path has no completeness condition: a command can fire with a required slot unfilled
Split out of #65 at G2 on the #66/#70 precedent. Pre-existing on `main`; PR #72 widens it by one band, which is the same trigger that sent #70 to the issue lane.

## The gap

The eager-flush gate has two completeness conditions — no missed required **slot** (#66) and no required element after the last matched one (#70). **The ordinary flush path has neither.** `MissedRequiredSlot` and `HasUnmatchedRequiredTail` are consulted only inside `TryEagerCommit`; `ParseInternal` never sees them.

So once a slot-missing candidate clears `minScore`, nothing stops it: it fires with the argument absent. And this is the *default* path — `eagerFlushOnCompleteMatch` is off by default.

## Repro (both on `main` today)

The shipped demo grammar, `launch {?quantity} {weapon} target {target}`:

```
"launch all missiles target"  ->  launch_weapon  score 0.6000  target=<ABSENT>  -- fires
```

`(1+1+1+1-1)/5` = exactly `0.60`, which clears the default gate. The handler receives a `launch_weapon` with no target.

Design §5.4 already documents this as "a live latent bug on `main`" for N≥5. PR #72 lifts one further band over the gate — N=8 with one missed slot **plus** one dropped required literal, `4.5/8 = 0.5625` → `5/8 = 0.625`. N=9 already fired on `main` at `0.6111`.

## Why it is worth fixing rather than documenting

Unlike the dropped-discriminator case (see the new KNOWN_LIMITATIONS entry), **the rule here is statable**: a command missing a required argument should not fire silently. The machinery already exists — `MatchResult.MissedRequiredSlot` is computed on every candidate, and the partial/pending path is the natural destination for these.

Note `allowPartialMatch` defaults **off**, so today a below-gate slot-missing candidate is simply dropped and an above-gate one fires. Neither is "routed to the pending path", despite what the parser's own comment used to claim (corrected in PR #72).

## Suggested shape

Refuse, or route to slot-fill, any winner with `MissedRequiredSlot` — on the flush path as well as the eager one. Worth checking whether that wants to be `minScore`-independent (a missing argument is a missing argument at any score).

## Not in scope

The dropped-discriminator tie (documented in KNOWN_LIMITATIONS instead — no scorer can recover an intent whose discriminating word is gone) and backlog item 2's coverage term.

Found by the independent review of PR #72.

jinwoo1601:
Closes #73

## What

The ordinary flush path now refuses to fire a command whose matched pattern left a **required slot** unfilled, and routes it to slot-fill where the command opted into `allowPartialMatch`.

- `VoxrCommandParser.HasUnfilledRequiredSlot(cmd, def)` — one shared predicate, answered from the finished `VoxrCommand` rather than `MatchResult`.
- `VoxrCommandRecogniser` — Step 7's gate becomes `Score < minScore || incomplete`; Step 4's `hasCompleteNewCommand` gains the same term.
- `VoxrBatchTestRunner.PassesThresholds` — same rule, on the #31 precedent that batch results keep predicting runtime behaviour.
- Four new PlayMode tests; `CHANGELOG.md` entry.

## Why

`TryEagerCommit` has refused this shape since #66, but that guard covers only the opt-in eager path. The default path asserted score and confidence and nothing else, leaving completeness to the missed-slot penalty (`−1.0`) to enforce by arithmetic. It does not enforce it: on the shipped demo grammar `"launch all missiles target"` scores `(1+1+1+1−1)/5` = **exactly `0.60`**, the default `minScore` itself, and delivered `launch_weapon` with no `target`.

**The fix is deliberately not where the issue suggested.** The issue proposes refusing the winner in `ParseInternal`. That would delete the only input the `allowPartialMatch` path has — that path is fed precisely by slot-missing candidates scoring *below* `minScore` — and would break `TryEagerCommit_WidenedSlotMissHole_StaysClosed`, which PR #72 wrote to pin that candidate as still-reported. So the condition sits in the recogniser, beside the gate it belongs to. `ParseInternal` is untouched.

Stated **independent of `minScore`**, as the issue asked me to check: a missing argument is a missing argument at any score.

## Scope — slots only

A missed required *literal* still fires; the command is fully determined and its arguments present, which is what #65 §5.1 exists to allow. That includes the dropped-discriminator case (`HasUnmatchedRequiredTail`), which the issue's opening line might read as in scope. It is not: the design ruled that case out of the issue lane on three disanalogies and shipped it as a `KNOWN_LIMITATIONS` entry instead. At the eager gate refusing costs only latency; on a final transcript it means firing nothing.

## Two extensions beyond the issue text

Both deliberate, both worth ruling on:

1. **Step 4's `hasCompleteNewCommand`.** Fixing only Step 7 would leave an incomplete command still cancelling a live pending command — taking a half-finished command away and putting nothing in its place. That would be a *new* bug introduced by a partial fix.
2. **`VoxrBatchTestRunner`.** Without it the harness reports PASS for an utterance the runtime now refuses, certifying a grammar against behaviour the user will never see.

## Validation

- [x] Unity **EditMode 116/116** and **PlayMode 373/373** green on this branch (Unity 6000.4.7f1, host project `VoXR TestGround`, both platforms run separately). No compile errors.
- [x] All four new tests ran in PlayMode and passed: `SlotMissing_AboveGate_DoesNotFire`, `SlotMissing_OptionalSlotOmitted_StillFires`, `PartialMatch_AboveGateButIncomplete_EntersPendingInsteadOfFiring`, `IncompleteNewCommand_DoesNotCancelALivePending`.
- [x] `TryEagerCommit_WidenedSlotMissHole_StaysClosed` still passes — `Parse()` still returns the slot-missing candidate at 5/8, so the parser-level reporting contract is intact.
- [x] Corrected three stale code comments and one `CHANGELOG` clause that asserted slot-missing candidates "reach the pending-command path" — they did not.
- [ ] On-device (Quest) verification **not run** (WSL cannot). Not obviously implicated: this is parser/recogniser logic covered by the PlayMode suite, not mic capture or `OnApplicationPause` delivery.


jinwoo1601 (commented):
# Independent review — PR #75

**Head reviewed:** `fe33f334e85ab28e312624f6a7f03315bb31a313` · **merge-base:** `126a0483cc59d702c577288531d7c0b86e1250d8`
**Method:** six finder agents with no context from the session that wrote the code, deduped, then seven adversarial verifiers (one per surviving substantive finding). 16 raw findings → 12 deduped → **5 confirmed by verification, 4 refuted**, plus 9 textual/coverage items confirmed by direct inspection.

**No blocker.** Per the issue lane nothing has been fixed on the branch; everything below is unruled and yours to decide.

The core gate came through clean: `LINE` re-derived the predicate, the struct guard, and the `!IsOptionalSlot` term (genuinely load-bearing — `ExtractSlotName("{?quantity}")` returns `"quantity"`); `PIT` returned no findings at all and confirmed the gate sits **outside** `#if UNITY_EDITOR`, so it closes identically in a player build. `XFILE` proved the two slot computations (`HasUnfilledRequiredSlot` vs `ComputeUnfilledSlots`) cannot disagree, and that `IndexByIntent` matches `CommandSetManager.BuildLookup` on comparer, duplicate-intent resolution, and absent-intent behaviour.

---

## Confirmed — recommended before merge

### 1. The three discriminating tests all sit on exactly `0.60`, and go vacuous under scoring drift — **major**
`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:898` · `Tests~/Runtime/VoxrPendingCommandTests.cs:115`, `:206`

All three tests that discriminate the fix use a candidate landing on the gate value itself. They **do fail under a plain revert today**, so they are not worthless. But they pin *"does not fire at 0.60"*, never *"does not fire at any score"* — which is this PR's own headline claim ("Stated **independent of `minScore`**") and invariant 1.

Verifier re-derived every path: if a future change moves these candidates below `0.60` **and the completeness code is deleted**, all three still pass. `SlotMissing_AboveGate_DoesNotFire` still reports unrecognised; `PartialMatch_AboveGateButIncomplete_EntersPendingInsteadOfFiring` still enters pending via the pre-existing below-gate route; `IncompleteNewCommand_DoesNotCancelALivePending` still leaves `hasCompleteNewCommand` false. `IsIncomplete` could be deleted wholesale with a green suite.

The drift is not hypothetical. **PR #72, one commit back, changed a scoring constant and by its own CHANGELOG "lifted a further band … over the gate".** And this PR's own new comment de-anchors the constant holding these at `0.60` — `VoxrCommandParser.cs:67-69`: *"-1.0 stays because a missing argument IS weaker evidence and should score lower, **not because the score is what stops the command firing**."*

The repo already has the right idiom, 200 lines from the new tests: `Tests~/Runtime/VoxrEagerCommitTests.cs:1141` asserts *"must never commit early, however it scores"* on an 8-element fixture at `5/8 = 0.625` — **materially above** the gate. The new flush-path test says *"whatever it scores"* and then pins `0.60`.

*Correction to the finding as first raised:* it said "every new test". `SlotMissing_OptionalSlotOmitted_StillFires` scores `4/4 = 1.0` and is a legitimate over-correction guard, not a fix-pin.

**Fix:** one flush-path test with an incomplete candidate materially above the gate (the `0.625` shape already exists in the eager fixture).

### 2. `VoxrBatchTestRunner`'s completeness branch has zero test coverage — **moderate**
`Runtime/Testing/VoxrBatchTestRunner.cs:199-206`

`grep -rn "HasUnfilledRequiredSlot\|required slot unfilled" --include=*.cs` returns **zero hits outside `Runtime/`**. None of the 24 tests in `Tests~/Editor/VoxrBatchTestRunnerTests.cs` reaches the branch — verifier enumerated all of them:

- `Run_BelowMinScore_RejectedCorrectly` — `"cease xyz"` at `0.50`, returns at the score branch above.
- The three `"launch all missiles target hotel one"` tests — `5/5 = 1.00`, all slots filled. These evaluate the condition to `false`, so they would catch an **over**-refusing predicate but not the under-refusal it exists to prevent.
- All `"cease fire"` tests — no slot elements at all.
- `"hello world"` / `null` / `""` — `parseResults.Length == 0`, `PassesThresholds` never called.

This is a behaviour change to a **public shipped type** (`Documentation~/api/batch-test-runner.md`), and the PR body justifies this half's existence precisely by a divergence risk that only a test can catch. The test is constructible from the file's existing fixture — `MakeCommands()` already carries `launch {?quantity} {weapon} target {target}`, so `input = "launch all missiles target"` with `expectedIntent = null` discriminates the diff.

---

## Confirmed — behaviour, narrow

### 3. Completeness is evaluated ahead of the confidence gate — **minor**
`Runtime/Commands/VoxrCommandRecogniser.cs:694-695` vs `:734`

Found independently by two angles. The new branch `continue`s before the confidence check, so an above-gate, below-confidence, incomplete result no longer reaches `minConfidence`.

Verification cut this down substantially, and the residue is narrow:
- **Pre-existing:** the `AllowPartialMatch` route has *never* consulted `minConfidence`; at base a below-gate candidate at confidence `0.15` already entered pending and evicted a live pending.
- **Genuinely new:** only the band `Score >= minScore ∧ incomplete ∧ 0 <= Confidence < minConfidence`.
- **Refuted sub-claims:** it does *not* defeat `IncompleteNewCommand_DoesNotCancelALivePending` "via a different door" (that test pins the case where the command *cannot* replace the pending); and the "now reports `OnUnrecognisedSpeech` where previously silent" half is the coherent tier choice, not a defect.
- Requires `allowPartialMatch: true` (default off) **and** real word-confidence data — the `-1` sentinel makes it unreachable from `InjectText` tests.

Fair residual point: the comment explains the routing but never says why completeness is evaluated *ahead* of confidence.

### 4. `RequiresConfirmation` + `AllowPartialMatch`: the confirm window is the remainder of one shared `pendingTimeout` — **nit**
`Runtime/Commands/PendingCommandHandler.cs:196` (untouched) · routing changed at `VoxrCommandRecogniser.cs:695`

An above-gate incomplete command with both flags now reaches `AwaitingConfirmation` via `PartialMatch`, carrying the original `CreatedTime` rather than a fresh stamp. `grep CreatedTime` over `Runtime/` finds exactly two writes and no restamp anywhere.

Mechanism is 100% pre-existing (`PendingCommandHandler.cs` has **zero lines in this diff**) and already traversed by the below-gate population, pinned by `Combined_PartialThenFollowUpThenConfirm`. And base's "full window" was not a correct baseline: for this population base entered `AwaitingConfirmation` on an incomplete command and fired it with the slot absent on "yes". Untested either way — both fixtures set `PendingTimeout = 30f` to take timing out of play.

---

## Confirmed — accuracy of claims, comments and docs

### 5. Validation box 4 overstates what was corrected — **minor**
The box says *"Corrected three stale code comments … that asserted slot-missing candidates 'reach the pending-command path' — they did not."* Only **one** comment asserted that (base `:873-874`), plus one inheriting it by reference. The third — the `RequiredLiteralMissPenalty` block — asserted the **opposite** and was already accurate; it was edited because the fix made its present tense stale, not because it was false. Issue #73's own text says as much.

### 6. `OnUnrecognisedSpeech` cause-enumerations are now incomplete — **minor**
`Documentation~/command-recognition.md:426-431` and `Documentation~/api/command-recogniser.md:42` each list three causes; neither includes "above `minScore` but missing a required slot", which head now fires on. Separately, `CHANGELOG.md:35`'s "the utterance is reported through `OnUnrecognisedSpeech` instead of firing incomplete" mis-describes the debounce/confidence subcase, where base fired **nothing** rather than firing incomplete.

Also stale-by-attribution (mechanism wording only, main clauses still true): `command-recognition.md:32-34`, `:364`, `:430`. And inside the file the brief excluded from scope: `scoring.md:143` ("two independent thresholds") and the `:184-193` "In order:" table, which now omits the completeness refusal.

### 7. Two comment claims are false as written — **nit**
- `VoxrCommandRecogniser.cs:817-818`: *"the lookup only fails if the active set changed between the parse and this loop."* `ProcessParsedResultsCore` parses at `:598` and loops at `:663` **synchronously** — this cannot happen. The guard may still be worth keeping; the stated reason is wrong.
- `VoxrCommandParser.cs:1338-1339`: *"The two agree by construction."* `MissedRequiredSlot` is set per **element**; `HasUnfilledRequiredSlot` asks per **name**. For `swap {target} and {target}` with the first filled and the second not, they disagree — eager refuses, flush fires. Pathological grammar, no behavioural fix needed, but the equivalence is over-claimed.

### 8. Smaller accuracy items — **nit**
- `VoxrCommandRecogniser.cs:611`: the comment attributes the pending cancel to "Step 6". The cancel is at `:650`; the `// ---- Step 6` header is at `:652`. It is the tail of Step 5.
- `VoxrCommandRecogniser.cs:812`: "the counterpart to the two TryEagerCommit already enforces" — dropped word ("the two **conditions**").
- `CHANGELOG.md:35`: "an incomplete command no longer cancels one" is true only for the non-`allowPartialMatch` subclass — with the flag on, `EnterPending` still cancels the prior pending. Worth qualifying as "no longer cancels one *without replacing it*".

### 9. Untested paths — **minor**
- The `"required slot unfilled"` reject string is asserted nowhere, though `Tests~/Editor/VoxrCommandRecogniserDiagnosticTests.cs` asserts the reason text for all four other rejection classes.
- Incomplete + `RequiresConfirmation` above the gate. (`Combined_PartialThenFollowUpThenConfirm` uses both flags but at `2/4 = 0.5`, below the gate — the pre-existing path.)
- A mixed utterance (one complete extraction + one incomplete) — Step 7 loops over `resultCount`, and this resolves invariant 6 differently from `SlotMissing_AboveGate_DoesNotFire`, as an emergent side effect rather than a decision.

---

- **"Routing to pending does not guarantee the argument arrives; the follow-up exit fires incomplete"** (raised major). `PendingCommandHandler.cs` has **zero lines in the diff**. The finding's scenario scores `0.333` — below the gate — so it demonstrates the pre-existing path. For the newly-routed population the hole needs ≥2 unfilled required slots (a ≥10-element pattern at the default gate), and **on exactly those inputs base fired the command outright with both slots absent**; post-PR the worst case fires with one filled. Exposure strictly reduced. *The underlying mechanism is real and worth its own issue* — three exits apply no completeness test (`TryFollowUpSlotFill`, `TryHandleConfirmCancel` on a `PartialMatch` pending, `HandleTimeout` FireAsIs).
- **"An incomplete new command can no longer preempt a follow-up slot-fill"** (raised minor). Arithmetically impossible at base: a command cannot both set `hasCompleteNewCommand` (needs `Score >= minScore`) and reach `EnterPending` (needs `Score < minScore`). The worked example scores `0.50`, so base and head behave identically on it. A real divergence exists in the reclassified band, but head is defensible and arguably better — base cancelled the pending *and* fired an incomplete command. Residue: the rationale is not exhaustive about the `allowPartialMatch` sub-case.
- **"Pending + `OnUnrecognisedSpeech` on the same utterance"** (raised minor). Pre-existing and explicitly documented — `scoring.md:203`: "**fires**, alongside `OnCommandPending`". Base did the same for below-gate partials.
- **"`scoring.md:149,193` contradicts the code"** (raised minor). Those sentences say sub-threshold matches *do* enter pending — still true. Neither contains "only", so "per those two sentences this cannot happen" is an inference the text does not carry. The third citation (`command-recognition.md:364`) is not the same phrasing and is made *more* true by the diff.

Also checked clean and reported by no angle: assembly/`internal` resolution, no `[SerializeField]` added or renamed, no new float equality, no duplicate-key or null-key risk in `IndexByIntent`, `_defsByIntent` assigned on both constructor paths, no new tuning surface (invariant 7), the `(1+1+1+1−1)/5 = 0.60` arithmetic (independently re-derived twice), and the literal claim "`ParseInternal` is untouched" (byte-identical, verified).

---

- **The Unity suite was not re-run.** The host project is required and a run takes minutes. The ticked "EditMode 116/116 / PlayMode 373/373" box is a claim this review could not independently confirm. It is *not* unbacked: `CLAIM` located the run artifacts, found all four new tests plus `TryEagerCommit_WidenedSlotMissHole_StaysClosed` recorded `Passed`, and cross-checked the counts against PR #72's stored run (369 + 4 new = 373). What could not be confirmed is that the tree hashed to `fe33f33` when they were produced. Unlike PR #72, no artifact was filed under `Planning~/verification-runs/<sha>/`.
- **On-device (Quest) verification.** Correctly marked not run; not implicated by scorer/recogniser logic.

---

*This review was produced by agents that inherited nothing from the session that wrote the code, and it reviews that session's own work. Findings were deduped for overlap, not for comfort.*

PR #78, opened 2026-08-14
Charge candidates for what they leave unexplained, inside selection (#65 §5.2, DR-4)

jinwoo1601:
Closes symptom 2 of #65, and #42.

A grammar registering both `decelerate` and `decelerate by {burn_level}`, hearing *"decelerate hard burn"* with the "by" dropped, fired the **bare** command and discarded the burn level with nothing in the log to signal it. No threshold reached this: the bare pattern matched perfectly at `1.0`, and nothing normalised to `1.0` can be beaten.

Coverage now charges a candidate for the in-grammar tokens it leaves unexplained **on either side**, computed **before** candidates are compared rather than applied to the winner afterwards. That relocation is the feature — the leading skipped-word term (#31) sat after selection precisely so it could not reorder, and symptom 2 cannot be fixed without reordering. Bare `decelerate` now scores `1/(1+2) = 0.333` against the slot-filled `2/3 = 0.667`, so the command fires **with** its argument.

```
score = rawScore / (denominator + (skippedBefore + orphanedAfter) × coverageWeight)
```

A trailing token is orphaned if it follows the candidate's `ConsumedEndIdx` and no active pattern could begin a match at it; counting stops at the first token that could. That test is what keeps sequential extraction intact — *"cease fire launch missiles target hotel one"* still scores `2/2`, not `2/7`. `[unk]` is never charged and is transparent rather than a run terminator, so one noise token cannot shield every real orphan behind it.

## Reading this diff

It is two separable concerns that interleave inside `VoxrCommandParser.cs`, which is why they share a commit:

- **The behaviour change** — the pattern-start caches (built in the constructor), the three per-utterance coverage tables, the coverage term in `TryMatchScored`, and the deletion of the post-selection block in `ParseInternal`.
- **The DR-4 rename** — `skippedWordPenalty` → `coverageWeight` across eleven sites, `[FormerlySerializedAs]` on the serialized field. `VoxrCommandRecogniser.cs` and `VoxrBatchTestRunner.cs` are *entirely* rename.

## A design amendment the build forced

Phase 7's A/B found the locked §5.2 rule **fires the wrong command**, and it needed a G1 re-lock (Amendment A3) before this could ship.

The orphan predicate asks *"could a pattern **begin** a match here?"*, but the matcher can begin a pattern at any token by missing its leading elements. So `["switch","to","navigation"]` heard against *"switch to weapons target hotel"* **misses** its final element, stops short of "weapons" — which begins `["weapons","mode"]` and so terminates its orphan run at zero for a tidy `2/3` — and beats the correct `["switch","to","weapons"]`, which **matched** that element, moved its origin past the very token that would have terminated its own run, and pays `3/(3+2) = 0.600`.

**A candidate was rewarded for matching less.** A3 charges the mis-predicted token instead of testing it. Measured threshold: safe at one leftover token, flips at two.

The suite was green throughout. Only a purpose-built corpus caught it.

## Measurement

699 utterances in five generation families, judged at the `minScore` gate. The committed 16 transcripts cannot exercise this feature at all — every one begins at the first recognised token and consumes to the end — so insertion and concatenation were added to the inherited ablation.

| | count |
|---|---|
| newly-firing | 0 |
| slots-gained | 0 |
| intent-change | 11 |
| count-change | 1 |
| slots-lost | 0 |
| newly-silent | 17 |
| score-only | 116 |

All 16 baselines unchanged; the grammar pin still emits 133 entries identical to `expectations.json`, so the native harness does not re-baseline. **No wrong command fires anywhere in the corpus.**

| | `main` | after |
|---|---|---|
| bytes/utterance | **343.0** | **343.0** |
| µs/utterance | 12.22 – 12.65 | 12.34 – 12.81 |

## Known limitations, measured and accepted

Bound for `KNOWN_LIMITATIONS.md` after acceptance, not silently absorbed:

- Where a **second** command in one utterance loses its own leading word, the first is still charged for the second's tokens and can fall below `minScore`. Loses a command; never fires a wrong one.
- Only the literal `[unk]` token is exempt, so `freeSpeechMode`, `InjectText` and `VoxrBatchTestRunner` — which deliver real tokens — charge trailing filler: *"cease fire please"* drops `1.0 → 0.667`.
- A bare pattern demoted to `0.333` fires nothing when **no sibling exists**.
- A slot-initial pattern over a permissive slot can zero trailing coverage grammar-wide.

## Breaking change

`VoxrBatchTestRunner`'s two public constructors renamed a parameter, so **named-argument** callers must update. Positional callers are unaffected. Inspector values survive via `[FormerlySerializedAs]` — but note that attribute governs a field's own deserialization and does **not** reach prefab-instance overrides; nothing shipped here is affected, and one prefab-override round-trip is on the human-verification list.

## Verification

EditMode **116/116**, PlayMode **399/399**. No on-device check required (design §7.5). A prior `review-cycle` (11 angles, 5 adversarial verifiers, 21 findings) is applied in `60a064e`.

### DIFF
--- a/Editor/VoxrDebugSessionLog.cs
+++ b/Editor/VoxrDebugSessionLog.cs
@@ -49,7 +49,8 @@ namespace VoXR.Editor
             + "numbers embedded in rejectReason text are formatted with the Editor's current "
             + "culture, so the decimal separator may be ',' — match on the surrounding words, "
             + "not the whole literal. The scoring model behind these numbers — the score "
-            + "formula, the skipped-word penalty, selection order, the two gates, and worked "
+            + "formula, the coverage weight charged for in-grammar words a match leaves "
+            + "unexplained before AND after it, selection order, the two gates, and worked "
             + "examples — is documented in Documentation~/scoring.md in the com.jinwoo1601.voxr "
             + "package.";
 
--- a/Runtime/Commands/VoxrCommandParser.cs
+++ b/Runtime/Commands/VoxrCommandParser.cs
@@ -149,8 +149,8 @@ namespace VoXR.Commands
 
         // Per-utterance coverage tables, rebuilt by BuildCoverageTables at the top of every
         // parse and reused across that parse's extraction rounds. Grown on demand and never
-        // shrunk, matching the _matchSlotBuf / _resultBuf pooling discipline — coverage adds
-        // no per-utterance allocation.
+        // shrunk. Like _matchSlotBuf / _resultBuf they are allocated off the parse path, so
+        // coverage adds no per-utterance allocation (those two are sized once instead).
         //
         //   _recognisedPrefix[i]  — non-[unk] tokens in [0, i), so the words a round's sliding
         //                           start walked past are one subtraction instead of a scan.
@@ -208,8 +208,18 @@ namespace VoXR.Commands
             public int WordCount;
         }
 
+        // additionalGrammarWords are words the caller also put in the DECODER's grammar but
+        // that appear in no pattern — in practice the confirm/cancel follow-up vocabulary
+        // (VoxrCommandRecogniser.GetFollowUpGrammarWords). They matter to coverage because the
+        // decoder returns them as real tokens rather than [unk], so without them the trailing
+        // term charges a speaker for saying "disengage, yes": nothing can begin a match at
+        // "yes", so it reads as an orphan and sinks a command that used to fire. They can
+        // legitimately begin something — a follow-up — so they terminate an orphan run exactly
+        // as a pattern start does. Null is correct for a caller that registered no follow-up
+        // vocabulary with the decoder either.
         public VoxrCommandParser(VoxrSlotDefinition[] slots, VoxrCommandDefinition[] commands,
-            float coverageWeight = DefaultCoverageWeight
+            float coverageWeight = DefaultCoverageWeight,
+            string[] additionalGrammarWords = null
         )
         {
             if (slots == null) throw new ArgumentNullException(nameof(slots));
@@ -217,7 +227,11 @@ namespace VoXR.Commands
 
             _slots = slots;
             _commands = commands;
-            _coverageWeight = coverageWeight > 0f ? coverageWeight : 0f;
+            // Rejects negatives and NaN, and also non-finite positives: at +infinity a
+            // candidate with nothing unexplained computes 0 * infinity = NaN, and NaN slips
+            // through every `<= 0f` floor because those comparisons are false for NaN.
+            _coverageWeight =
+                coverageWeight > 0f && !float.IsInfinity(coverageWeight) ? coverageWeight : 0f;
 
             // Build slot name -> index mapping
             _slotIndex = new Dictionary<string, int>(slots.Length, StringComparer.Ordinal);
@@ -334,6 +348,19 @@ namespace VoXR.Commands
                     }
                 }
             }
+            // Words the decoder can return that begin no pattern but do begin a follow-up.
+            // Folded in here rather than kept separate because the predicate's question is
+            // "could anything the grammar knows about start here?", and the answer for a
+            // confirm/cancel word is yes.
+            if (additionalGrammarWords != null)
+            {
+                foreach (string word in additionalGrammarWords)
+                {
+                    if (!string.IsNullOrEmpty(word))
+                        _startLiterals.Add(word);
+                }
+            }
+
             _startSlots = startSlots.ToArray();
 
             // Compute max slots per pattern to size reusable buffers.
@@ -1258,7 +1285,7 @@ namespace VoXR.Commands
             // requiredAfterLastMatch selects WHICH orphan table to read (Amendment A3): a
             // candidate whose own next required element failed at this position owns the token
             // it mis-predicted and may not claim some other pattern could have begun there.
-            // Both tables are built once per utterance, so this is two array reads.
+            // Both tables are built once per utterance, so coverage is three array reads.
             AssertCoverageTablesMatch(tokens);
 
             float coverage =
@@ -1473,18 +1500,6 @@ namespace VoXR.Commands
             // ordering of the two.
             _coverageTokens = tokens;
 
-            // Nothing is charged at zero weight, so skip the predicate sweep entirely rather
-            // than compute a table that will be multiplied away. The tables are still cleared
-            // rather than left stale, so the accessors keep answering honestly ("nothing is
-            // charged") instead of returning the previous utterance's counts.
-            if (_coverageWeight <= 0f)
-            {
-                Array.Clear(_recognisedPrefix, 0, n + 1);
-                Array.Clear(_orphanRun, 0, n + 1);
-                Array.Clear(_forcedOrphanRun, 0, n + 1);
-                return;
-            }
-
             _recognisedPrefix[0] = 0;
             for (int i = 0; i < n; i++)
                 _recognisedPrefix[i + 1] = _recognisedPrefix[i] + (tokens[i] == UnkToken ? 0 : 1);
@@ -1531,7 +1546,7 @@ namespace VoXR.Commands
 
         // Guards the precondition that BuildCoverageTables ran over the array being scored.
         // Editor-only: the coupling is real but the check is not worth a branch in a shipped
-        // build, and every production caller is two lines away from its own build call.
+        // build, and both production callers build the tables at the top of the same method.
         [System.Diagnostics.Conditional("UNITY_EDITOR")]
         void AssertCoverageTablesMatch(string[] tokens)
         {
@@ -1571,6 +1586,12 @@ namespace VoXR.Commands
 
         // Could any registered active pattern begin a match at tokens[idx]? Deliberately
         // CONSERVATIVE: where the answer is uncertain it must be yes, and nothing is charged.
+        // Corrected by Amendment A3: the asymmetry below held only while coverage sat BELOW
+        // the selection barrier. Once coverage reorders, under-charging ONE candidate relative
+        // to a sibling is not "today's behaviour" — it is how the wrong command came to beat
+        // the right one, which is what A3 exists to close. The conservative default stands;
+        // the argument for it does not extend to the reordering regime.
+        //
         // The failure modes are not symmetric — over-charging destroys sequential extraction,
         // while under-charging merely leaves a score where it already is today.
         //
--- a/Runtime/Commands/VoxrCommandParser.cs
+++ b/Runtime/Commands/VoxrCommandParser.cs
@@ -69,12 +69,26 @@ namespace VoXR.Commands
         // stops the command firing.
         const float RequiredLiteralMissPenalty = 0f;
 
-        // Weight added to the score denominator per in-grammar word the sliding start skips
-        // before a match begins (issue #31). At 1.0 the score becomes the fraction of the
-        // utterance the pattern actually covers, so a lone one-element pattern found in the
-        // tail of a longer utterance scores 0.5 instead of a full 1.0 and falls below the
-        // default minScore. 0 restores the previous behaviour (skipped words cost nothing).
-        internal const float DefaultSkippedWordPenalty = 1.0f;
+        // Weight added to the score denominator per in-grammar token a match leaves
+        // unexplained: those the sliding start skipped to reach it (issue #31) and those left
+        // orphaned after it (issue #65 §5.2). At 1.0 the score is exactly the fraction of the
+        // utterance the pattern accounts for, so a lone one-element pattern found in the tail
+        // of a longer utterance scores 0.5 and falls below the default minScore.
+        //
+        // Renamed from skippedWordPenalty by DR-4, because one weight now governs both sides
+        // and the old name described only the leading half. The rename also stops hiding a
+        // coupling: setting this to 0 restores pre-#31 behaviour AND silently switches off
+        // the issue #42 fix, which "coverage weight" at least admits to.
+        internal const float DefaultCoverageWeight = 1.0f;
+
+        // Kept one minor version as DR-4 prescribes. Inert in practice: the type is internal
+        // and InternalsVisibleTo names only first-party assemblies, so there is no external
+        // call site for the warning to reach.
+        [Obsolete(
+            "Renamed to DefaultCoverageWeight — the weight governs orphaned trailing tokens "
+                + "as well as leading skipped ones (issue #65 §5.2)."
+        )]
+        internal const float DefaultSkippedWordPenalty = DefaultCoverageWeight;
 
         // Upper bound on optional elements in a single pattern for eager-commit analysis
         // (issue #25). ExpandOptionals enumerates 2^optionals concrete forms; past this the
@@ -88,7 +102,7 @@ namespace VoXR.Commands
 
         readonly VoxrSlotDefinition[] _slots;
         readonly VoxrCommandDefinition[] _commands;
-        readonly float _skippedWordPenalty;
+        readonly float _coverageWeight;
 
         // Per-slot lookup: first word -> list of (fullValue, wordCount), sorted longest first.
         readonly Dictionary<string, List<SlotValueEntry>>[] _slotLookups;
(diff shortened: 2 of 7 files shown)


jinwoo1601 (commented):
## Independent review — `review-pr`, **full profile**

**Head reviewed:** `60a064e428a6773e9d8d888daf788e1915d834bf` · **Base:** merge-base `f182c78`
**Angles run:** CLAIM + all 11 of the roster (LINE, GONE, XFILE, PIT, WRAP, REUSE, SIMP, EFF, ALT, CONV, GAP). None dropped.
**Verification:** every finding below carries an adversarial verdict — some from a `finding-verifier`, some re-derived directly against source or measured against the A/B rig.

Agents inherited nothing from the sessions that wrote this branch. The PR body was treated as **claims to audit**, and `#65` was read from the issue thread rather than from the PR's summary of it.

This branch already went through a `review-cycle` on its first commit (21 findings, applied in `60a064e`). This review is of the result. Two findings below are against fixes *that cycle* introduced, which is noted where it applies.

---

## Blocking-class

### B1 · Follow-up vocabulary is charged as orphans on the **default** decoder path
`Runtime/Commands/VoxrCommandParser.cs:301` · counterpart `VoxrCommandRecogniser.cs:196,319` · **CONFIRMED by measurement**

`VoxrCommandRecogniser` deliberately widens the *decoder* grammar with confirm/cancel vocabulary — `_grammar.Rebuild(_slots, commands, GetFollowUpGrammarWords())` — so `yes`, `confirm`, `cancel`, `abort`, `negative`, `do`, `it`, `go`, `never`, `mind`, `that` all come back from VOSK as **real tokens, not `[unk]`**.

`_startLiterals` is built from **patterns only** (`:301-337`), so `CanStartPattern` can never answer yes for any of them, and the new trailing term charges them. Measured against `main` on the shipped demo grammar:

| utterance | before | after |
|---|---|---|
| `disengage yes` | `1.000` | **`0.500`** — below `minScore`, fires nothing |
| `disengage confirm` | `1.000` | **`0.500`** |
| `disengage cancel` | `1.000` | **`0.500`** |
| `resume fire do it` | `1.000` | **`0.500`** |
| `cease fire affirmative` | `1.000` | `0.667` |

The confirm/cancel handler only runs when `_pending.HasPending`, so with nothing pending these are plain charged orphans. **This is not the ruled F5/M-6 carve-out**, which names `freeSpeechMode`/`InjectText`/batch — this is the normal grammar-constrained path, charging words the package itself put in the decoder grammar. A user saying "disengage, yes" gets nothing.

It is also **absent from the A/B corpus entirely**, which is why the build did not catch it — see M5.

*Fix shape:* the orphan predicate's notion of "in-grammar" should be the same word set the recogniser actually handed the decoder, or follow-up vocabulary should be exempted the way `[unk]` is.

---

## Major

### M1 · §9.6's `prepend` attribution is self-refuting, and hides a structural limit
`Planning~/features/coverage-in-selection/architecture.md` §9.6 · **CONFIRMED by measurement**

The as-built log attributes the single `prepend` newly-silent case to the leading term: *"`mode_navigation` matching `["navigation","mode"]` from index 1 is charged for the leading "switch" … `2/(2+1) = 0.667 → 0.500` … the **leading** term doing exactly what #31 always did."*

That mechanism yields `0.667` on **both** revisions (`main` already applied the leading charge post-selection), so it cannot move at all. Measured:

```
main: switch navigation mode -> mode_navigation:0.666667
head: switch navigation mode -> mode_navigation:0.500000
```

The real winner is `["switch","to","navigation"]` matched at index **0** (`to` misses without advancing the cursor), and the move is `2/3 → 2/(3+1)` — a **trailing** demotion, in the shipped demo grammar.

The consequence is larger than the misattribution. The healthy sibling `["navigation","mode"]` scores `0.667` and **structurally cannot replace it**, because earliest-start outranks score. So goal 2 — *"a candidate that explains more of the utterance must be able to beat one that explains less"* — is unreachable wherever the better candidate starts later. No doc states this.

### M2 · The **leading** half of the headline claim is untested — proven by mutation
`Runtime/Commands/VoxrCommandParser.cs:1266` · **CONFIRMED**

The PR's structural claim is that coverage is computed before candidates are compared. Because `IsBetterCandidate` ranks earliest start above score, and `SkippedBefore` is the *same constant* for every candidate at a given `startIdx`, the leading term can only change an outcome between two admissible candidates **sharing a start index** whose denominators differ. No test has that shape.

Verified by building the mutant — select on the trailing term only, report with both, i.e. `main`'s post-selection placement for the leading half — and compiling the full suite against both parsers:

```
CUR: pass=121 fail=0     MUT: pass=121 fail=0     diff(cur,mut) = EMPTY
```

Identical across 1095 utterances too (the 699-row corpus plus 396 literals scraped from `Tests~/`). **Half the headline change can be reverted with the suite staying green**, and brief invariant 1 ("the score a candidate is selected on equals the score it is reported with") goes with it — the mutant violates it and passes.

Two things bound this. `TryEagerCommit` is *structurally* immune: a non-`None` verdict requires `bestStartIdx == firstRecognisedIdx`, where the leading term is zero by construction. And the leading term is behaviourally inert across the whole 699-utterance corpus and the demo grammar — which is exactly why nothing catches the revert.

*Proposed test, verified by execution:* slot `burn = ["hard"]`; commands in order `decelerate → ["decelerate"]`, `decelerate_by → ["decelerate","by","{burn}"]`, `burn_now → ["{burn}","now"]` (makes `hard` a start, zeroing both trailing terms), `set_target_mode → ["set","target","mode"]` (makes `target` unstartable in-grammar filler). Input `"target target decelerate hard"` → current: **1 result, `decelerate_by 0.400`, `burn=hard`**; mutant: **2 results, `decelerate 0.333` | `burn_now 0.500`** — the spoken argument is discarded and the utterance splits in two. Assert intent **and** `results.Length == 1`.

### M3 · The new Inspector tooltip states something A3 makes false
`Runtime/Commands/VoxrCommandRecogniser.cs:41` · **CONFIRMED by inspection**

> `"filler is never charged, and leftover words that could begin another command are not either, so chained commands still work."`

Under Amendment A3 (`VoxrCommandParser.cs:1528`) a leftover word that *could* begin another command **is** charged when the candidate's own next required element failed there. This is user-facing text in the Inspector. It also contradicts a correction made elsewhere in the same diff: the tooltip keeps *"the score becomes the fraction of the utterance the pattern covers"* while `VoxrCommandParser.cs:77` now says *"Close to … but not exactly"*.

This also falsifies the body's claim that `VoxrCommandRecogniser.cs` is *"entirely rename"*. (`VoxrBatchTestRunner.cs` genuinely is.)

---

## Minor

### m1 · "A prior review-cycle is applied in `60a064e`" is true of the doc, not the code
`Runtime/Commands/VoxrCommandParser.cs:1574` · **CONFIRMED by inspection**

That cycle ruled D4's asymmetry rationale falsified and `architecture.md` carries the correction. The shipped comment still reads verbatim: `// The failure modes are not symmetric — over-charging destroys sequential extraction, // while under-charging merely leaves a score where it already is today.` The PR's own A/B refutes it — the A3 flip *was* under-charging changing the winner.

### m2 · Invariant 5 / requirements F11 is now stale and will mis-adjudicate the next amendment
`Runtime/Commands/VoxrCommandParser.cs:1267` vs `requirements.md` F11

A3 feeds `requiredAfterLastMatch` — per-candidate match state — into the orphan test. F11 still says the test *"takes no input from `IsBetterCandidate`, `MissedRequired`/`MatchedRequired`, or the `Score <= 0f` filter."* §4.2 amended F3 for A3 and left F11 untouched, so the two rows contradict each other. **The code is right and the invariant is stale:** the real concern is *cross-candidate* coupling, which A3 does not create. Worth restating before the deferred admissibility probe is argued against a clause it does not actually violate.

### m3 · The zero-weight short-circuit is dead work under a false rationale
`Runtime/Commands/VoxrCommandParser.cs:1476-1486` · four angles (EFF, GONE, SIMP, ALT, GAP)

`_coverageWeight` is `readonly`, so an instance at weight ≤ 0 takes this branch on every call and **never writes a non-zero value into the three tables** — the `Array.Clear`s provably cannot change a stored value, and the justification ("instead of returning the previous utterance's counts") describes an unreachable state. Deleting the block entirely is behaviour-identical.

Separately, at weight 0 the accessors stop meaning what they document: `SkippedBefore` is declared as *"the count `CountRecognisedTokens` produces by scanning"* and returns 0 instead. *This block was added by the previous review cycle* — an argument against a ruled decision, not an unfixed defect.

### m4 · `0 × ∞ → NaN` is newly reachable
`Runtime/Commands/VoxrCommandParser.cs:1264-1270` · LINE + PIT independently · **CONFIRMED by inspection**

The sanitiser `coverageWeight > 0f ? coverageWeight : 0f` rejects negatives and NaN but admits `+∞`. With both counts zero — the common case — `(0 + 0) * ∞` is `NaN`; `NaN <= 0f` is false, so it passes `IsBetterCandidate`'s floor and `ParseInternal`'s break, and a command is reported with `Score = NaN`, breaking DR-1. The recogniser's `>= minScore` rejects it; `VoxrBatchTestRunner`'s `< _minScore` does not. Only reachable via the public batch-runner constructor. On `main` an infinite weight gave `raw/∞ = 0`, not `NaN`.

### m5 · Three test-coverage holes in claims the PR relies on

- **`Coverage_ResidualHazard_…` cannot discriminate what it exists to pin.** Its `hard_stop` = `["hard","stop"]` matches 1 required and misses 1, so DR-7 **admits** it — the test passes identically under an orphan predicate defined over *admitted candidates*, which is exactly the definition invariant 5 forbids.
- **A3's "first position only" scoping is unpinned.** In all four A3 fixtures every token after the mis-predicted one is a non-starter, so "charge one then revert to the ordinary run" and "charge everything" give identical counts.
- **No score is asserted at any weight but `0` and `1`.** `* Mathf.Ceil(w)`, `* w²`, or a weight applied asymmetrically to the two terms all pass the entire suite.

### m6 · Conventions and stale text

`Tests~/Runtime/VoxrCommandParserTests.cs:1306` — the section banner still reads *"Written before the trailing coverage term exists … today every score below is 1.0"*, false in the merged tree, and the very next test scores `2/3`. · `VoxrCommandParser.cs:152` — the tables are said to match *"the `_matchSlotBuf` / `_resultBuf` pooling discipline"*, but both of those are `readonly` and sized once, never grown. · `:1261` — *"this is two array reads"*; it is three. · `:1533` — *"every production caller is two lines away from its own build call"*; they are ~28 and ~23 lines away. · The two renamed span tests broke the `SpanTieBreak_` grouping prefix that the PR's own new tests keep, and one kept its pre-change assertion message directly beneath a comment retracting it. · Two `CanStartPattern` tests call `BuildCoverageTables` inertly, implying a dependency the predicate does not have. · `LeftoverTokens_StillMatches`'s comment still claims `2/2 = 1.0` for an utterance the PR's own limitations say now scores `0.667`.

---

## Claim audit

Beyond M3 and m1:

- **"Closes … #42"** — `#42` was already closed by `#58` on 2026-08-12; `#65`'s own text records it. And symptom 2 is closed only for the common case: the diff's own `Coverage_ResidualHazard_…` asserts bare `decelerate` still wins at `1.0` with the burn level stranded, and `VoxrCommandParser.cs:457` says *"The hazard is REDUCED, not eliminated."* That residue is in none of the four "known limitations" bullets.
- **"No wrong command fires anywhere in the corpus"** — the report has seven columns, none of which is a wrong-command column, and the corpus has no ground-truth intent. The artifact supports *"no utterance's outcome changed to a wrong command"*, which is a delta claim, not an absolute one.
- **Breaking change** — the section names only the named-argument break. The larger, default-on break is measured in the same PR: `newly-silent 17`, `intent-change 11`, `newly-firing 0`, `slots-gained 0`. Existing grammars stop firing commands they used to fire, with no opt-in. That belongs under the heading that drives semver and the post-G2 `CHANGELOG`.
- **Verified sound and not reported:** the score formula; the `1/(1+2)` vs `2/3` and `2/2`-not-`2/7` arithmetic; `[unk]` transparency on both the ordinary and `_forcedOrphanRun` paths; "eleven sites" (exactly 11); the A3 "safe at one, flips at two" threshold; `398 + 1 = 399`; and **all three of `#65`'s stated constraints**, including the third — `TryEagerCommit` now calls `BuildCoverageTables` and `TryMatchScored(..., 0)`, so the scan and the flush gate on the same coverage-bearing score.

---

## Considered and refuted

- **REUSE: nothing.** Reusing `WarningForms` for the pattern-start walk would *under*-populate `_startLiterals` past its expansion cap and therefore **over-charge** orphans — the asymmetric failure the predicate's conservatism exists to prevent. The diff adds zero new `[unk]`-skip loops; the four are pre-existing.
- **WRAP: nothing.** Invariant 9's one-way property is sound, including under the zero-weight return and across buffer-vs-flush token arrays (for `Commit` they are the same string). `CountNumberSequenceWords` is exactly equivalent; the `minWords` cap is valid because `maxWords >= minWords` is enforced at both construction paths.
- **`minWords == 0` making every token a start** — raised by two angles, refuted twice: `VoxrSlotDefinition.NumberSequence` throws below 1, and `VoxrSlotAsset` routes through the same factory.
- **Three tables being over-engineering** — refuted: `_forcedOrphanRun` is not O(1)-derivable from `_orphanRun`, because `consumedEndIdx` provably lands on `[unk]`.
- **Comment volume** — refuted by measurement: the new region runs 0.96 comment:code against a densest window of **4.69** in the base file.
- **A3's altitude** — judged correct. The per-position admissibility probe is blocked in its crude form by the degenerate collapse, a fidelity tie-break would change the locked selection keys, and DR-6 alignment measured worse (35 newly-silent vs 17).
- **"The A/B rests on synthetic token streams, contra #65's real-decoder precondition"** — raised, and **refuted on both halves**. The locked design already ruled the decoder run out at §7.1 (*"No decoder run is needed. The committed fixture transcripts … are replayed through the parser with the scorer varied"*); #65's line sits under *Suggested starting points*, and the G1 pass #65 asked for resolved it into §7's plan. On vocabulary: every token of all 699 rows was checked against the 133-entry grammar array — **635 rows contain only tokens the constrained decoder can emit**. The out-of-grammar probes (`please`, `now`) touch **2 of the 28** behavioural rows, **0 of the 11 intent-changes**, and each is duplicated in effect by six in-grammar siblings on the same base utterance, so removing them changes no aggregate. Those two are deliberate coverage of the `freeSpeechMode`/`InjectText`/batch paths, already ruled (F5/M-6).

  What survives is narrower and worth stating: the corpus licenses nothing about **frequency**. The perturbations are exhaustive-by-construction, not sampled from acoustics, so nothing here says how often a real speaker produces a head-dropped second command or a trailing in-grammar orphan. That is a prior-probability gap, not a vocabulary-realism one — and it is what B1 fell through, since the class that is most likely in practice is the one the generator never produced.

---

## Not verifiable from this review

Stated so none of it reads as a passing check:

- **The Unity counts (EditMode 116/116, PlayMode 399/399).** They run in a separate host project unreachable from here; this package has no `Assets/` and compiles nothing standalone. Author claim, unaudited.
- **The A/B and perf figures.** The rig and corpus are gitignored and therefore **not in this diff**. The method was audited; the numbers were not re-run. Note M2 and M5 both concern what that corpus cannot see.
- **The prefab-instance override question.** Whether Unity remaps `m_Modifications` property paths through `[FormerlySerializedAs]` cannot be settled from this repo — no offline docs. The PR discloses this. Nothing shipped here is affected (no prefab, preset or `CustomEditor`), but it wants one round-trip in a host project.
- **On-device Quest behaviour.** Not required for this change (design §7.5), and unreachable.

---

Nothing here was fixed. Per the feature lane, this review is evidence for the G2 ruling and the human decides what happens to each finding.

jinwoo1601:
## Review fixes pushed — `4064ecc`

All findings from the review above were ruled **fix** and are applied on this branch.

**B1 (blocking) — fixed.** The parser now receives the same word list the recogniser hands the decoder (`additionalGrammarWords`, folded into `_startLiterals`). `"disengage yes"` is back to `1.000`; the new test asserts both halves as a pair — `1.000` informed, `0.500` uninformed — so the defect itself is pinned, not just its absence.

Worth recording: **the 699-utterance A/B is unchanged at 17 / 11.** No corpus utterance contains follow-up vocabulary, so no amount of re-running the measurement would have surfaced this. It was found by reading the recogniser against the parser, which is the case for keeping `XFILE` in the roster.

**M2 — fixed.** The leading term now has a test that reorders two same-start siblings (`1/(1+2)` vs `2/(3+2)`). The mutant that previously passed 121/121 with an empty diff now fails.

**M1, m1, M3 — fixed.** §9.6's `prepend` re-attributed to the trailing term; `CanStartPattern`'s surviving "failure modes are not symmetric" corrected in code, not just in the doc; the `[Tooltip]` sentence A3 falsified rewritten.

**m4, m5 — fixed.** Non-finite weights rejected in the sanitiser. `Coverage_ResidualHazard_` now uses a DR-7-**refused** candidate, so it finally discriminates the active-patterns rule from the admitted-candidates rule. A3's first-position-only scoping pinned at `0.500` vs `0.400`. Fractional weight pinned at `1/(1+2w)`.

**m3, m6 — fixed.** The zero-weight short-circuit is **deleted** rather than re-documented; four angles were right that it was dead. Conventions and the stale statements corrected, including the two renamed tests keeping the `SpanTieBreak_` prefix.

**Refuted, not fixed** — the "synthetic corpus" finding. 635 of 699 rows are decoder-emittable, the out-of-grammar probes touch 2 of 28 behavioural rows and 0 of 11 intent-changes, and design §7.1 had already ruled the decoder run out. What survives is a **frequency** gap, not a vocabulary one — recorded in architecture §9.8.

**Two items carried to G2 unruled**, both surfaced by this review rather than resolved by it:

1. **Coverage cannot reorder across start indices.** M1's real content: in `"switch navigation mode"` the healthy sibling scores `0.667` and structurally cannot replace the `0.500` winner, because earliest-start outranks score. Goal 2 is unreachable wherever the better candidate starts later. Newly named; belongs with requirements §8's demotion consequence.
2. **Requirements F11 / invariant 5 is stale.** A3 legitimately feeds per-candidate state into the orphan test. The invariant should be restated in terms of *other candidates' verdicts* before it mis-adjudicates the deferred admissibility probe.

**Verified:** EditMode 116/116, PlayMode 403/403. No fallout in `VoxrCommandRecogniserInjectionTests` (40/40), `VoxrPendingCommandTests` (35/35), `VoxrCommandSetTests` (5/5), `VoxrDynamicSlotTests` (17/17). Grammar pin still 133 entries, so the native harness does not re-baseline.

jinwoo1601:
## The two carried items, resolved — no code change

Both were ruled at review. Neither needed a commit: one is a requirements reword, the other a measurement. (`Planning~/` is local-only, so this comment is the shared record.)

### 1. "Coverage cannot reorder across start indices" — measured, and my framing was overstated

Ruled: measure before ruling. Enumerated every `(command, pattern, startIdx)` across all 699 utterances at full round-1 scoring — so a later candidate's score **includes** its leading charge, not the flattering suffix-parse number — looking for a round-1 winner under the gate with a later-starting admissible candidate over it.

| | count |
|---|---|
| round-1 winner under gate, later candidate over it | 29 |
| **recovered** — a later extraction round still fires something | **28** |
| **genuinely silent** | **1** |

**The property is real; the class is one utterance.** `"switch navigation mode"` (`mode_navigation` `0.667` → nothing). It needs a shape the other 28 lack: the round-1 winner must *consume through* the start index the better candidate needed. `["switch","to","navigation"]` matches at 0 **and again at 1**, skipping the "to", so it takes both tokens and round 2 begins at "mode", where nothing starts. Everywhere else the winner consumed a single token and sequential extraction reached the alternative on the next round.

So the review's own wording — *goal 2 is "unreachable wherever the better candidate starts later"* — is true of the selection keys and badly overstated as a claim about behaviour. Correcting it here rather than letting it stand: this is a `KNOWN_LIMITATIONS.md` entry, not grounds for reopening §5.3's key order over one prepend-shaped utterance.

### 2. Requirement F11 — narrowed to the property it actually protects

F11 said the orphan test may take no input from any per-candidate match state. A3 makes it read exactly one thing — *did **this** candidate's own next required element fail at this token?* — so the row contradicted the shipped code.

The code is right and the wording was too broad. What F11 exists to protect is that coverage must not become a function of DR-7's verdicts, because a rejected candidate withdrawing its pattern's claim on a token would silently lower a **different** candidate's score. A3 creates no such coupling — every candidate reads only its own state.

Narrowed to *"no candidate's orphan count may depend on any other candidate's verdict"*, and now tested in **both** directions rather than one: `Coverage_ResidualHazard_`'s fixture is deliberately DR-7-**refused**, so it passes only if a refused candidate's pattern still terminates the orphan run — which is precisely what distinguishes the two definitions.

---

Nothing further is outstanding from the review. The branch stands at `4064ecc`, EditMode 116/116, PlayMode 403/403.

PR #84, opened 2026-08-14
Rewrite the scoring docs for the re-derived model (#83)

### LINKED ISSUE #83: Rewrite the scoring model docs for the re-derived model (#65 backlog item 3)
Backlog item 3 of the locked scoring-model design, filed so it has a shared trace now that #65 is closing. The design doc lives in a gitignored planning area, so without this the item would exist nowhere public.

Items 1 (#72, fidelity) and 2 (#78, coverage in selection + DR-4) have merged. Both changed the model that `Documentation~/scoring.md` describes; neither rewrote it, by explicit G0 ruling, to keep the behaviour change and the doc rewrite separately reviewable.

**What is already done** (post-G2 of #78, PR #79): the `skippedWordPenalty` → `coverageWeight` identifier rename across the four docs that name it, a `CHANGELOG` behaviour-change note, and five `KNOWN_LIMITATIONS` entries.

**What this issue covers** — the prose that still describes the pre-#65 model:

- `Documentation~/scoring.md` — the per-element score table still shows `missed required literal −0.5` (item 1 set it to `0`); §80's formula and its surrounding text describe the leading charge only, applied *after* selection; the worked examples trace the old selection behaviour. It currently carries a `> **Changed in #65 §5.2.**` pointer as a stopgap.
- `Documentation~/command-recognition.md` — the "Skipped-word penalty" section and its heading anchor, which four other pages link to.
- `Documentation~/api/command-recogniser.md` — the row is updated but still links to that anchor.

Worth doing as one pass: the anchor rename touches every inbound link, and the model only makes sense explained whole — fidelity, coverage, selection order and the two gates together.

Reference material, all measured rather than derived: the A/B report, the 699-utterance corpus, and the start-index sweep under the feature's planning area, plus the worked cases in the #78 PR body and review.

jinwoo1601:
Closes #83.

Backlog item 3 of the locked scoring-model design (#65 §9, DR-5). Items 1 (#72) and 2 (#78) changed the model `Documentation~/scoring.md` describes; neither rewrote it, by explicit G0 ruling, so the behaviour change and the doc rewrite stayed separately reviewable. Both `scoring.md` and the command-recognition guide have been carrying a `Changed in #65 §5.2` stopgap over prose describing the leading skipped-word charge alone, applied *after* selection. Both stopgaps are gone.

## What

**`Documentation~/scoring.md`**

- **§2 rewritten as Coverage** — the formula with both terms named, the orphan test that stops counting at the first token another pattern could begin on, the Amendment A3 exception that charges a token the candidate's own next required element just mis-predicted, `[unk]`'s exemption *and* its transparency, the per-round restart, and a `coverageWeight` table including that `0` reverts the #42 protection along with the scoring.
- **The consequences are threaded through the rest of the page**, not parked in one section:
  - §1 now says the ratio is the *first* half of a score and notes coverage cannot move a candidate across the `> 0` floor or the admission rule.
  - §3 key 2 carries coverage. The `intercept track` example was itself stale — it claimed both siblings score `1.0` and key 3 separates them, which coverage made false (`0.60` vs `1.00`); key 3's surviving role is now stated. Added: **key 1 bounds what coverage can do**, since earliest start outranks score.
  - §4 says why `cease fire` keeps `2/2` rather than `2/7`, and the prefix-steals-the-head bullet now names which key guards which case.
  - §6 conditions 1 and 2 dropped the claim that the eager scan omits the charge — both paths compute the same number now — and gained the structural point that the whole-buffer condition leaves *trailing* coverage unable to decide a verdict at all. The `fire at` row is re-derived (`0.50` vs `0.33`, now failing the score gate too).
- **Worked examples.** **B** is inverted — it was the page's demonstration that a bare pattern discards a spoken argument, and now demonstrates the slot-filled pattern winning. **B2** (the same demotion with no sibling to land on) and **D** (a two-command utterance whose second command lost its leading word, where the cleanly-spoken command is the one rejected) are added, both ending in their session-log entries. **A** is re-derived: it quoted losing candidates at fidelity-only scores (`0.75`, `0.25`) that selection no longer sees (`0.60`, `0.17`).
- The field-to-rule map and symptom table gain the coverage rows and the question to ask first — compute `matched / elements` for the winning pattern; if the logged `score` is lower, the difference is what the match left unexplained.

**`Documentation~/command-recognition.md`** — `Skipped-word penalty` → **`Coverage`**, rewritten, with the trailing side in the table and the orphan amnesty stated.

**Anchor rename fan-out** — `api/command-recogniser.md`, `editor-testing.md`, and the guide's own two references follow `#skipped-word-penalty` → `#coverage`. `troubleshooting.md`'s link to worked example B follows its renamed heading.

**`Documentation~/troubleshooting.md`** — new symptom entry for the post-upgrade demotion (the change existing grammars are most likely to meet); the slot-missing entry re-pointed at its one surviving cause, a `coverageWeight` of `0`.

**`README.md` / `index.md`** — the one-line descriptions of `scoring.md` said "penalties".

## Two scope notes for the reviewer

1. **The function-word hazard section of `command-recognition.md` is outside the issue's stated scope** (which named the guide's Skipped-word penalty section only). It asserted that the bare pattern wins and the recognised slot value is discarded — the exact behaviour #65 removed — so leaving it would have shipped a page contradicting the code. Corrected in place; the `?by` remedy is kept, because `0.67` clears the default gate by only `0.07`.
2. **Left alone deliberately:** the `[Unreleased] → Added` entry for #47 describes `scoring.md` as covering `missed required literal −0.5` and a penalty "applied *after* selection, so it filters via `minScore` without changing which pattern wins". Both are stale, and both sit in the *same unreleased block* as the `Changed` entries that correct them. I did not rewrite another PR's entry on my own judgement — flagging it as your call.

## Validation

> **Updated after the review on this PR.** The original body claimed no `Runtime/` or `Tests~/` file was touched. That is no longer true: closing review finding 5 required new tests, so the suites were run rather than declared inapplicable. Every box below reflects the current head.

This PR is documentation plus five tests that pin what the documentation says. No `Runtime/` file is touched at any commit on this branch.

- [x] **Unity suites, both platforms, host project `VoXR TestGround` (Unity 6000.4.7f1):** EditMode **116/116**, PlayMode **408/408**, `failed="0"` on both. All five new tests ran in PlayMode and passed. `Tests~` was reverted and the `.meta` orphans cleaned; `git status` is clean apart from the branch's own changes.
- [x] **Every numeric claim measured, not derived.** 50 assertions replayed through the **real parser sources staged from this worktree** using the #65 A/B rig — every candidate score in §§1–3, §6 and §7, both `coverageWeight = 0` and the negative/NaN/infinity clamps, the follow-up-vocabulary amnesty, `[unk]` transparency, the A3 scores as literals, the residual hazard, the eager-verdict flip, and the leading-optional start-set effect. **50/50 pass.** Harness at `Planning~/features/coverage-in-selection/ab-rig/DocCheck.cs` (gitignored, so not in this diff); reproduce with `./stage.sh WORKTREE <outdir> && cd <outdir> && dotnet run -p:StartupObject=AbRig.DocCheck`.
- [x] **Every internal link and heading anchor resolves**, checked programmatically across `Documentation~/`, `README.md`, `KNOWN_LIMITATIONS.md` and `CHANGELOG.md` — this matters because the pass renames a heading other pages link to.
- [x] No `Changed in #65 §5.2` stopgap and no stale `skipped-word` prose remains. `skippedWordPenalty` survives only where it should: the `named … before #65` migration notes, the 1.4.0 row of the README version history, and the two historical CHANGELOG entries.
- [ ] On-device (Quest) check: not applicable — nothing here touches capture, the bridge, or lifecycle.

## Review

`review-pr` ran at the slim profile (six angles, fresh context) and is posted above. It found this PR **committing the very defect it exists to remove** — documentation asserting behaviour the parser does not have — four times. All four are fixed on the branch, along with the four lower-severity findings, in `e25443f`, `57279ba` and `14a5f60`. The most consequential: **coverage does not remove the #42 discarded-argument hazard, only its common case** — when the stranded value's own first word begins another pattern the bare form is charged nothing and strands the argument at the *default* weight, which is why the parser's construction-time warning was never narrowed. Measured, documented, and now pinned by a test.


jinwoo1601 (commented):
## Independent review — PR #84

**Depth profile:** issue lane → **slim** (`CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP`). No angles dropped. `LINE` died mid-run on a session limit and was re-run to completion; its findings below come from the completed re-run.

**Reviewed at:** `84a31facfad7190a26866a2b81bf277ca1c947cc`, against merge-base `bc0a40e`. Six finder agents inherited nothing from the session that wrote the docs; every surviving finding was then adversarially verified.

**Since reviewed:** the four CONFIRMED blockers plus two further confirmed findings are fixed in `e25443f` and `57279ba` (issue-lane rule: a confirmed blocker is the run's own unfinished work). Head is now `57279ba`. Everything below blocker is left unfixed and unruled, for you.

This PR's whole subject is documentation asserting behaviour the parser does not have. The review found the PR **committing that same defect four times**. That is the headline: the reconciliation pass itself needed reconciling.

---

## CONFIRMED — fixed on this branch

### 1. Coverage does not remove the discarded-argument hazard (#42) — only its common case · blocker

`scoring.md` §7 B said *"Coverage removes the wrong-pattern hazard"*; `troubleshooting.md` said coverage *"fixed it"* and offered `coverageWeight = 0` as the only remaining cause; the CHANGELOG called that *"the one surviving cause"*.

The parser says otherwise in its own words at `VoxrCommandParser.cs:484-489` — *"The hazard is REDUCED, not eliminated, and the residue is what this still warns about."* The orphan run stops at the first token that could begin a match, so when the stranded value's **own first word** begins some pattern, the bare candidate is charged nothing and strands the value exactly as before.

Measured against the real parser, at the **default** `coverageWeight`:

```
control  (nothing starts on "hard"):        decelerate:0.667[burn_level=hard burn]
residual (["hard","stop"] also registered): decelerate:1.000 | hard_stop:0.333
                                            ^^^ argument discarded, the #42 bug in full
```

A reader hitting the symptom would have checked the one documented cause, found `coverageWeight` was not `0`, and concluded the docs did not describe their case. It also explains a fact the page left unexplained: this is precisely why `WarnOnDroppableRequiredLiteral` was **not** narrowed when #65 shipped. Corrected in `scoring.md` §7 B and its symptom row, `command-recognition.md`'s hazard section, `troubleshooting.md`, and the CHANGELOG entry. *(Found by `GONE`; confirmed by measurement.)*

### 2. The trailing term is not round-independent "by construction" · blocker

`scoring.md` §2 claimed the orphan test does the same job on the trailing side that `searchStart` does on the leading side. It does not — `OrphanedAfter` takes no `searchStart`, and the run is a property of the utterance, not the round. **The page's own worked example D is the counter-case**, where `cease_fire` is charged for three tokens a later round then explains. §2 now states the asymmetry and points at D. *(Found independently by `LINE` and `GAP`; `GONE` flagged it too.)*

### 3. Trailing coverage *can* decide an eager verdict · blocker

`scoring.md` §6 condition 3 claimed it *"can never be what decides an eager verdict"*. The reasoning was that a candidate passing the whole-buffer test has nothing trailing it — true, but it misses that coverage runs in **selection**, which picks the candidate the conditions are then applied to. Measured on buffer `"decelerate hard burn"`:

```
coverageWeight 1.0 -> Commit     (coverage demotes bare `decelerate`; the buffer-spanning form is checked)
coverageWeight 0.0 -> None       (bare `decelerate` wins selection and fails the whole-buffer test)
```

*(Found by `LINE`; confirmed by measurement.)*

### 4. Wrong baseline for `coverageWeight = 0` · blocker

`troubleshooting.md` said zeroing it *"restores the pre-#65 scoring"*. It restores **pre-#31** — pre-#65 shipped the leading charge at `1.0`. The other two pages already said pre-1.4.0 / pre-#31, so the PR shipped three baselines for one setting. A reader would have zeroed the weight believing they were reverting to the well-tested #65 predecessor and silently lost the #31 protection the same page warns about. *(Found by `XFILE`.)*

### Also fixed

- **`coverageWeight` has a different lifetime from `minScore`** and the new troubleshooting entry offered them as interchangeable knobs. `minScore` is read per parse; `coverageWeight` is captured into a `readonly` field at parser construction, and there is **no `OnValidate` anywhere in the repo**, so a Play Mode Inspector edit does nothing until `RebuildParser()`. Following the entry live would have produced a false negative on a correct diagnosis. *(`PIT`, verified.)*
- **A pattern's leading optional elements are pattern starts too** — §2 said there were "two consequences" of the conservative start test; there are three. The walk continues past optionals (`VoxrCommandParser.cs:346-347`), so `["?please","fire"]` puts *both* words in the start set and a stray "please" terminates the orphan run **grammar-wide**. Measured: `"disengage please now"` scores `0.333` normally and `1.000` once an unrelated `["?please","fire"]` exists. Matters because the same page recommends bringing filler words in as optionals. *(`GAP`, verified.)*
- **"Three things are never charged"** in the guide — the third is not "never", given the Amendment A3 exception the reference documents four lines later. *(`LINE` + `XFILE`.)*
- **Two miscounts in my own claims.** The CHANGELOG said "the four pages linking to that anchor"; the base had exactly **three** inbound links across three files (`api/command-recogniser.md:14`, `command-recognition.md:132`, `editor-testing.md:153`) — `troubleshooting.md` followed a *different* rename. And validation box 3 said "two surviving `skippedWordPenalty` mentions"; `git grep` over tracked Markdown returns **seven**. *(`CLAIM`, confirmed by grep.)*
- **Validation box 1 was over-broad.** It claimed "every numeric claim measured", but the A3 scores were printed and then compared only to each other — an assertion built from the measured value cannot fail. Now asserted as literals (`0.60`, `0.33`). The harness also could not be re-run via the recipe the box cites (`stage.sh` did not stage it; duplicate `Main`); `stage.sh` and `Rig.csproj` are fixed, and the recipe now reproduces from a clean stage. Harness is at **50** assertions, and now pins this review's own findings. *(`CLAIM`.)*

---

## CONFIRMED — left for you, unfixed

### 5. Nothing pins these numbers against future drift · major

Verified in both parts. The scores this PR newly publishes are **not** protected by the committed suite:

- **Losing-candidate scores** (§7 A's `0.60` / `0.17`, §3's `0.60`) are structurally unobservable — `TryMatchScored` is `private`, so `InternalsVisibleTo` does not reach it, and only the winner's score reaches `VoxrCommand.Score`. The PR's own harness gets at them by reflection, and that harness is **gitignored and untracked**.
- **Worked example D's cross-round shape** (`0.40` then `0.67`) has no test. `Coverage_SequentialExtraction_ChargesNothingForALaterCommand` pins only the terminating direction (both at `1.00`), and `CreateParser()` never registers `approach_target`.

This is the exact drift class issue #83 exists to close, so it is worth your ruling rather than a silent pass. The concrete closer for the cross-round half: one test in `Tests~/Runtime/VoxrCommandParserTests.cs` on D's grammar asserting `0.4f` rejected then `0.667f` accepted. **I did not add it** — it would touch `Tests~/`, contradicting the PR's "no `Runtime/` or `Tests~/` file is touched", and it could not be verified from here (Unity host project, editor closed). Note also that `KNOWN_LIMITATIONS.md:394` names the intended future fix for D's shape; when it lands, example D and its two log entries become fiction.

### 6. `api/batch-test-runner.md` does not carry what this PR now routes readers to it for · minor

`scoring.md` §2 now sends readers there for the `[unk]`-exemption boundary and the "batch score is a lower bound" caveat. That page mentions neither, and its constructor table omits `coverageWeight` entirely — while `editor-testing.md:153` says "both constructors also take an optional `coverageWeight`", which the code confirms. *(`GAP` + `XFILE`.)*

### 7. `api/command-recogniser.md` overstates the serialization migration · major

*"Inspector values migrate automatically"*, unqualified — where `CHANGELOG.md:24` and the source comment at `VoxrCommandRecogniser.cs:52-58` both scope it: `[FormerlySerializedAs]` governs a field's own serialized data, and a prefab-**instance** override is stored as a literal property path that may not migrate. **Pre-existing text on a line this PR edited**, so it is your call whether it belongs in this pass. Worth noting the failure mode chains into finding 5's territory: a dropped override silently reverts `coverageWeight` to `1.0`, producing exactly the post-upgrade symptom the new troubleshooting entry describes — whose remedy list does not mention checking the override survived. *(`PIT` + `XFILE`.)*

### 8. Smaller, all pre-existing or cosmetic · minor / nit

- `command-recognition.md:139-140`'s tuning snippet (`commandRecogniser.minScore = 0.6f;`) **cannot compile** — both fields are private `[SerializeField]`s with no accessor. Pre-existing, in a section this PR rewrote around. *(`PIT`.)*
- The §7 JSON snippets show `"score": 0.33` where the real field holds the full float; the preamble says "abridged" about omitted fields but not about rounding. *(`PIT`.)*
- The `?by` remedy lost its log-entry JSON, which was the only place showing that an optional marker survives into `attempts[].pattern` as `decelerate ?by {burn_level}` — relevant to anyone grepping a session log. *(`GONE`.)*
- `troubleshooting.md` link text reads "worked examples B2 and D" but targets the B2 anchor only. *(`XFILE`.)*
- `README.md:164` / `index.md:14` dropped "penalties" for "coverage"; `RequiredSlotMissPenalty`'s `−1.0` is still documented in §1 but no longer signposted from either hub line. *(`GONE`.)*

---

## Refuted, so you can see what was considered and killed

- **"`nearly every token becomes a possible start` overstates the slot-initial hazard, because `CanStartPattern` caps its probe at `MinWords`."** REFUTED. `NumberSequence`'s default is `minWords: 1`, which is what the codebase actually uses for the `heading`/`elevation` slots the page's own examples reference — so at the default the predicate *is* trivially true for any digit token. Capping at `MinWords` makes the hazard broader, not narrower. The bullet is also inherited unchanged from `KNOWN_LIMITATIONS.md`, which the brief declared out of scope.
- **Anchors and links.** Independently re-checked twice: every internal link and heading anchor resolves across all tracked Markdown, no reference to any old anchor survives anywhere including non-`.md` files, and all six new anchors slug correctly under GitHub's algorithm (the `.` in `2.`/`B2.`/`D.` and the comma in D's heading are stripped, not hyphenated). No finding.
- **The `coverageWeight` clamp.** `scoring.md`'s "negative, NaN, or infinity → treated as `0`" holds exactly, including the NaN case (all NaN comparisons are false, so `> 0f` fails) and `-0f`. No finding.
- **Issue #83's stated scope** is fully covered; its fourth named defect (the score table still showing `−0.5`) was already fixed at the base revision, so nothing was left undone there.

---

## Not verifiable from here

- **The Unity suites were not run**, and the PR correctly says so — no `Runtime/` or `Tests~/` file is touched, independently confirmed (`git diff --name-only` returns eight `.md` paths at every commit on this branch). Finding 5 is about test coverage that *does not exist*, not about tests that were skipped.
- **On-device (Quest) behaviour** — not applicable to a docs change, and not asserted anywhere.
- **Rendered GitHub Markdown** — anchors were checked with a local slug implementation, not by rendering.
- The verification harness runs the **real parser sources staged from the worktree** under `dotnet` in WSL, which is what makes the numeric claims measured rather than derived. It is not a Unity run and does not substitute for one; nothing here needed one.

**Not approving and not merging** — GitHub blocks self-approval in a solo repo, and the G2 ruling is yours regardless. Findings 5–8 are posted unruled.

jinwoo1601:
Pushed `e25443f` and `57279ba`, addressing the four CONFIRMED blockers in the review above (the discarded-argument hazard overstated as closed; the trailing term claimed round-independent; trailing coverage claimed unable to decide an eager verdict; the wrong `coverageWeight = 0` baseline), plus the `coverageWeight` lifetime, the missing third consequence of the orphan predicate, the "never charged" overstatement, and the two miscounts in my own claims. Findings 5–8 are left unfixed for your ruling.

jinwoo1601:
All eight findings from the review above are now addressed on this branch.

**Finding 5 (major) — nothing pinned the published numbers.** Closed with five tests, using the internal coverage probes the admission test already uses rather than widening any `Runtime/` visibility:

| Test | Pins |
|---|---|
| `Coverage_SequentialExtraction_ChargesTheFirstCommandForASecondThatLostItsWord` | §7 D end to end — `2/(2+3)` = `0.40` rejected, then `2/3` = `0.67` fired, plus the control where both keep `1.00` |
| `Coverage_LosingCandidateTerms_MatchWhatTheWorkedExamplesPublish` | the coverage terms under §7 A's `0.60`/`0.17` and §3's `0.60` — unreachable through `VoxrCommand.Score` |
| `Coverage_LeavesTheStrandedArgumentHazard_WhenTheValueBeginsAPattern` | the residual #42 case, and that `?by` still reaches it |
| `Coverage_ALeadingOptionalLiteral_BecomesAPatternStartForTheWholeGrammar` | §2's third consequence, measured grammar-wide |
| `TrailingCoverage_ChangesTheVerdict_ByDecidingWhichCandidateIsChecked` | §6 condition 3 — `Commit` at weight `1.0`, `None` at `0` |

**Finding 6** — `api/batch-test-runner.md` now carries `coverageWeight` on both constructor rows and the lower-bound caveat §2 routes readers there for.
**Finding 7** — `api/command-recogniser.md` now scopes the migration claim to what `[FormerlySerializedAs]` actually covers, and the post-upgrade troubleshooting entry lists a lost prefab-instance override as a candidate cause, since it produces exactly that symptom.
**Finding 8** — the tuning snippet no longer shows an assignment that cannot compile (both fields are private `[SerializeField]`s); §7's preamble notes scores are shown to two decimals while the field carries the full float; the `?by` branch has its log entry back, with the point that the `?` survives into `attempts[].pattern`; `troubleshooting.md` links B2 and D separately; and both hub blurbs keep signposting the miss costs.

**Verification.** Unity, host project `VoXR TestGround`, both platforms: **EditMode 116/116, PlayMode 408/408, `failed="0"`**. All five new tests ran in PlayMode and passed. The doc-claim harness is at 50/50 and now pins this review's own findings. The PR body's validation section is corrected — it previously claimed no `Tests~/` file was touched, which this work made untrue.

jinwoo1601:
Closed the two items I had flagged as your call, per your ruling.

**`KNOWN_LIMITATIONS.md` — the #42 entry no longer reproduced.** For the grammar it names, "decelerate hard burn" now fires the slot-filled pattern at `2/3` = `0.67` with the burn level extracted. It is rewritten around the case that *does* survive — register anything that can begin a match on the stranded value's first word and the orphan run terminates there, so the bare pattern is charged nothing and wins at `1.0` with the argument discarded, at the default weight. Its "no threshold or penalty tuning reaches this" now attaches to that case, where it remains true, and the workaround records why the construction-time warning was never narrowed: `?by` reaches the residue, coverage does not. The same `coverageWeight = 0` baseline slip is corrected there too.

**`CHANGELOG.md` — the `[Unreleased] → Added` entry for #47** described the page as written rather than as it ships, in the same unreleased block as the entries correcting it. Corrected surgically: the `−0.5` miss cost, the "applied *after* selection … without changing which pattern wins" clause, and three counts that had drifted (three worked examples, four eager conditions, three symptom rows — the page now carries five, six and seven). The entry's scope and its account of what #47 itself found are untouched.

Documentation only, so the suites are unaffected by this commit — **EditMode 116/116, PlayMode 408/408** stand from `14a5f60`. Links and anchors re-checked. Every page that mentions `coverageWeight = 0` now names the same baseline.

Nothing further outstanding from my side. Head is `801aa93`.

PR #85, opened 2026-08-15
Move the completeness tests off the gate value so they witness their claim (#76)

### LINKED ISSUE #76: The #73 completeness tests all sit on exactly minScore, so they cannot witness the claim they test
PR #75 closed #73 by making the ordinary flush path refuse a command with an unfilled required slot, **independent of `minScore`**. The `review-pr` pass confirmed the fix but flagged that its three discriminating tests cannot demonstrate that independence, and the finding was not applied before merge.

All three construct a candidate scoring **exactly `0.60`**, the default `minScore`:

| Test | File |
|---|---|
| `SlotMissing_AboveGate_DoesNotFire` | `Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs` |
| `PartialMatch_AboveGateButIncomplete_EntersPendingInsteadOfFiring` | `Tests~/Runtime/VoxrPendingCommandTests.cs` |
| `IncompleteNewCommand_DoesNotCancelALivePending` | `Tests~/Runtime/VoxrPendingCommandTests.cs` |

Each is `(1+1+1+1-1)/5 = 0.60` — a 5-element pattern with one required slot stranded, sitting on the `>=` boundary.

### Why this is weak

The fix's claim is "an incomplete command does not fire **at any score**". A test pinned to the gate value can only witness "does not fire **at 0.60**". The two are indistinguishable while `0.60` happens to clear the gate, so the suite would stay green if the completeness check were deleted and replaced by a `> minScore` comparison — the exact regression these tests exist to catch.

### Why it is time-sensitive

They are also **fragile in a way that fails open**. Any change to the score denominator moves these candidates off `0.60`; if one lands below the gate, the test still passes — but for the wrong reason (rejected by score, never reaching the completeness branch), silently going vacuous with no failure to signal it.

`feat-coverage-in-selection` (backlog item 2 of #65, §5.2) **rewrites that denominator** — it adds `(skippedBefore + orphanedAfter) x coverageWeight` to it. These three tests are directly in its path.

### Suggested fix

Re-derive each onto a score comfortably **above** `0.60` so the completeness branch is the only thing that can reject the command, and add one case well above the gate (e.g. a single stranded slot on a long pattern, ~`0.85`) so the "at any score" claim has a witness that no plausible scoring drift can move under the gate. Per the #65 design's §7.3 discipline, the new expectations are **hand-derived and argued**, not updated to whatever the code emits.

### Related

- Fix under test: #73 / PR #75 (`fe33f33`)
- Collides with: #65 backlog item 2, `feat-coverage-in-selection`
- The second unapplied #75 finding — `VoxrBatchTestRunner`'s new completeness branch has zero test coverage — is worth folding into the same pass.

jinwoo1601:
Closes #76

## What

Almost entirely test-only. The one production change is a single `internal` test-only accessor added in review (see *Review fixes applied* below); the #73 behaviour is untouched, and what changes is whether the suite can tell that behaviour from a plausible impostor.

| Test | Before | After | Shape |
|---|---|---|---|
| `SlotMissing_AboveGate_DoesNotFire` | 0.60 | **0.75** | 8 required elements, trailing `{target}` stranded |
| `SlotMissing_OnTheGate_DoesNotFire` | — | **0.60** | #73's own repro, retained, now score-pinned |
| `SlotMissing_FarAboveGate_DoesNotFire` | — | **0.857** | 14-element pattern — the "at any score" witness |
| `PartialMatch_AboveGateButIncomplete_EntersPendingInsteadOfFiring` | 0.60 | **0.75** | same 8-element shape, `allowPartialMatch` |
| `IncompleteNewCommand_DoesNotCancelALivePending` | 0.60 | **0.75** | 8 elements, **medial** `{burn_level}` stranded |
| `Run_IncompleteCommandAboveGate_RejectedAsIncomplete` | — | **0.75** | new — `VoxrBatchTestRunner` |
| `Run_ExpectedRejection_IncompleteCommandAboveGate_Passes` | — | **0.75** | new — other side of that branch |

`SlotMissing_OptionalSlotOmitted_StillFires` (the over-correction guard: an omitted OPTIONAL slot is not a missing argument) is unchanged by this PR but was verified alongside the seven above, since the re-derivations move grammars around it.

## Why

With every element required and exactly one slot stranded the score is `(N − 2) / N`, so headroom above the gate is bought only with pattern length: 5 elements gives the 0.60 the issue objects to, 8 gives 0.75, 14 gives 0.857.

**Every case now pins its score, and every recogniser case reads the live gate off the recogniser.** This is the part the issue's suggested fix doesn't spell out but its fail-open argument demands: a bare does-not-fire assertion cannot distinguish *"refused as incomplete"* from *"never cleared the gate"*. The two guards close different doors — the pin catches the candidate moving, `MinScore` catches the gate moving underneath a candidate that didn't. With both, drift fails loudly instead of going vacuous, which is also what lets #73's original 0.60 repro stay in the suite honestly rather than being deleted for sitting on the boundary.

The suite now fails on both mutations it is meant to catch:

- **Delete `IsIncomplete`** (gate becomes score-only) → all five recogniser cases fire/cancel and fail; the batch-runner case starts passing where it asserts failure.
- **Move the denominator** so a candidate drops below the gate → the score pin fails, rather than the refusal quietly passing for the wrong reason.

### The issue's time-pressure premise has expired — the weakness has not

`feat-coverage-in-selection` has already merged; coverage is in the denominator at `Runtime/Commands/VoxrCommandParser.cs:1291-1297`. I verified against the real parser that all three tests **still sit at exactly 0.60** on `main`: coverage is zero for an utterance the pattern consumes end to end, so the collision the issue anticipated did not fire and nothing went silently vacuous. The structural weakness the issue identifies is unaffected by that, and is what this PR fixes.

Concretely, for the 0.857 witness the coverage term would have to charge more than six whole tokens of unexplained speech (`12 / (14 + c) < 0.60` needs `c > 6`) before the completeness branch stopped being the only possible rejector.

### Second unapplied #75 finding, folded in as the issue asked

`VoxrBatchTestRunner.PassesThresholds`'s completeness branch (`Runtime/Testing/VoxrBatchTestRunner.cs:199-206`) had **zero** tests reaching it. Two now do — one asserting the rejection (including that the reported reason is `required slot unfilled` and *not* the score wording), one asserting a grammar author can pin the refusal as the expected outcome. Both use a local grammar rather than the file's shared fixture, for the same reason the **two re-derived** recogniser tests moved off it — `SlotMissing_OnTheGate_DoesNotFire` deliberately stays on the shared fixture, being #73's reported shape.

## Review fixes applied

`review-pr` (slim profile) ran on this PR and confirmed four findings; all four are fixed here, and the review comment above records the six it refuted.

1. **`Assert.GreaterOrEqual(score, 0.6f)` could not fail** — it duplicated a private `minScore` the tests cannot read, *and* was implied by the exact score pin two lines above it. Fixed by adding `internal float MinScore => minScore;` to `VoxrCommandRecogniser`'s existing test-only accessor block, and reading it at all three guard sites. The guard now genuinely fails if the serialized default rises above a pinned candidate — the fail-open route the review identified. **This is the one production line in the diff**: `internal`, read-only, no behaviour.
2. Dropped an inaccurate ordering claim ("pins its score **before** asserting the refusal" / "the score pin **fails first**") — true for five of seven cases, but `PartialMatch_…` reads the pin off `pending.Value` and so cannot assert it before `pending.HasValue`.
3. Corrected "the recogniser tests moved off it" to "the two re-derived recogniser tests".
4. Named the eighth verified test rather than leaving an unnamed count.

## Validation

- [x] **Unity EditMode 118/118** and **PlayMode 410/410** green, `failed="0"`, zero `error CS` in either log (Unity 6000.4.7f1, host project `VoXR TestGround`, both platforms run separately). Verified per-`test-case` rather than on the root attribute alone, and all eight tests named in this description confirmed **discovered and passed** — not phantom-green. **Re-run after the review fixes**, since those touched production code; see the re-run comment below for the second run's counts.
- [x] Every expected score **hand-derived and argued in the test comment first**, then independently checked against the real parser via the `Planning~/` A/B rig staged from this worktree. All nine numbers matched, including the 0.60 baseline the issue describes. Per #65 §7.3 no expectation was read off the code. Two independent review angles later re-derived all nine by hand and agree.
- [x] No `CHANGELOG.md` entry: nothing user-facing changed. The sole production line is an `internal` accessor invisible outside the assembly — no public API, no behaviour.
- [ ] On-device (Quest) verification **not run** (WSL cannot). Not implicated: no runtime behaviour changed, and the only production line is a test-only property.

## Adjacent, deliberately left alone

The #75 review's finding 9 also noted the **recogniser's** `"required slot unfilled"` diagnostic reject string is asserted nowhere, unlike the four other rejection classes in `Tests~/Editor/VoxrCommandRecogniserDiagnosticTests.cs`. Issue #76 names only the batch-runner half, so I left it; it is a clean follow-up issue if you want it.

### DIFF
--- a/Runtime/Commands/VoxrCommandRecogniser.cs
+++ b/Runtime/Commands/VoxrCommandRecogniser.cs
@@ -912,6 +912,14 @@ namespace VoXR.Commands
         internal bool TestGrammarRebuildDeferred => _grammar.GrammarRebuildDeferred;
         internal float TestEffectiveBufferWindow => EffectiveBufferWindow;
 
+        // Read-only, and read by the completeness tests (issue #76) rather than copied as a
+        // literal. Those tests assert that a candidate CLEARS this gate, so that the refusal
+        // they then observe can only have come from the completeness term. Against a hard-coded
+        // 0.60 that assertion cannot fail: raising the serialized default would drop the
+        // candidates below the gate, reject them on score, and leave the tests passing without
+        // ever reaching the branch they exist to pin.
+        internal float MinScore => minScore;
+
         internal void TestForceTimeoutNow()
         {
             if (!_pending.HasPending) return;
--- a/Tests~/Editor/VoxrBatchTestRunnerTests.cs
+++ b/Tests~/Editor/VoxrBatchTestRunnerTests.cs
@@ -204,6 +204,103 @@ namespace VoXR.Tests.Editor
             Assert.IsTrue(result.Passed, result.FailureReason);
         }
 
+        // ─── Flush-path completeness (issues #73, #76) ──────────────────
+        //
+        // PassesThresholds refuses a command with an unfilled required slot independently of
+        // score, in step with the recogniser's own Step 7 gate — without it the harness would
+        // report PASS for an utterance the runtime refuses, certifying a grammar against
+        // behaviour the user will never see. That branch shipped in PR #75 with no coverage.
+        //
+        // The grammar is local rather than the shared fixture because on the shared
+        // five-element pattern a single stranded slot scores exactly 0.60, and a case pinned to
+        // the gate value cannot distinguish "rejected as incomplete" from "rejected on score"
+        // if the denominator ever moves (issue #76). Eight required elements — nothing optional,
+        // so the denominator is the pattern length outright — with seven matched and {target}
+        // stranded gives (7 x 1 - 1) / 8 = 0.75 instead.
+        const string IncompleteAboveGate = "launch all missiles from tube three at";
+
+        static VoxrBatchTestRunner CreateLongFormRunner() =>
+            new VoxrBatchTestRunner(
+                new[]
+                {
+                    new VoxrSlotDefinition("weapon", new[] { "missiles", "torpedoes" }),
+                    new VoxrSlotDefinition("target", new[] { "hotel one", "hotel two" }),
+                    new VoxrSlotDefinition("quantity", new[] { "all", "one", "two" }),
+                    new VoxrSlotDefinition("tube", new[] { "one", "two", "three" }),
+                },
+                new[]
+                {
+                    new VoxrCommandDefinition(
+                        "launch_weapon",
+                        new[]
+                        {
+                            new[]
+                            {
+                                "launch",
+                                "{quantity}",
+                                "{weapon}",
+                                "from",
+                                "tube",
+                                "{tube}",
+                                "at",
+                                "{target}",
+                            },
+                        }
+                    ),
+                }
+            );
+
+        [Test]
+        public void Run_IncompleteCommandAboveGate_RejectedAsIncomplete()
+        {
+            var runner = CreateLongFormRunner();
+            var result = runner.Run(
+                new VoxrTestCase
+                {
+                    input = IncompleteAboveGate,
+                    expectedIntent = "launch_weapon",
+                    description = "Clears minScore but leaves {target} unfilled",
+                }
+            );
+
+            Assert.AreEqual(
+                6f / 8f,
+                result.Score,
+                0.001f,
+                "the hand-derived score no longer holds — re-derive it and argue the new value, "
+                    + "because a candidate that fell below minScore would be rejected for a "
+                    + "reason that has nothing to do with completeness"
+            );
+            Assert.IsFalse(result.Passed, "an incomplete command must not be certified");
+            StringAssert.Contains("required slot unfilled", result.FailureReason);
+            StringAssert.DoesNotContain(
+                "minScore",
+                result.FailureReason,
+                "and the reported reason must be the one that actually rejected it"
+            );
+        }
+
+        [Test]
+        public void Run_ExpectedRejection_IncompleteCommandAboveGate_Passes()
+        {
+            // The other side of the same branch: a grammar author pinning the runtime's refusal
+            // as the expected outcome gets a PASS, so the harness can be used to hold the
+            // behaviour in place rather than only to notice it.
+            var runner = CreateLongFormRunner();
+            var result = runner.Run(
+                new VoxrTestCase
+                {
+                    input = IncompleteAboveGate,
+                    expectedIntent = null,
+                    description = "Incomplete above the gate is expected to be refused",
+                }
+            );
+
+            Assert.IsTrue(result.Passed, result.FailureReason);
+            Assert.IsNull(result.ActualIntent, "nothing may be accepted");
+            Assert.AreEqual(6f / 8f, result.Score, 0.001f, "and it really did clear minScore");
+        }
+
         [Test]
         public void Run_ExpectedAcceptance_ButConfidenceRejects_Fails()
         {
--- a/Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs
+++ b/Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs
@@ -888,23 +888,61 @@ namespace VoXR.Tests.Runtime
         // TryEagerCommit has refused a command with an unfilled required slot since #66. The
         // ordinary flush path — the one on by default — had no such condition, so a slot-missing
         // candidate that cleared minScore fired with its argument simply absent.
+        //
+        // The claim is "an incomplete command does not fire AT ANY SCORE", so the cases below
+        // are spread across the range instead of clustered on the gate value (issue #76). A case
+        // pinned to 0.60 witnesses only "does not fire at 0.60", which the suite would keep
+        // reporting green if the completeness term were deleted and the gate loosened to
+        // `> minScore` — the exact regression these tests exist to catch.
 
-        [Test]
-        public void SlotMissing_AboveGate_DoesNotFire()
+        // Pins the candidate's score against the caller's hand-derived expectation, then asserts
+        // the recogniser refuses to fire it.
+        //
+        // The pin is the point. Without it a does-not-fire assertion FAILS OPEN: any change to
+        // the score denominator that pushed the candidate below minScore would leave the test
+        // green for the wrong reason — rejected on score, never reaching the completeness branch
+        // — with nothing to signal that it had gone vacuous. Per the #65 §7.3 discipline the
+        // expected scores are hand-derived and argued at each call site, never updated to
+        // whatever the code happens to emit.
+        void AssertIncompleteDoesNotFire(
+            VoxrSlotDefinition[] slots,
+            VoxrCommandDefinition[] commands,
+            string utterance,
+            float expectedScore,
+            string strandedSlot
+        )
         {
-            // #73's repro, and the deliberate counterpart to the boundary test above: both land
-            // on exactly 0.60 and both clear the >= gate, so score cannot be what separates them.
-            // "launch all missiles target" matches launch, {?quantity}, {weapon} and target, then
-            // strands {target}: (1 + 1 + 1 + 1 - 1) / 5 = 0.60. Firing it hands the handler a
-            // launch order with nothing to launch at.
-            ConfigureWithSyncDefaults();
+            var probe = new VoxrCommandParser(slots, commands).Parse(utterance);
+            Assert.AreEqual(1, probe.Length, "the utterance must yield exactly one candidate");
+            Assert.IsTrue(probe[0].IsMatch, "and that candidate must be a match");
+            Assert.AreEqual(
+                expectedScore,
+                probe[0].Command.Score,
+                0.001f,
+                "the hand-derived score no longer holds — re-derive it and argue the new value "
+                    + "before touching anything else, because everything below rests on it"
+            );
+            Assert.GreaterOrEqual(
+                probe[0].Command.Score,
+                0.6f,
+                "the candidate must clear the default minScore, or the refusal below proves "
+                    + "nothing about completeness"
+            );
+            Assert.IsFalse(
+                probe[0].Command.HasSlot(strandedSlot),
+                $"'{strandedSlot}' must really be unfilled — that is what the recogniser rejects on"
+            );
+
+            _recogniser.Configure(slots, commands);
+            _recogniser.BufferWindow = 0f;
+            _recogniser.CommandCooldown = 0f;
 
             int recognisedCount = 0;
             string unrecognised = null;
             _recogniser.OnCommandRecognised += _ => recognisedCount++;
             _recogniser.OnUnrecognisedSpeech += text => unrecognised = text;
 
-            _recogniser.InjectText("launch all missiles target");
+            _recogniser.InjectText(utterance);
 
             Assert.AreEqual(
                 0,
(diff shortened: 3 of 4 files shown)


jinwoo1601 (commented):
# Independent review — PR #85

**Head reviewed:** `9a7ecdc88ebfb28c5d248c56f74330974bb972b2` · **merge-base:** `de5cff8ebc0f81e2f2e99e59754b85643a16a661`
**Method:** six finder agents with no context from the session that wrote the code, deduped, then eight adversarial verifiers over ten surviving findings. **4 confirmed, 6 refuted.**

**No blocker.** Per the issue lane nothing has been fixed on the branch; everything below is unruled and yours to decide.

**The arithmetic is solid.** Two angles re-derived all nine asserted scores independently, by hand, from `TryMatchScored` — walking each pattern element against each utterance token — and both agree with each other and with the committed numbers. `LINE` additionally proved the coverage term is **zero in all six grammars** rather than assuming it (`consumedEndIdx` lands on `tokens.Length` in every case, so `_orphanRun[n] = _forcedOrphanRun[n] = 0`, and `SkippedBefore(0,0) = 0`), and confirmed every `probe.Length == 1` assumption against `ParseInternal`'s extraction loop and `IsBetterCandidate`'s filters. `(N−2)/N`, `(7×1−1)/8 = 0.75`, `(13×1−1)/14 = 0.857`, `12/(14+c) < 0.60 ⇒ c > 6`, `0.857 − 0.6 = 0.257` — all check out.

**`GONE` returned zero findings**, mechanically: 35 deleted lines, and **not one `Assert.*` call removed anywhere in the diff**. It specifically refuted the coverage-loss hypothesis I would have worried about most — the `{?quantity}` optional-slot shape. `SlotMissing_OnTheGate_DoesNotFire` still passes `MakeSlots(), MakeCommands()`, i.e. the identical shared fixture, and `SlotMissing_OptionalSlotOmitted_StillFires` is byte-identical to base. The one place an optional slot did disappear (`VoxrPendingCommandTests.cs:129` → required `{quantity}`) was **inert** coverage: the base utterance filled `{?quantity}` with `"all"`, so the `!IsOptionalSlot` skip never fired there.

**`XFILE` and `PIT` came back clean on the mechanics.** Accessibility resolves (`VoxrCommandParser` is `internal`, `AssemblyInfo.cs` grants IVT to exactly the asmdef `name`, 64 pre-existing PlayMode uses prove the path); the asserted literal matches `VoxrBatchTestRunner.cs:204` character for character; `DoesNotContain("minScore", …)` holds on every reachable `FailureReason` path; every score assertion supplies a delta and NUnit binds the intended `AreEqual(double, double, double, string, …)` overload with no silent misbinding; no exact float equality; every nullable deref is `HasValue`-guarded; assembly placement is right.

---

## Confirmed

### 1. The "clears the gate" guards hard-code a copy of a private `minScore` — and cannot fail as written — **minor**
`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:925` · `Tests~/Runtime/VoxrPendingCommandTests.cs:179`, `:245`

Raised independently by `XFILE` and `PIT`. Two distinct problems in one line.

**It is unreachable as a failure.** The exact pin two lines above already fixes the score to a constant:

```
921:  Assert.AreEqual(expectedScore, probe[0].Command.Score, 0.001f, …)
925:  Assert.GreaterOrEqual(probe[0].Command.Score, 0.6f,
928:      "the candidate must clear the default minScore, or the refusal below proves nothing about completeness")
```

With `expectedScore ∈ {3f/5f, 6f/8f, 12f/14f}` the `GreaterOrEqual` is implied by the `AreEqual`. It buys documentation, not discrimination.

**And the quantity that could falsify its message is never read.** `minScore` is `[SerializeField] float minScore = 0.6f` (`VoxrCommandRecogniser.cs:31`), private, and absent from both test-only accessor blocks (`:341-346`, `:903-913`) — which do expose `BufferWindow`, `PendingTimeout`, `TestEffectiveBufferWindow` and six others, so the pattern exists and this field simply is not in it. Verifier walked `minScore = 0.8f` concretely: all three recogniser cases take `cmd.Score < minScore` at `:726`, `continue` at `:761`, and `OnUnrecognisedSpeech` still fires at `:822` — so `recognisedCount == 0` and `unrecognised == utterance` both pass without `IsIncomplete` ever being evaluated. That is exactly the fail-open mode this PR exists to close, reintroduced through `minScore` instead of through the denominator.

Two qualifications the verifier established, both of which cut against overstating this. First, a raised default is **not** silent suite-wide — `MissedLiteral_BoundaryCase_ExactlySixTenths_Fires` (`:875-877`) goes red loudly. The gap is that a maintainer fixing *that* test gets no signal these three guards went hollow. Second, the two batch-runner tests are immune: they pass `minScore` explicitly.

`git show de5cff8` confirms `GreaterOrEqual` appears in neither file at base — **this idiom is introduced by this diff**, not inherited, so "it's the house style" is not available as a defence.

Remedy options, in ascending cost: delete the tautological line; or name the literal once as a shared constant; or add `internal float MinScore => minScore;` mirroring the existing `internal float TestEffectiveBufferWindow => EffectiveBufferWindow;`. Note the third contradicts this PR's own "No `Runtime/` production code changed" claim.

### 2. "Every case now pins its score before asserting the refusal" is false for two of seven — **nit**
`Tests~/Runtime/VoxrPendingCommandTests.cs:163` vs `:173` · `Tests~/Editor/VoxrBatchTestRunnerTests.cs:299` vs `:301`

Verifier recorded the order for all seven cases, not just the alleged two: cases 1, 2, 3 (the helper), 5 and 6 pin first; cases 4 and 7 do not. In `PartialMatch_AboveGateButIncomplete_…` the pin reads `pending.Value.Score`, so it structurally cannot precede `Assert.IsTrue(pending.HasValue)` — but it *could* precede `Assert.IsFalse(recognised.HasValue)` at `:163`, and `IncompleteNewCommand_…:236` proves the separate-probe idiom that would make the ordering fully achievable. Case 7 has no excuse at all: `result` is materialised at `:290` and the pin could sit at `:299`.

I tried to rescue this as "pins, rather than merely asserts" — presence not sequence. The adjacent bullet **"the score pin fails first"** forecloses that reading with an unambiguously ordinal word, and *that* clause is itself wrong for case 4 (under the denominator mutation `:163` and `:167` pass and the pin fails third). The ambiguity defence makes the body worse, not better.

Materiality is nil: NUnit reds the test on any failing assertion regardless of position, and in both cases the pin is still the assertion that reports. This is wording.

### 3. "for the same reason the recogniser tests moved off it" is true of two of three — **nit**
`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:971-977`

`SlotMissing_OnTheGate_DoesNotFire` deliberately does *not* move off the shared fixture — it passes `MakeSlots(), MakeCommands()`, and its own comment says so. The verifier also found the generic is loose in a second way the finding did not name: both pending-file tests configured inline grammars in base *and* head, so they never sat on that file's shared fixture either. The batch-runner half of the sentence is true. Reading "the recogniser tests" restrictively as "the ones that moved" is self-sealing and does not survive.

Low materiality — the table row and the Why paragraph both disclose the retention explicitly, and the real rationale is stated verbatim in the added test comment.

### 4. "all eight named tests" points at a named set of seven — **nit**
PR body, Validation box 1

The body names exactly seven distinct test identifiers; the diff adds or modifies exactly seven `[Test]` methods (2 + 3 + 2). The eighth is almost certainly `SlotMissing_OptionalSlotOmitted_StillFires`, verified as an unchanged guard but never named in the body. Verifying more than the diff changed is good practice, not an error — the defect is narrower: the one claim in that box *designed* to be independently checkable ("not phantom-green") cannot be fully checked by a reader with no eighth name to look up. Verifier raised confidence from low to high; the arithmetic is exact, not a judgement call.

---

- **"The batch-runner file has no over-correction guard for optional slots"** (raised minor). The remedy it asks for already exists: `SlotMissing_OptionalSlotOmitted_StillFires` runs *verbatim the proposed utterance* `"launch missiles target hotel one"` against the identical pattern and asserts the command fires with `quantity` unfilled. Dropping `!IsOptionalSlot` reds it. Also: the mutation lives in shared code (`VoxrCommandParser.cs:1703`), not in the batch-runner block, and the project's verification binding requires running **both** platforms — there is no configuration where the PlayMode guard is skipped. Base had zero batch-runner completeness coverage, so this diff strictly improves matters.
- **"The flag-OFF side never excludes the pending route"** (raised minor). `IncompleteNewCommand_DoesNotCancelALivePending` *is* an OFF-flag witness: `set_burn` uses the two-arg ctor, is pinned incomplete at `6f/8f` above the gate, and under the only realistic mutation (dropping the `AllowPartialMatch` term at `:730`) `EnterPending` cancels the live pending unconditionally — `cancelledCount == 0` fails, and the follow-up assertion fails too. The suite already goes red without the proposed assertion.
- **"The expected-rejection batch test's score pin doesn't say which branch rejected it"** (raised minor). Under the mutation these tests exist to catch, it fails *on its own*: the candidate is accepted, `Run` takes the `expected rejection but got intent` path, and both `Assert.IsTrue(result.Passed)` and `Assert.IsNull(result.ActualIntent)` go red. The vacuity scenario needs someone to raise a public ctor default on a shipped harness API — not the silent denominator drift #76 was raised about.
- **"The probe parser is constructed with different arguments than the parser it pins"** (raised minor/nit by two angles). Refuted with a proof, not a case check: the diverging args reach the score through exactly one channel (`_coverageWeight` at `:1295`, `_startLiterals` at `:1601`), `BuildEffectiveSlots` is identity with no providers registered, `coverageWeight` is the same value on both sides, and `GetFollowUpGrammarWords` only ever *adds* start-literals — which monotonically shortens orphan runs. So the recogniser's score is **≥ the pinned probe score, always**. The pin is a valid lower bound, not a coincidence of the chosen utterances, and a denominator change charges the probe *more*, so the pin fails first. The suggested alternative is genuinely unavailable: nothing fires and nothing pends, so no event carries a score, and `LastMatchDiagnostics` is `internal` *and* `#if UNITY_EDITOR`-gated. Residual: the helper never records *why* a 2-arg probe is a sound stand-in — a one-line comment gap.
- **"`probe[0].IsMatch` is a tautological assertion"** (raised nit). Factually right that it cannot fail today — one construction site, all match-constructed. But `Documentation~/api/data-types.md:75` publishes the contract "`Command` … only valid when `IsMatch` is true", and the next line dereferences `.Command`. It is the caller-side precondition the public API requires, correctly placed.
- **"The Step-5 consumer of `hasCompleteNewCommand` is unexercised for the incomplete case"** (raised nit, self-conceded as redundancy). Both consumers read one boolean computed once at `:646-660`; no mutation of the `&& !IsIncomplete(cmd)` term can change Step 5 without also changing the Step-6 cancel outcome the test already asserts. Pre-existing shape besides — base injected `"set burn to now"` against the same grammar, equally producing no follow-up.

---

- **The Unity suites were not re-run.** The ticked "EditMode 118/118 / PlayMode 410/410" box is a claim this review could not independently confirm by execution. It is *not* unbacked: `CLAIM` reconciled both counts exactly against the base revision — EditMode 107 `[Test]` + 9 `[TestCase]` = 116, +2 new; PlayMode 392 non-expanding attributes + 16 `ValueSource` cases from `Replay_ManifestCase` = 408, +2 new. What could not be confirmed is that the tree hashed to `9a7ecdc` when the run was produced. As with PR #75, no artifact was filed under `Planning~/verification-runs/<sha>/`.
- **On-device (Quest) verification.** Correctly marked not run and correctly argued not implicated — the diff touches only `Tests~/`.
- **The second validation box** ("checked against the real parser via the `Planning~/` A/B rig") is unauditable from the diff: `Planning~/` is gitignored, so the rig and its output are not in the repo. The claim is nonetheless *corroborated independently* — two review angles re-derived all nine numbers by hand and agree.

---

*This review was produced by agents that inherited nothing from the session that wrote the code, and it reviews that session's own work. Findings were deduped for overlap, not for comfort — three of the four confirmed findings are against the PR author's own description of the work.*

jinwoo1601:
**Review fixes pushed** — `b0a8f08`. All four CONFIRMED findings from the review above are addressed.

| Finding | Severity | Fix |
|---|---|---|
| 1. "clears the gate" guards hard-code a copy of a private `minScore`, and cannot fail | minor | `internal float MinScore => minScore;` added to `VoxrCommandRecogniser`'s existing test-only accessor block; all three guards now read it |
| 2. "pins its score **before** asserting the refusal" false for two of seven | nit | ordering claim dropped from the PR body, along with the "the score pin **fails first**" clause the verifier showed was wrong for the same case |
| 3. "the recogniser tests moved off it" true of two of three | nit | corrected to "the **two re-derived** recogniser tests", noting `SlotMissing_OnTheGate_DoesNotFire` deliberately stays on the shared fixture |
| 4. "all eight named tests" points at a set of seven | nit | eighth named: `SlotMissing_OptionalSlotOmitted_StillFires` |

On finding 1, I took the accessor rather than the cheaper remedies. Deleting the tautological line or naming the literal as a shared constant would both have removed the symptom while leaving the fail-open route open — the guards would still never read the gate. `internal float MinScore => minScore;` mirrors the adjacent `TestEffectiveBufferWindow` exactly, adds no public API and no behaviour, and makes the assertion able to fail: the two guards now close different doors, with the score pin catching the candidate moving and `MinScore` catching the gate moving underneath a candidate that didn't.

This does mean the PR is no longer strictly test-only — **one production line** — so the body's "no `Runtime/` production code changed" claim was corrected too, and the suites were re-run rather than relying on the pre-fix result.

**Re-run: EditMode 118/118, PlayMode 410/410, `failed="0"`, zero `error CS`.** Totals unchanged from the pre-fix run, so nothing was silently added or un-discovered. The accessor's visibility across the `InternalsVisibleTo` boundary is proven rather than assumed — all three `_recogniser.MinScore` call sites resolved at compile time and executed; had it been invisible the run would have failed with `CS0122`/`CS1061`.

The six refuted findings were left alone, as posted.

PR #86, opened 2026-08-15
Apply the completeness test at the follow-up slot-fill exit (#77)

### LINKED ISSUE #77: Pending state machine applies no completeness test at any of its three firing exits
#73 established that a command missing a required argument must not fire, and PR #75 enforced it on the ordinary flush path via `IsIncomplete` (`VoxrCommandRecogniser.cs:819`) at Step 4 and Step 7. `IsIncomplete` delegates to `VoxrCommandParser.HasUnfilledRequiredSlot` (`:1347`).

The pending state machine has **three** exits that fire a command, and **none** of them applies that test. Found while verifying PR #75; ruled not chargeable to that PR, so filing separately.

## 1. Follow-up slot-fill can complete a command that is still incomplete — a real defect

`PendingCommandHandler.TryFollowUpSlotFill` (`:104`) walks `pending.UnfilledSlots` in order, `break`s at the first slot it cannot fill, and then returns a command as long as **at least one** new slot was filled:

```csharp
// Must have filled at least one new slot
if (_followUpSlotBuf.Count == existingSlots.Length)
    return null;
```

`ComputeUnfilledSlots` (`:225`) lists **required** slots only — the same predicate `HasUnfilledRequiredSlot` uses — so with two or more unfilled required slots, an utterance that fills the first and not the second returns a command that `IsIncomplete` would reject. `VoxrCommandRecogniser.cs:632` then calls `_pending.Complete(followUpResult.Value)` and fires it, with no completeness test anywhere on that path.

This is pointed rather than incidental: #73's fix routes incomplete commands **to** the slot-fill path. That path is now the designated destination for exactly the commands #73 refuses to fire, and it does not re-check them on the way out.

The re-score does not save it either — `ScoreFollowUp` re-scores against the matched pattern, so a partly-filled command can score close to a complete one, which is precisely the "score is not a completeness test" argument #73 was decided on.

## 2 & 3. Two exits that may be deliberate but are unstated — needs a ruling, not necessarily a fix

- **`TryHandleConfirmCancel` on a `PartialMatch` pending** (`VoxrCommandRecogniser.cs:569`). A `PartialMatch` pending has unfilled required slots by construction. Confirming fires it as-is.
- **`HandleTimeout` with `FireAsIs`** (`:444`). Same, on the timeout path.

Both sit behind opt-ins that default off (`AllowPartialMatch`, `PendingTimeoutBehavior.FireAsIs`), and "fire as-is" plainly names the intent — so these are arguably working as designed. But #73 concluded that firing a command with an absent argument hands the handler a command it cannot execute, and that reasoning does not obviously stop at an opt-in. Either these are deliberate and should say so in `Documentation~/`, or the completeness condition belongs here too.

## Suggested resolution

- Case 1: apply `IsIncomplete` before `_pending.Complete(...)`, and on failure keep the pending alive with the remaining slots rather than firing — a partial fill is progress toward the command, not the command.
- Cases 2 and 3: rule whether the opt-ins are intended to authorise firing an incomplete command. If yes, document it on `allowPartialMatch` and `pendingTimeoutBehavior`.
- Coverage: cases 1 and 3 have no test that constructs a **two**-required-slot pending, which is what makes case 1 reachable.

## Related

- #73 / PR #75 (`fe33f33`) — the flush-path condition this mirrors
- #66, #70 — the two `TryEagerCommit` completeness conditions
- #76 — the other unapplied #75 review finding

jinwoo1601:
Closes #77

## What

**Case 1 — fixed.** `VoxrCommandRecogniser` Step 5 now applies `IsIncomplete` to the follow-up slot-fill result before completing it. When the result is still missing a required slot it calls the new `PendingCommandHandler.AdvanceSlotFill`, which re-arms the pending on the newly filled command with `UnfilledSlots` recomputed, and returns `ReEnteredPending` so `OnCommandPending` fires again carrying the updated command. The command fires only once nothing required is left. Editor diagnostics gain a matching `still pending (partial: unfilled [...])` entry with `accepted: false`.

**Cases 2 and 3 — ruled deliberate, documented, no behaviour change.** Confirming a partial-match pending and `FireAsIs` on timeout both still fire as-is. `Documentation~/command-recognition.md` gains *The two ways an incomplete command still fires*, naming them as the deliberate exceptions and spelling out the consequence — opt into either and the handler must tolerate every required slot being absent — with a `HasSlot` example, since `GetSlot` returns `string.Empty` rather than `null` for an unfilled slot. Cross-referenced from the `FireAsIs` row in `api/data-types.md` and the `AllowPartialMatch` row in `api/command-definitions.md`.

Also documented: multi-slot fills accumulate across utterances (`command-recognition.md`), and the new session-log entry (`scoring.md`).

## Why

#73 established that a command missing a required argument must not fire, and routed such commands **to** the slot-fill path. That made this path the designated destination for exactly the commands #73 refuses — and it did not re-check them on the way out. `TryFollowUpSlotFill` breaks at the first slot it cannot fill and returns as soon as one new slot is filled, so with two or more unfilled required slots an utterance answering only the first produced a command `IsIncomplete` would reject, which Step 5 then fired. `ScoreFollowUp` does not stand in for the test: it re-scores against the matched pattern, so a partly filled command can score alongside a complete one.

Keeping the pending alive rather than discarding the fill is what makes this a refusal to fire rather than a refusal to progress — answering one slot per utterance now works.

## Two calls worth your ruling

1. **Cases 2 and 3 went the "deliberate → document" way, not the "add the condition" way.** No behaviour changed. The issue leans this way (`"fire as-is" plainly names the intent`) and documenting is the zero-risk branch, but it is a product ruling and it is yours, not mine. If you want them refused instead, say so and I will file it.
2. **`CreatedTime` carries over unchanged on a partial fill**, so the whole multi-utterance exchange runs against the single `pendingTimeout` window that started when the command first entered pending — a fill does not extend it. That matches the one existing precedent (`Complete` preserves `CreatedTime` when it re-enters for confirmation). The alternative — progress refreshes the deadline — is arguably friendlier at the default 5 s, but it changes an established timing rule and would belong on the confirmation re-entry too, so I kept it out of this fix. An unanswered remainder now reaches the timeout, where the default `Cancel` discards it; `Timeout_DefaultCancel_OnAPartlyFilledPending_Cancels` pins that.

## Validation

- [x] Unity **EditMode 118/118** and **PlayMode 414/414** green (`failed="0"`, 0 skipped) against the working tree, via host project `VoXR TestGround` on Unity 6000.4.7f1. Both platforms run because `Tests/Runtime` is PlayMode — which is where all four new tests live.
- [x] Four new tests in `VoxrPendingCommandTests`, all built on the **two-required-slot pending** the issue named as untested — the shape that makes case 1 reachable at all, and which nothing else in the suite constructs:
  - `FollowUp_FillingOnlyOneOfTwoRequiredSlots_StaysPending` — case 1. Asserts the command does not fire, is not cancelled, **and** that the fill was kept (without that last assertion the test would pass equally on a slot-fill that matched nothing), then that a third utterance completes it with both slots attached.
  - `ConfirmVocabulary_OnAPartialMatchPending_FiresAsIs_ByDesign` — case 2, pinning the ruling.
  - `TimeoutFireAsIs_OnATwoSlotPartialMatchPending_FiresAsIs_ByDesign` — case 3, pinning the ruling.
  - `Timeout_DefaultCancel_OnAPartlyFilledPending_Cancels` — the other side of the fix.
- [x] No stray `.meta` orphans or manifest drift left behind by the test run; working tree reverted and verified clean.
- Not run: `NativeBridge~/` is untouched, so the bridge build and desktop harness were skipped.
- No on-device check needed: the change is pure command-recognition logic with no mic capture, no MonoBehaviour lifecycle hook, and no native surface. The timeout path it touches is exercised in PlayMode.


jinwoo1601 (commented):
## Independent review — slim profile (issue lane)

**Head reviewed:** `83dac336030ff73460bb8437de76b79edb7131c7` · **Base:** merge-base `3e9d0ff`
**Angles run:** `CLAIM`, `LINE`, `GONE`, `XFILE`, `PIT`, `GAP` — the full slim profile, none dropped.
**Method:** six finder agents against the diff with no context from the session that wrote it, deduped, then one adversarial verifier per surviving finding.

**Result: 9 CONFIRMED, 1 REFUTED. No blockers.** The core fix is sound — `GONE` and `LINE` independently established that the completeness branch is correct, that termination is guaranteed (the unfilled set strictly shrinks per fill), and that `ComputeUnfilledSlots` (captured definition) and `IsIncomplete` (live definition) cannot diverge while a pending is live, because every set-mutating entry point cancels the pending first. `GAP` independently re-derived the fixture score as `3/7 ≈ 0.43` and confirmed the two behavioural tests fail if the fix is reverted — they witness their claim. `GONE` also found that this PR **improves** an untested path: at base, a `RequiresConfirmation` + `AllowPartialMatch` command with two unfilled slots re-entered confirmation with `UnfilledSlots = Array.Empty<string>()`, permanently disabling further slot-fill; head keeps the remaining slots fillable.

Every finding below is a consequence of this PR, not inherited debt, unless stated.

---

### 1. Step 5's editor diagnostic reads `_pending.Current.Value` *after* firing `OnCommandPending` — a subscriber that resolves the pending makes it throw
**CONFIRMED · minor · high confidence** · `Runtime/Commands/VoxrCommandRecogniser.cs:686`
*Reported independently by four of the six angles (`PIT`, `LINE`, `XFILE`, `GONE`).*

```csharp
InterpretResolution( followUpIncomplete ? _pending.AdvanceSlotFill(...) : _pending.Complete(...) );
#if UNITY_EDITOR
string followUpReason = followUpIncomplete
    ? "still pending (partial: unfilled "
        + $"[{string.Join(", ", _pending.Current.Value.UnfilledSlots)}])"
```

`InterpretResolution` on `ReEnteredPending` synchronously invokes the public `OnCommandPending` (`:905-906`) before line 686 runs. A subscriber may call `CancelPendingCommand()` (`:275`), `Configure(...)` (`:187`, `:217`), `SetActiveSets(...)` (`:234`), or set `enabled = false` — Unity dispatches `OnDisable` synchronously, and it runs `InterpretResolution(_pending.Cancel())` (`:460`). All route to `Cancel()`, which nulls `_pendingCommand` (`PendingCommandHandler.cs:241-247`). `Current` is `VoxrPendingCommand?`, so `.Value` then throws `InvalidOperationException`. There is no `catch` anywhere on the path from `DispatchFinalResult` down, and no re-entrancy guard.

This is the only unguarded `_pending.Current` deref in the file — `:472-473` and `:950-951` both guard on `HasPending` — and the only diagnostic block in `ProcessParsedResults` that reads handler state *after* the events rather than capturing a local *before* them. Step 1 uses `ccResolution.Command` captured at `:600`; Step 7 computes `var unfilled = ...` at `:756` before `InterpretResolution` at `:762-763`. The base revision had no such read. The scenario is exactly the handler shape this PR's own new documentation invites (`command-recognition.md:386`: "A prompt driven off `OnCommandPending` therefore sees each fill as it lands") — e.g. a prompt that cancels when its menu is closed.

Player builds are provably unaffected: every directive in the file is plain `#if UNITY_EDITOR`, never `|| DEVELOPMENT_BUILD`. That is what holds this at minor. Secondary consequence the verifier noted: the throw unwinds past `_buffer.ClearWords()` (`:555`), so the editor also keeps stale words in the utterance buffer.

**Fix:** hoist the read above the event — capture the unfilled list (or the whole reason string) after `AdvanceSlotFill` re-arms the pending but before `InterpretResolution`, matching the capture-before-fire pattern its Step 1 and Step 7 siblings already use.

---

### 2. The new documentation overstates the ruling: `FireAsIs` alone cannot fire an incomplete command
**CONFIRMED · minor · high confidence** · `Documentation~/command-recognition.md:433`, `Documentation~/api/data-types.md:89`

The new text says opting into *either* exception obliges the handler: "**Opt into either** and the handler for that command **must** tolerate every required slot being absent", and `data-types.md`: "Handlers for a command whose recogniser uses it must tolerate every required slot being absent."

The two opt-ins are not independent. The verifier enumerated every write to `_pendingCommand`: `EnterPending` site A (`VoxrCommandRecogniser.cs:759`) is gated on `partialDef.AllowPartialMatch` (`:754`); site B (`:815`) passes `Array.Empty<string>()` and is reachable only past the completeness gate at `:749-750`, so its command is complete; `Complete`'s confirmation re-entry sets `UnfilledSlots` empty; and `AdvanceSlotFill` requires a non-empty `UnfilledSlots` by induction back to site A. `AllowPartialMatch` is read in exactly one place in the entire runtime. So a pending holding an incomplete command **requires `allowPartialMatch`**, and `pendingTimeoutBehavior = FireAsIs` on its own only ever fires complete commands.

The error is fail-safe — it makes developers *more* defensive than needed, never less — which is why it stays at minor. But it is a false statement in an API reference. Note the PR's own test is named `TimeoutFireAsIs_OnATwoSlotPartialMatchPending_...` and configures `allowPartial: true`: the test names the precondition the prose drops. Same inaccuracy mirrored at `api/command-definitions.md:17`, which credits `AllowPartialMatch` with only the confirm-phrase consequence and omits that it equally gates the `FireAsIs` exposure.

---

### 3. `OnCommandPending`'s API-reference row is now wrong
**CONFIRMED · minor · high confidence** · `Documentation~/api/command-recogniser.md:43`

The row reads "Fired when a command **enters** pending state (partial match with unfilled required slots, or awaiting explicit confirmation)". After this PR the event also re-fires on a command already pending, once per partial fill — N times for N slots.

The verifier checked whether this was inherited and concluded it was **not**: at base, the only `ReEntered` producer was `Complete`, which set `Reason = AwaitingConfirmation` — an actual transition into one of the two states the row enumerates. `AdvanceSlotFill` returns `ReEntered` while `Reason` stays `PartialMatch` and `CreatedTime` is preserved, so the row's enumeration is false for N−1 of N fires. The base row was accurate; this PR made it wrong. The event's declaration (`VoxrCommandRecogniser.cs:122`) carries no XML doc comment, and the re-fire is stated in exactly one place in `Documentation~/` — the guide paragraph this PR added. Partly mitigated by the row's own `See [Pending Commands]` link, which does reach that paragraph.

---

### 4. The ruling is not reachable from the ScriptableObject page that documents the same flag
**CONFIRMED · minor · medium-to-high confidence** · `Documentation~/api/scriptable-objects.md:45`, `Documentation~/api/command-recogniser.md:24`

Issue #77 asked: "If yes, document it on `allowPartialMatch` and `pendingTimeoutBehavior`." `api/command-definitions.md:17` got the caveat and cross-reference; `api/scriptable-objects.md:45` documents the **same flag** for Inspector authoring and was left untouched. At merge-base the two rows were semantically equivalent, so the divergence is PR-introduced. Reachability there is genuinely dead-ended: `scriptable-objects.md` contains no link to `command-recognition.md` at all, and `Documentation~/inspector-authoring.md` never mentions partial match — a zero-code author has no path to the ruling.

The `pendingTimeoutBehavior` half is weaker and the verifier said so: `### Timeout Behaviour` and `### The two ways an incomplete command still fires` are adjacent headings in the guide, so the ruling is one heading away. What survives is that `api/command-recogniser.md:24` deviates from its own table's convention — six sibling rows each deep-link to a guide section. **If this is fixed, the text must state the conjunction** (`FireAsIs` *and* a command using `allowPartialMatch`), not clone the sentence finding 2 challenges.

---

### 5. The `OnCommandPending` re-fire payload is never asserted — only its intent
**CONFIRMED · minor · high confidence** · `Tests~/Runtime/VoxrPendingCommandTests.cs:966`, `:1000-1004`

The sole subscriber is `cmd => pendingIntents.Add(cmd.Intent)`; the sole assertion is on `pendingIntents.Count`. Every slot assertion reads `_recogniser.PendingCommand` — handler state, a separate surface — not the event argument.

Mutation test, run by the verifier: change `AdvanceSlotFill` to return `ReEntered(pending.Command)` (the stale pre-fill command) while leaving the stored `Command = partiallyFilled` intact. **Every assertion in all four new tests still passes**, as do all pre-existing tests, because `AdvanceSlotFill` is unreachable without a two-unfilled-slot pending and nothing else constructs one. Yet that one expression is the whole of the contract published at `command-recognition.md:386` — "`OnCommandPending` fired again **carrying the updated command**". The test's assertion message claims a capability its assertion does not test, which is precisely the #76 failure mode.

**Closing assertion:** capture the payloads and assert `pendingPayloads[1].GetSlot("weapon") == "missiles"` — the only assertion in the suite that fails under the mutation.

---

### 6. `Reason` carry-over is untested, and two-unfilled-slots × `RequiresConfirmation` has no coverage at all
**CONFIRMED · minor · high confidence** · `Runtime/Commands/PendingCommandHandler.cs:199`

`AdvanceSlotFill` writes `Reason = pending.Reason`, load-bearing at `Complete`'s guard `RequiresConfirmation && Reason == PartialMatch` (`:213-214`). `ConfigureTwoSlotPartial` sets only `allowPartialMatch: true`, so that guard always short-circuits on its first conjunct in every test that reaches `AdvanceSlotFill`. The pre-existing `Combined_PartialThenFollowUpThenConfirm` strands exactly **one** slot (verified by derivation: `"target"` is not a value of slot `target`, so it consumes the literal), so it never enters `AdvanceSlotFill`.

Mutation: write `Reason = AwaitingConfirmation` instead. **Whole suite stays green**, while a `requiresConfirmation` command would fire with its confirmation gate skipped. Severity is capped at minor because the enum's zero value is the safe one (`PartialMatch = 0`), so only an explicitly wrong assignment breaks it. There is also a second, cosmetic consumer nobody tests: `Editor/VoxrDebugWindow.cs:336`.

**Closing test:** the `ConfigureTwoSlotPartial` fixture with `requiresConfirmation: true` — assert the *completing* fill does not fire and that only a subsequent confirm phrase does.

---

### 7. The `CreatedTime` carry-over ships a user-facing promise that no test pins
**CONFIRMED · minor · high confidence** · `Runtime/Commands/PendingCommandHandler.cs:200`

`TestForceTimeoutNow` overwrites `current.CreatedTime = -1000f` (`VoxrCommandRecogniser.cs:952`) before triggering the only reader (`:473`), so every timeout test is blind to whether `AdvanceSlotFill` preserved or refreshed the field. Mutation to `CreatedTime = Time.time`: **zero tests fail.** `CreatedTime` appears in `Tests~/` exactly once — in a comment.

The PR frames this as an open question for the human, which lowers urgency but does not close the gap: the doc sentence at `command-recognition.md:386` ("filling a slot does not extend it") is new in this PR and merges either way. The verifier confirmed it is cheaply closable — `EditorPendingCommand` exposes the record over IVT, and a `[UnityTest]` with `yield return null` between injections distinguishes preserve from refresh deterministically. Fair caveat: `Complete`'s re-entry has carried `CreatedTime` untested since before this PR, so the convention, not just this line, is unpinned.

---

### 8. `scoring.md`'s short-circuit table has no key cell for the new reason string
**CONFIRMED · nit · high confidence** · `Documentation~/scoring.md:439`

The table's first column is `rejectReason`. `still pending (partial: unfilled [...])` is a real non-empty reason string, but it appears only inside the body of the row keyed `*(empty, `accepted: true`)*` — a reader looking it up will not find it. The folding idiom in the sibling rows is real, but the verifier found the direction matters: at base, every folded sibling had an **empty** `rejectReason` (both `:614` and `:487` discard their computed label on the accepted branch), so before this PR every non-empty reason string in the table was also a key. This is the first that is not.
**Fix:** widen the row key, e.g. ``| *(empty, `accepted: true`)* — or `still pending (partial: unfilled [...])` |``.
*Half of this finding was refuted:* "Four paths" remains accurate — this is a second outcome of the existing follow-up path, not a fifth path.

---

### 9. The new diagnostic string is doc-published but unpinned
**CONFIRMED · nit · high confidence** · `Runtime/Commands/VoxrCommandRecogniser.cs:685`

`scoring.md:439` publishes the wording and the `accepted: false` flag as a contract; the only `.cs` occurrence in the repo is the producer. The reporting angle's own caveat is accurate — `cancelled via vocabulary` and `timeout — cancelled` are equally unpinned at base, so this extends an existing gap rather than creating one. It survives as a finding because this repo has explicit recent canon the other way: `Tests~/Runtime/VoxrCommandParserTests.cs:2121-2128` — "*These tests exist so the page fails a test rather than a reader*" (#83).

---

- **Fixture comment justifies the routing on the wrong condition** (`VoxrPendingCommandTests.cs:914-916`). **REFUTED.** Read in full, the comment names incompleteness three times and leads with it ("Every test below needs a pending with TWO unfilled required slots, which is what makes the follow-up exit's gap reachable at all"). Every factual assertion in it is true — independently re-derived as `3/7 ≈ 0.4286` below `minScore 0.6`, admitted by DR-7. True-but-not-exhaustive is materially weaker than wrong, and the #76 analogy does not transfer: the asserted behaviour is the Step 5 follow-up exit, which has no score gate at all.
- **PR body says "cross-referenced from the `FireAsIs` row"** when the addition is a paragraph below the table. Checked; a fair paraphrase, not chargeable.

---

- **The `118/118` / `414/414` test counts.** This is a bare UPM package — `Tests~` cannot run standalone — so the reported run came from the host project and was not re-executed here. The `CLAIM` angle confirmed the named environment is capable of the check; it did not re-run it. What *was* verified by reading is that the tests exist, are wired correctly, and that two of the four fail if the fix is reverted.
- **On-device (Quest) behaviour.** Nothing in this diff touches mic capture, a lifecycle hook, or native code — asserted from the diff, not from the environment.
- **`NativeBridge~/`** is untouched; no build was run and none was required.

No blocker surfaced, so per the issue lane nothing here was fixed on the branch — every finding is posted unruled for the human. Not approving and not merging.

jinwoo1601:
Pushed `3d5cf6c`, which addresses **all nine CONFIRMED findings** from the review above. The one REFUTED finding was left standing as refuted.

| # | Finding | What the push does |
|---|---|---|
| 1 | Step 5 reads `_pending.Current.Value` after firing `OnCommandPending` | Read hoisted above `InterpretResolution` (`VoxrCommandRecogniser.cs`), matching the capture-before-fire pattern Step 1 and Step 7 already use |
| 2 | Docs overstate: `FireAsIs` alone cannot fire an incomplete command | Corrected in `command-recognition.md`, `api/data-types.md`, `api/command-definitions.md` — **and in the CHANGELOG entry**, which carried the same error and the review did not catch |
| 3 | Ruling unreachable from the pages the issue named | Caveat + cross-reference added to `api/scriptable-objects.md` (`allowPartialMatch`) and `api/command-recogniser.md` (`pendingTimeoutBehavior`). Per the verifier's warning, both state the **conjunction** rather than cloning the sentence finding 2 challenges |
| 4 | `OnCommandPending` row says "enters pending state" | Row now records that one pending command raises the event once per fill, carrying the updated command |
| 5 | Re-fire payload never asserted | The test records `List<VoxrCommand>` payloads instead of intent strings and asserts `pendingPayloads[1].GetSlot("weapon") == "missiles"` — the assertion that fails under the stale-command mutation |
| 6 | `Reason` carry-over untested | New `FollowUp_AcrossTwoUtterances_StillReachesTheConfirmationGate`: two-slot pending with `requiresConfirmation`, asserting the *completing* fill does not fire and only the confirm phrase does. It also pins the regression the review found this PR **repairs** — pre-#77, `Complete` emptied `UnfilledSlots` on the first fill and stranded the second slot for good |
| 7 | `CreatedTime` promise unpinned | New `[UnityTest]` reading `EditorPendingCommand.CreatedTime` across a frame. It carries an `Assert.Greater(Time.time, createdOnEntry)` precondition, so it cannot pass vacuously if the clock fails to advance |
| 8 | New session-log string unpinned | New EditMode test `FollowUpPartialFill_LogsStillPending_NotAccepted`, asserting the reason string, `IsAccepted == false`, and that the reason names the slot still **outstanding** rather than the one just filled |
| 9 | `scoring.md` row has no key for the new string | Row key widened to `*(empty, accepted: true)* — or still pending (partial: unfilled [...])`. The "Four paths" prose is untouched, since that half of the finding was refuted |

Also sharpened the `ConfigureTwoSlotPartial` comment — not because the refuted finding stood, but because the verifier's reasoning (the score is neither necessary nor sufficient for the routing) is worth recording against a future reader "fixing" the fixture by adjusting its score.

**Re-verified:** EditMode **119/119**, PlayMode **416/416**, `failed="0"`, 0 skipped — `+1` and `+2` against the previous run, matching the three new tests exactly, each confirmed present as `runstate="Runnable"`. `grep -c "error CS"` is 0 in both logs, so the restructured double `#if UNITY_EDITOR` block compiles under both define sets.

Still not verifiable from here, unchanged from the review above: on-device Quest behaviour (nothing in the diff touches it) and `NativeBridge~/` (untouched, no build run).

jinwoo1601:
Pushed `487f672` — **the `CreatedTime` question raised in the PR body is now settled the other way, on the author's ruling.** The PR description's "Two calls worth your ruling" §2 is superseded by this comment.

### What changed

`pendingTimeout` measured total elapsed time since the command first entered pending. Keeping a pending alive across utterances — the whole point of the fix — put that under load it never carried before: the default 5 s now had to cover *every* utterance of a multi-slot exchange plus a `bufferWindow` each, so a speaker answering one slot at a time could be cut off mid-answer and lose the command. The documentation this branch adds advertises exactly that workflow, which is what made the old rule untenable rather than merely debatable.

`AdvanceSlotFill` and `Complete` now take the current time and stamp it onto `CreatedTime`, so **each fill buys another window**. What still bounds a stalled pending is silence — the first window nobody answers ends it. Time is passed in rather than read, keeping `PendingCommandHandler` free of `UnityEngine`, as `EnterPending` already established.

The **confirmation re-entry restarts on the same rule**: a command that reaches its confirmation stage by being filled slot-by-slot should not arrive with less time to confirm than one that was complete when first heard. That half touches behaviour that predates this issue, so it is the most reviewable thing in the push — it is a one-line revert if you want the restart confined to the fill path.

### Coverage

- `FollowUp_PartialFill_RestartsThePendingTimeoutWindow` — the previous commit's pin, inverted from `AreEqual` to `Greater` and renamed. Old name is absent from the result XML, so no stale duplicate survived.
- `CompletingFill_RestartsTheWindowForConfirmation` — new, covering the `Complete` re-entry.

Both keep an `Assert.Greater(Time.time, …)` precondition before the real assertion, so neither can pass vacuously if the frame fails to advance the clock.

### Verification

EditMode **119/119**, PlayMode **417/417**, `failed="0"`, 0 skipped. `error CS` count 0 in both logs. Delta is EditMode +0 / PlayMode +1, matching the single new test exactly.

The five existing timeout tests were flagged as plausible casualties of the semantic change and all pass unchanged — they drive the clock through `TestForceTimeoutNow`, which overwrites `CreatedTime` outright and is therefore blind to which rule is in force. That is the same blindness the review's finding 7 identified; it is why these two `[UnityTest]`s read the field directly instead.

Docs and CHANGELOG re-stated: `pendingTimeout` bounds silence, not the exchange. Grepped `Documentation~/`, `CHANGELOG.md`, `Runtime/` and `Tests~/` for the old wording — no stale copies remain.

PR #87, opened 2026-08-15
Forward coverageWeight from the batch test window (#80)

### LINKED ISSUE #80: VoxrBatchTestWindow never forwards coverageWeight, so batch results can name a different intent than runtime fires
Split out of the #78 review, where it was verified and ruled out of scope for that branch (it pre-dates the change).

`Editor/VoxrBatchTestWindow.cs:327` and `:335` construct `VoxrBatchTestRunner` with `minScore` and `minConfidence` only. The window mirrors those two as serialized fields but has no `coverageWeight` mirror, so the runner's parser always uses the default `1.0` regardless of what the project's `VoxrCommandRecogniser` is set to.

`Documentation~/editor-testing.md` tells authors to pass their tuned value "so batch results predict runtime behaviour" — an instruction the first-party window does not follow.

**Verified against the base revision**, because my first framing of this was wrong. I claimed #65 §5.2 escalated it "in kind" — that before, a mismatched weight only moved the score, and now it can change the intent. That is false: `VoxrBatchTestRunner.Run` accepts the first command clearing the thresholds and breaks, so a post-selection charge that pushed round 1 under `minScore` already promoted round 2's different intent. Compiling the base parser reproduces it:

```
BASE w=1 (harness default): [report(0.500) launch(1.000)] -> batch reports "launch"
BASE w=0 (project runtime): [report(1.000) launch(1.000)] -> runtime fires "report"
```

So this is a pre-existing harness/runtime fidelity gap that #65 widens, not one it creates.

**Fix shape:** a third serialized threshold field on the window, forwarded to both constructor calls, matching how `minScore`/`minConfidence` are already mirrored. Note it adds a `[SerializeField]`, which is why it was kept off the feature branch (invariant 13 forbade new serialized surface there).

jinwoo1601:
Closes #80

## What

`VoxrBatchTestWindow` gains a third serialized threshold, `coverageWeight`, mirroring how `minScore` and `minConfidence` are already mirrored, and forwards it to **both** `VoxrBatchTestRunner` constructor calls in `CreateRunner()`. It is drawn beside the other two as **Coverage Weight** and defaults to `VoxrCommandParser.DefaultCoverageWeight` — read from the parser rather than hard-coded, so it cannot drift from the recogniser's own default.

Also: a new EditMode suite pinning the forwarding; the one doc site that recorded the gap (`Documentation~/editor-testing.md`'s "the editor window does not forward this value" caveat); and a new UI paragraph where that page previously said nothing about the thresholds at all.

A follow-up commit on this branch corrects a misquote the review of this PR turned up: `KNOWN_LIMITATIONS.md` attributed "predict runtime behaviour" to `editor-testing.md`, which says "track". The misquote predates this change; the new UI paragraph had picked "predict" up from it, so the page now uses one phrase for one idea. Documentation only — no code, so the suites above still stand.

## Why

The window mirrored two of the runner's three thresholds and passed only those, so the batch parser always ran at the default `1.0` however the project's `VoxrCommandRecogniser` was tuned — while `Documentation~/editor-testing.md` told authors to pass their tuned value "so batch results track runtime behaviour", advice the first-party window did not follow.

The cost is not confined to a moved score. `VoxrBatchTestRunner.Run` accepts the first candidate clearing the thresholds and breaks, so a weight the window failed to apply can hold the leading candidate under `minScore` and let the intent behind it through — the batch then names a command the runtime does not fire. Per the issue, that was verified against the base revision: it predates #65 (a post-selection charge already promoted round 2's different intent) and #65 widens it rather than creating it.

## Validation

- [x] **EditMode + PlayMode suites green, Unity 6000.4.7f1** via the `VoXR TestGround` host project, per the project's verification bindings: EditMode **124/124**, PlayMode **417/417**, 0 failed / 0 skipped / 0 inconclusive, no compile errors. The 5 new cases are the EditMode delta and all pass.
- [x] **The two scores the new tests assert were measured against the real parser**, not derived by hand: `VoxrCommandParser` staged from this worktree and compiled under `dotnet` (the `Planning~` A/B rig). On a grammar registering only `cease fire`, "cease fire please now" scores **0.500** at `coverageWeight` 1.0 and **1.000** at 0 — so the default-weight window refuses the command (`minScore` 0.6) and a window forwarding 0 fires it. That difference is the test's witness.
- [x] Both constructor branches covered — the window picks between them on whether `activeSetNames` is populated, and each needed forwarding separately; the new tests are `[TestCase]`-parameterised over the two.
- No on-device (Quest) check needed: editor-only window code, no runtime, native-bridge, or MonoBehaviour-lifecycle surface.

## Notes for review

- The new field adds serialized surface to an `EditorWindow`, which is why the issue kept it off the #78 feature branch (invariant 13 forbade new serialized surface there). Nothing else in the package serializes this window.
- The tests reach the window's private fields and `CreateRunner` by reflection deliberately — testing the forwarding is not a reason to widen the window's own surface.


jinwoo1601 (commented):
## Review — PR #87

**Depth:** slim profile (issue lane) — `CLAIM`, `LINE`, `GONE` — **plus two added angles**: `XFILE`, because the diff's correctness rests on contracts that fail at run time rather than compile time (positional constructor arguments, an `internal const` across an `InternalsVisibleTo` seam, five reflection string literals naming private members in another assembly), and `PIT`, because the new test instantiates and destroys a real `EditorWindow` and the diff adds a `[SerializeField]` to a type Unity persists.

**Head reviewed:** `8da067f7fd8258c80cabb8f60d6585b1a41048f5` · **merge-base:** `2594bf47dce5758a4c3706bff51c1b94b2c6e587`

Five finder angles, three surviving findings, each adversarially verified: **two CONFIRMED (both nits, both confined to the PR narrative), one REFUTED**. `LINE`, `XFILE` and `PIT` returned clean. **Nothing in the merged result is wrong** — the code, the CHANGELOG entry, and the doc edits all hold up.

---

### 1. PR body and issue #80 misquote the doc — `nit` · CONFIRMED · high

The body says the doc *"told authors to pass their tuned value `"so batch results predict runtime behaviour"`"*. The sentence being quoted is unchanged on both sides of the diff and reads **track**, not predict — base `Documentation~/editor-testing.md:153`, head `:155`. `grep predict` over that file at base returns nothing.

The counter-argument was hunted and rejected: this diff *does* add a "predict" sentence at `Documentation~/editor-testing.md:140`, but it is added here, so it cannot be what the doc "told authors" in the body's past tense. Issue #80 carries the same quoted phrase, so the PR inherited it rather than inventing it.

Merged content is clean: `CHANGELOG.md:41` quotes it correctly as "track", and the doc's own sentence is untouched.

### 2. "the two doc sites that recorded the gap" — only one recorded it — `nit` · CONFIRMED · high

The body's *"Also: a new EditMode suite pinning the forwarding, and the two doc sites that recorded the gap."* A repo-wide grep at the merge-base returns exactly **one** site stating the gap: `Documentation~/editor-testing.md:153` — *"and the editor window does not forward this value, so it always runs at the default"*.

The diff's second doc hunk adds a new paragraph to the Visual UI section, where the base said nothing about thresholds at all (base `:136-139` covers only Run All, the results table, and CSV export). One hunk closed a recorded gap; the other filled a silence. The verifier also killed the charitable reading — both hunks are in the *same file*, so "two doc sites" cannot mean two documents either.

Nothing merged is affected; head `:140` and `:155` are both accurate post-fix.

---

### Refuted

- **The new `Coverage Weight` field is unclamped, so a typed `-1`/NaN/∞ silently becomes "coverage off"** — mechanically true (`VoxrCommandParser.cs:233-234`, NaN takes the `else` arm) but not a defect: that mapping is published canon (`Documentation~/scoring.md:151` — *"Negative, NaN, or infinity | Treated as `0`"*), the mirrored recogniser field is unclamped identically (`VoxrCommandRecogniser.cs:63`, forwarded raw at `:200`/`:304`), and — decisively — clamping at the window would **create** the bug this PR fixes: window `-1` → coverage on while runtime `-1` → coverage off is exactly the harness/runtime divergence #80 exists to close. Fidelity is currently exact for every input, including garbage.

### Notes (unverified, no action implied)

- `KNOWN_LIMITATIONS.md:367` already attributes *"predict runtime behaviour"* to `editor-testing.md` and has since before this PR — the origin of finding 1, pre-existing and outside this diff. Issue candidate.
- `Documentation~/command-recognition.md:143` still says the Batch Test Runner takes "both" thresholds as constructor arguments; there have been three since #65. Pre-existing.
- The new suite pins the field→constructor half of the path by reflection but never drives `OnGUI`, so the GUI→field half is untested.
- The `#80` CHANGELOG entry is a single ~2000-character paragraph. Consistent with the surrounding entries; flagged only for the record.

- **The Unity run itself.** This is a bare UPM package — nothing compiles standalone. The PR's EditMode 124/124 + PlayMode 417/417 claim was produced against the `VoXR TestGround` host project and could not be re-executed by the finder agents; they audited it for consistency instead, and it reconciles: counting attributes across `Tests~/Editor` gives exactly 124 (5 of them new), and `Tests~/Runtime` gives 402 declared with `Replay_ManifestCase` expanding 1→16 for exactly 417. The tests also genuinely fail at the merge-base, so they witness the change rather than passing either way.
- **Unity's deserialization of a newly added `[SerializeField]` into an already-saved window layout.** Reasoned, not executed: both persistence paths are name-matched transfers onto a freshly constructed object, so an existing window should come back at the field initialiser (`1.0`) rather than `0` — the damaging outcome. Platform contract; not testable from here.
- On-device (Quest) behaviour — not applicable to this diff (no `Runtime/`, no `NativeBridge~/`), and human-only in this project regardless.

*Both confirmed findings are below blocker, so per the issue lane they are left unfixed and unruled for you.*

jinwoo1601:
Pushed `ddebbf0` and corrected the PR body, addressing both confirmed findings from the review above.

- **Finding 1** (misquote): the PR body now quotes `editor-testing.md` as "track runtime behaviour", which is what it says. The review's accompanying note is fixed in the tree rather than left for later — `KNOWN_LIMITATIONS.md:367` no longer attributes "predict runtime behaviour" to that page, and the UI paragraph this PR added, which had picked "predict" up from it, now says "track" too, so the page uses one phrase for one idea.
- **Finding 2** ("the two doc sites that recorded the gap"): the body now says one site recorded it and the other hunk added a UI paragraph where the page previously said nothing about the thresholds.

Documentation only, no code touched, so the EditMode 124/124 + PlayMode 417/417 run against `8da067f` still covers the merged result. Both findings were nits rather than blockers — fixed on your explicit ruling, not on my own.

PR #88, opened 2026-08-15
Terminate the orphan run where the matcher can begin, not where a pattern starts (#82)

### LINKED ISSUE #82: Orphan predicate asks where a pattern could BEGIN, but the matcher can begin anywhere by missing leading elements
The deferred half of design Amendment A3 (#65 §5.2), recorded at the G1 re-lock and confirmed by the #78 review.

`CanStartPattern` asks *"could any active pattern begin a match at this token?"* The matcher can begin a pattern at **any** token by missing its leading elements — which is the single most common VOSK failure and the reason this design exists. So the predicate and the matcher disagree about where a pattern starts.

A3 closed the half where that disagreement fired the **wrong command**. Two consequences remain, both measured over 699 utterances and both now in `KNOWN_LIMITATIONS.md`:

1. **A second command that lost its own leading word is charged to the first** (11 of 699 + 1 count-change). `"cease fire target hotel one"` → `cease_fire` falls `1.0` → `0.4` and stops firing, because `"target"` begins no pattern even though `approach target {target}` does in fact match there. Say the second command in full and both fire at `1.0`.
2. **A demoted winner can block a better later-starting match** (1 of 699). Selection ranks earliest start above score, so coverage cannot reorder across start indices. Sequential extraction recovers 28 of the 29 blocked candidates; the exception is when the winner's consumed span covers the start the better candidate needed.

**Fix shape considered and deferred:** a per-position admissibility probe — run the matcher at each token once per utterance and treat a position as a start if any pattern yields an admissible candidate there. Cost is one extraction round's work per utterance, done once rather than per candidate.

**What blocks the cheap version:** crudely widening the predicate (any token any pattern could match anywhere) collapses into the degenerate case — it makes `"hard burn"` a terminator and disables the feature grammar-wide, un-fixing #42.

**Note for whoever picks this up:** requirements F11 was narrowed during the #78 review specifically so it would not wrongly rule this out. It now forbids depending on *other candidates' verdicts*, which the probe would not do. The earlier wording banned reading any per-candidate state at all, which A3 already does legitimately.

jinwoo1601:
Closes #82

## What

The orphan run's terminator now asks the **matcher** instead of asking only what patterns start with.

`CanStartPattern` answers *"does a pattern's first matchable element match this token?"*. Selection asks nothing of the kind — it tries every pattern at every non-`[unk]` index — so a pattern whose leading elements the decoder dropped begins wherever its surviving elements do. The two disagreed about exactly the tokens that matter.

`BuildCoverageTables` now terminates the run at `IsAdmissibleStart`: a token is a start when some active pattern, **matched from that token**, ends up having matched **strictly more** of its required elements than it missed. `CanStartPattern` remains the cheap first test and is never overruled.

That threshold is deliberately one notch stronger than the admission rule selection applies. Admission asks whether something is a candidate at all — answered for a candidate that may lose its round and never fire. Terminating another command's orphan run moves score off a command that *is* firing, so it takes strictly more evidence for than against. **The first version of this PR used the weaker rule and it reverted #42 on the grammar the docs prescribe as #42's remedy; see the review and the follow-up comment.** Like admission itself this is a count, not a threshold — nothing to configure, independent of `minScore` and `coverageWeight`.

`TryMatchScored` gains a `forStartProbe` parameter. The probe is the one caller that runs *before* the coverage tables exist — it is what builds them — so it takes no coverage term. That costs it nothing: it reads only the score's sign and the two required-element counts, and coverage can only enlarge the denominator, so it can never flip a positive score negative.

## Why

`"cease fire"` + `"approach target hotel one"` heard as **`"cease fire target hotel one"`** charged `cease_fire` for all three trailing tokens, because `"target"` begins no pattern — even though `approach target {target}` *does* match there and the next extraction round *did* go on to fire it. `cease_fire` fell `1.0` → `0.4` and stopped firing, so the command spoken **cleanly** was the one lost while the damaged one fired.

This is the residue design Amendment A3 deferred (§5.2 / A3.5), and the issue names the fix shape: a per-position admissibility probe.

**Why this is not the "crude widening" the design refused.** Widening would let the slot value a bare pattern strands terminate its own orphan run, reverting the #42 protection grammar-wide. The strict margin refuses exactly those: `decelerate ?by {burn_level}` reaches `{burn_level}` from "hard" having missed one required element and matched one, which is not more evidence for than against. Measured — the #42 pair still inverts at `0.33`/`0.67`, the `?by` remedy still fires with its argument, one anchoring literal still tames a permissive slot, and the `["hard","stop"]` residue is unchanged.

**F11** (as narrowed after the #78 review) forbids a candidate's orphan count depending on *another candidate's verdict*. The probe does not: it runs after `CanStartPattern` and never overrules it, so no pattern's claim is ever withdrawn — only added — and its answer is a function of (grammar, tokens, index) alone, with no `searchStart` and no selection state.

## Validation

A/B rig per project memory `grammar-ab-rig`, validated first against the committed `grammar` pin in `NativeBridge~/harness/expectations.json` (byte-for-byte) before any delta below was trusted.

**699-utterance corpus, judged at the recogniser's gate (`minScore` 0.6), vs `origin/main`:**

| bucket | count |
|---|---|
| intent-change | 11 |
| count-change | 1 |
| **newly-firing** | **0** |
| **newly-silent** | **0** |
| **slots-lost** | **0** |
| **slots-gained** | **0** |
| score-only | 4 |

The 11 + 1 are A3.5's residue closing: in every one, the *first* command returns to `1.000` and the second was already firing at `0.667`. Nothing regressed in either direction.

Also measured:

- **Eager-commit verdicts: 0 of 699 moved** on finished utterances, and **0 of 4,014 moved** across every growing prefix of the corpus (`--prefix-verdicts`, added for this) — the prefixes being the buffer states `TryEagerCommit` actually runs on. The "eager verdict always names the command the flush will fire" invariant is intact; the 33 buffers whose flush column moved kept the same intent *and* pattern index.
- **The ~50 documentation numbers the local doc harness pins**, replayed through the real parser (`DocCheck`): **exactly one moves** — `scoring.md` worked example D, which was this defect written up as a limitation.
- **`architecture.md` §9.9's exhaustive start-index sweep, re-run:** `29 blocked / 28 recovered / 1 silent` → **`28 / 27 / 1`**. That limitation therefore keeps both its root cause and its measured case, and its `KNOWN_LIMITATIONS.md` entry is left as it was. The harness now lives beside the rig (`ab-rig/Sweep.cs`); it reproduces the documented baseline exactly on `main` first, with a self-check that its restatement of the selection key order names the same winner `Parse` fires on all 699 utterances (0 mismatches).
- **`GenerateGrammarJson` is byte-identical**, so the Tier C harness baseline is unaffected.
- **Cost:** `~12.3 → ~16.1 µs` per `Parse`, at **no added allocation** — `492.8` bytes/utterance on both sides, because the probe tells the number matcher to skip building the joined slot value it would only discard. Time figure is the flush path; `TryEagerCommit` builds the same tables, so the sweep also runs once per VOSK partial where `eagerFlushOnCompleteMatch` is enabled, and is skipped entirely at `coverageWeight` `0`.

Untouched: selection keys, the admission rule itself, `[unk]` handling, the Amendment A3 *rule*, `coverageWeight` and its clamping, and `CanStartPattern`. `_forcedOrphanRun` is computed from the table this PR redefines, so A3's charge shifts wherever the new terminator ends a run earlier — the rule is unchanged, its values are not.

- [x] Unity **EditMode** and **PlayMode** suites, both platforms (`Tests~` renamed into the `VoXR TestGround` host project, Unity 6000.4.7f1) — see the run summary in the PR comments.
- [x] 699-utterance A/B, eager-verdict diff, doc-number pin, start-index sweep, grammar pin, and the `--bench` cost figures above — all replayed through the **real parser sources** staged at `origin/main` and at this working tree.
- [ ] **On-device (Quest) check not needed:** no mic, lifecycle, or native-bridge code is touched — this is pure parser scoring, and it is covered by the PlayMode suite.

## Docs

- `Documentation~/scoring.md` — the orphan-test definition in §2, and worked example D rewritten from "the limitation" into the example of the rule working.
- `Documentation~/command-recognition.md` — the one-line summary of the same rule.
- `KNOWN_LIMITATIONS.md` — *"A second command that loses its own leading word is charged to the first"* deleted. *"A demoted winner can block a better-scoring match that starts later"* keeps its repro and root cause, with its sweep counts updated to `28 / 27 / 1`.
- `Documentation~/troubleshooting.md` — drops a pointer to worked example D, which no longer illustrates a limitation.
- `CHANGELOG.md` — under `[Unreleased] / Fixed`.

## Review

`review-pr` ran at the slim profile plus `XFILE` and `PIT`, and **found a blocker in the first version of this branch**: the probe admitted on DR-7's own rule, which reverted #42 on the grammar the docs prescribe as its remedy. Fixed in `49ae7a9`. The full review and the re-measurement are in the PR comments; everything below blocker is posted there unruled, for you.

Worth knowing when you weigh this: the 699-utterance corpus returned **zero regressions** on the broken build. It contains no bare-pattern/extension pair of the shape that exposes the defect, so a clean A/B on it is evidence about the demo grammar, not about the grammar space.

## For your ruling

The `#65` entry in the same `[Unreleased]` block still carries its own measurement (*"nothing newly fires … 17 stop clearing `minScore`"*), which was true of that change against the previous release but is no longer the net for the release as a whole. I left the measurement alone rather than rewrite another issue's entry, and added a cross-reference to the #82 entry for the release net — say the word if you would rather they were merged.

Design canon: `Planning~/design-docs/scoring-model.md` gains an **A4** section recording this as built, explicitly marked *recorded, not ratified* — §5.2 and A3.5 still carry the pre-#82 wording, and the re-lock is yours to make. That file is gitignored, so it is not in this diff.


jinwoo1601:
## Unity suite — both platforms green

Host project `VoXR TestGround`, Unity 6000.4.7f1, `Tests~` renamed in and the working tree under test (not a committed snapshot).

| platform | result |
|---|---|
| EditMode | **PASS — 124/124** (0 failed, 0 skipped, 0 inconclusive) |
| PlayMode | **PASS — 420/420** (0 failed, 0 skipped, 0 inconclusive) |

One existing test changed rather than broke, and it is worth naming explicitly: `Coverage_SequentialExtraction_ChargesTheFirstCommandForASecondThatLostItsWord` pinned the limitation this PR closes — by name, and by its own comment (*"pins the documented behaviour rather than asserting it is right"*). It lives in the block that pins the numbers `Documentation~/scoring.md` publishes, so it moved with worked example D: renamed to `...SparesTheFirstCommandWhenASecondLosesItsWord`, expected score 0.40 → 1.00. The first run of the suite failed on exactly that one assertion and nothing else, which is the cleanest evidence available that the blast radius is the intended one.

Three tests added for the new predicate: that it opens the position only reachable by missing leading elements, that it **refuses** the one the design refused (`decelerate` / `decelerate by {burn_level}` — the #42 anti-collapse guard, asserted end-to-end as well as at the table), and that it never overrules `CanStartPattern`, which is F11's property stated as a test.

jinwoo1601 (commented):
# Independent review — PR #88

**Profile:** issue lane → **slim** (`CLAIM`, `LINE`, `GONE`), **plus `XFILE` and `PIT`**. `XFILE` was added because the diff changes a method signature with a reflection-bound consumer and spreads one behavioural rule across four prose files; `PIT` because it adds a nested loop returning a struct by value and re-enters a method that writes shared per-instance scratch buffers. Both additions earned their place — `XFILE` found four findings and `PIT` confirmed the build-order guard is load-bearing.

**Head reviewed:** `f4aae46d772f85d4e41ddcf6de3956767b607577`, against merge-base `0029307`. Five finder agents, none of which inherited any context from the session that wrote the code.

**Headline: one CONFIRMED blocker.** The change as reviewed reverted the #42 protection it claims to preserve, on the grammar the docs prescribe as #42's remedy. It has been fixed on this branch — see the follow-up comment. Findings below are stated against the reviewed SHA.

---

## 1. BLOCKER — the admission threshold was DR-7's own, and DR-7 is too weak here: #42 reverts

`Runtime/Commands/VoxrCommandParser.cs:1664` · **CONFIRMED** (found independently by `LINE` and `GONE`, then reproduced against base)

The guard was `probe.MissedRequired <= probe.MatchedRequired`. `MissedRequired`/`MatchedRequired` count the **whole** pattern, so any pattern with as many required elements at-or-after the slot as before it passes. Take the pair `command-recognition.md` prescribes as the *safe remedy* — `["decelerate"]` + `["decelerate","?by","{burn_level}"]` — on **"decelerate hard burn please"**. Probed from "hard": one required miss (`decelerate`), one required match (`{burn_level}`) → admitted. The stranded value then terminates the **bare** pattern's own orphan run. Replayed through the real parser at both revisions:

```
base 0029307 :  decelerate_by:0.667[burn_level=hard burn]      <- argument kept
head f4aae46 :  decelerate:1.000[]                             <- argument discarded
```

That is #42, restored, at the default `coverageWeight`. `"decelerate hard burn now"` (required element after the slot) does the same: `0.750` with the slot → `1.000` without it. The new guard test only pinned the three-element `["decelerate","by","{burn_level}"]` form, which escapes by arithmetic (2 missed > 1 matched) — so the PR generalised an anti-collapse argument from the one instance where it happens to hold.

**Fixed** by requiring strictly more matched than missed (`MissedRequired < MatchedRequired`), which is deliberately one notch stronger than DR-7: admission decides whether something is a candidate at all, while terminating another command's orphan run moves score off a command that *is* firing. Two regression tests added in the shapes that bite.

## 2. MAJOR — the same rule silently broke the documented "anchor it behind a literal" workaround

`KNOWN_LIMITATIONS.md:385`, `Documentation~/scoring.md:122` · **CONFIRMED** (`GONE`)

The slot-initial limitation tells authors to "anchor them behind a literal". Under `<=`, `["heading","{heading}"]` probed from any digit gives 1 missed / 1 matched → admitted, so one anchoring literal bought nothing and the limitation widened from slot-initial to slot-second with no doc saying so. Measured: `OrphanedAfter` on the anchored grammar was `0` at head against `3` at base. Same root cause as #1; **fixed** by the same change, with a test pinning the workaround.

## 3. MAJOR — two shipped docs stated the same sweep with different numbers

`Documentation~/scoring.md:175` ↔ `KNOWN_LIMITATIONS.md:396` · **CONFIRMED** (`XFILE`)

The PR rewrote the `KNOWN_LIMITATIONS` entry to `27 / 27 / 0` but left `scoring.md` §5 saying "29 candidates were blocked … 28 were recovered — see Known Limitations for the one that was not", pointing at a file that then said there was none. **Fixed**: the corrected guard puts the real figure at **28 / 27 / 1**, the case still reproduces, and both pages now say so — the `KNOWN_LIMITATIONS` entry keeps its original repro rather than being rewritten.

## 4. MAJOR — the PR's own run summary overstated the blast radius

`Tests~/Runtime/VoxrCommandParserTests.cs:1194` · **CONFIRMED** (`CLAIM`)

The posted comment said the first Unity run "failed on exactly that one assertion and nothing else, which is the cleanest evidence available that the blast radius is the intended one." The diff also revised `OrphanRun_StopsAtATokenThatCouldBeginAPattern` (`OrphanedAfter(3)` 4→0, `OrphanedAfter(4)` 3→0) — edited *before* that run, so it never failed and the inference was not available. **Corrected** in the follow-up comment.

## 5–8. MINOR — stale cross-references this diff created · all **CONFIRMED** (`XFILE`), all **fixed**

| # | Where | What was wrong |
|---|---|---|
| 5 | `Documentation~/troubleshooting.md:73` | Cited worked example D and "the two shapes with no user-level workaround"; D had been inverted and one shape's `KNOWN_LIMITATIONS` entry deleted. |
| 6 | `CHANGELOG.md:22` | The #65 entry pointed at `KNOWN_LIMITATIONS.md` for the causes of its 11 intent-changes — the entry this diff removed. Now cross-references the #82 entry for the release net. |
| 7 | `Planning~/…/ab-rig/DocCheck.cs:463` | The standing doc-number pin still asserted `cease_fire:0.400`, so the next run would report a failure against a doc that is now correct. |
| 8 | `Planning~/…/phase7-startindex-sweep.tsv` | Still held the pre-#82 29 rows while the PR published re-swept numbers. |

## 9–13. MINOR — claim accuracy · all **CONFIRMED** (`CLAIM`), all **fixed in the CHANGELOG/PR text**

| # | Claim | Problem |
|---|---|---|
| 9 | "every other column is zero" | `report.py` prints seven columns; `score-only` was 18. Now says *behavioural* columns and states the score-only count. |
| 10 | "the Amendment A3 exception and `_forcedOrphanRun` … untouched" | True of the source line, false of the values: `_forcedOrphanRun[i] = 1 + _orphanRun[i+1]` reads the table this PR redefines. Now says the A3 *rule* is unchanged and its charge shifts where the run ends earlier. |
| 11 | "Cost … per `Parse`" + "the probe runs once per utterance" | `TryEagerCommit` also calls `BuildCoverageTables`, once per VOSK partial. Now scoped to the flush path with the eager path named. |
| 12 | "~50 documentation numbers … exactly one moves" | Scoped to what the local harness pins; a second published number (the sweep's silent case) sat outside it. Now says "pins", and the corrected guard leaves that number unmoved anyway. |
| 13 | The re-swept `28/27/1` figure | Was produced by an ad-hoc script with nothing in the repo to reproduce it. The sweep harness now lives beside the rig as `ab-rig/Sweep.cs`, self-checked so its restatement of the selection keys must name the same winner `Parse` fires on all 699 utterances (0 mismatches). |

## 14. MINOR — a hedge was deleted and replaced with a stronger claim than the code supports

`Documentation~/scoring.md:146` · **CONFIRMED** (`GONE`)

Base said the leading and trailing sides "usually" land in the same place; the PR replaced that with "It lands in the same place because the start test and the matcher ask the same question." They do not: the probe asks only whether a pattern *could* be matched from a token, while "a later round explains it" also requires that candidate to win its round and clear the gate. **Fixed** — the hedge is restored, now naming the residual direction (a score left higher than ideal, never a command charged for someone else's words).

---

### Refuted / checked clean

- **Build-before-read on the coverage tables** — no slot is read before it is written, and no stale value from a longer prior utterance is reachable. `PIT` and `LINE` independently confirmed the `forStartProbe` suppression is *load-bearing*: with `consumedEndIdx == idx` the probe would otherwise read `_orphanRun[i]` in a grown-and-never-cleared pooled array.
- **Shared scratch state** (`_matchSlotBuf` et al.) — the probe runs strictly before any reader, cannot interleave, and cannot overrun `_maxSlotsPerPattern`.
- **Struct copies, floating point, `#if UNITY_EDITOR` divergence** — all clean. `coverageWeight`'s constructor clamp keeps NaN off the probe path.
- **`[unk]` transparency, Amendment A3's `switch to X` flip, sequential extraction** — re-derived numerically and unmoved.
- **`TryMatchScored`'s new parameter** — the only reflection consumer was made signature-agnostic in the same change and works at both revisions.
- **"Start" really means "consumes `tokens[idx]`"** — the cursor advances only on a match, so a positive score implies the first matching element matched at `idx` itself.
- **The unticked Quest box** — accurate; nothing in the diff touches mic, lifecycle, or native-bridge code.

### Notes (unverified, not defects)

- The probe calls the full matcher, so a `NumberSequence` slot builds and discards a joined string per probe — the allocation `CanStartPattern`'s split with `CountNumberSequenceWords` exists to avoid. The `493 → 525` bytes figure is therefore grammar-specific and worse on numeric grammars.
- `IsAdmissibleStart` restates the admission comparison inline rather than sharing it with `IsBetterCandidate`; they are now deliberately different rules, which is worth a shared named helper if a third ever appears.
- The probe runs even when `coverageWeight == 0`, where the tables cannot affect any score.
- `_matchSlotCount` is written and never read anywhere in the repo (pre-existing).

### Not verifiable from here

- **On-device (Quest)** behaviour — not attempted, and not implicated by this diff.
- **The eager-verdict measurement** samples complete utterances, not the growing prefixes `TryEagerCommit` actually runs on. `PIT` gave a monotonicity argument that the untested direction is safe (admissibility only increases with buffer length, so eager score ≤ flush score, and the failure mode is `None` → wait, never an early commit the flush would not make) — that is an argument, not a measurement.
- **Unity suite counts** — reported separately; this review did not run them.

jinwoo1601:
Pushed `49ae7a9`, which addresses **finding 1 (blocker)** and, with it, findings 2–14 from the review above.

**The blocker:** the probe's admission test was DR-7's own `missed <= matched`. It is now `missed < matched`. DR-7 asks whether something is a candidate at all — answered for a candidate that may lose its round and never fire; terminating another command's orphan run moves score off a command that *is* firing, so it takes strictly more evidence for than against. Two regression tests added in the shapes that bite: the `?by` remedy pair, and the anchored permissive slot.

**Re-measured end to end at the corrected threshold**, vs `origin/main`:

| | before review | after fix |
|---|---|---|
| intent-change / count-change | 11 / 1 | **11 / 1** |
| newly-firing | 2 | **0** |
| newly-silent, slots-lost, slots-gained | 0 | **0** |
| score-only | 18 | **4** |
| eager verdicts moved | 0 of 699 | **0 of 699** |
| pinned doc numbers moved | 1 of 50 | **1 of 50** (worked example D) |
| start-index sweep | 27 / 27 / 0 | **28 / 27 / 1** |
| grammar pin | identical | **identical** |

The blast radius is *tighter* than what was reviewed, not wider — the fix removes movement rather than adding it. The one limitation entry the earlier version claimed to have emptied keeps both its root cause and its repro, so it is restored rather than rewritten.

**Corrections to my earlier claims, per findings 4 and 9–13:** the first Unity run did **not** fail on only one assertion — a second existing test had been revised before that run, so the blast-radius inference I drew from it was not available. The CHANGELOG's "every other column is zero" ignored `score-only`; "the A3 exception … untouched" was true of the line and false of the values; the cost figure covered only the flush path when `TryEagerCommit` runs the same sweep per partial; "~50 documentation numbers" meant the numbers the local harness pins. All corrected in the entry. The sweep harness that produced the `28 / 27 / 1` figure now lives beside the rig as `ab-rig/Sweep.cs`, self-checked against `Parse`'s actual winner on all 699 utterances.

**Worth recording:** the 699-utterance corpus returned *zero regressions* on the build that reverted #42. It has no bare-pattern/extension pair of the shape that exposes it. A clean A/B on this corpus is evidence about the demo grammar, not about the grammar space.

Unity re-run after the fix: **EditMode 124/124, PlayMode 422/422.** Everything below blocker in the review is left unruled for you.

jinwoo1601:
Pushed `04c0512` — every non-blocker from the review is now addressed. All four Notes plus the one measurement gap.

**Notes**

| Note | Action |
|---|---|
| Probe rebuilt the joined `NumberSequence` value it discards — the allocation `CanStartPattern` avoids by construction | `TryMatchNumberSequence` takes `valueNeeded`; the probe gets a named placeholder. Match/no-match and `consumed` are both decided before the string is touched, so nothing moves. **492.8 bytes/utterance on `main` and on this branch** — the regression is gone, not reduced, so the "grammar-specific" caveat is moot. |
| Probe ran at `coverageWeight == 0`, where no orphan count can reach a score | `IsAdmissibleStart` short-circuits. This makes `_orphanRun` weight-dependent — invisible to scoring, visible to a test reading `OrphanedAfter` — so a test pins **both** halves: the table differs (0 vs 4), the parse results do not. |
| Admission comparison restated inline rather than shared | Left as two rules, since they are deliberately one notch apart and there are only two — but `IsBetterCandidate` now points at the stricter sibling, so DR-7's reader meets it before "fixing" the inconsistency. A shared helper if a third appears, per the note. |
| `_matchSlotCount` written, never read | Removed. Flagging that it is **pre-existing**, not something this branch introduced — say the word and I will restore it to keep the diff tight. |

**The measurement gap (review's "not verifiable from here")**

`--verdicts` sampled only the finished line — the one buffer state `TryEagerCommit` never actually runs on. `ab-rig/Program.cs` gained `--prefix-verdicts`, which replays every growing prefix of the corpus:

> **0 of 4,014 prefix buffers change verdict.**

So the eager-gate claim is now measured rather than resting on the monotonicity argument. That argument was sound, but it was an argument.

**Everything above is behaviour-neutral** — the 699-utterance corpus output is byte-identical to the previous build: 11 intent-change + 1 count-change, every other behavioural column zero, sweep 28 / 27 / 1, grammar pin unchanged. `DocCheck` is now **50/50** (worked example D's pin updated to the behaviour it documents).

Unity: **EditMode 124/124, PlayMode 424/424**, including the two new tests.

Nothing from the review is left open. Still yours to rule on: the `#65` CHANGELOG entry's own measurement, cross-referenced rather than merged.

PR #89, opened 2026-08-15
Log the droppable-required-literal warning in the Editor only (#81)

### LINKED ISSUE #81: WarnOnDroppableRequiredLiteral fires at full breadth on grammars #65 §5.2 made correct
Split out of the #78 review, ruled out of scope for that branch.

The detector's contract changed under #65 §5.2 without its trigger changing. It used to mean *"this shape discards a spoken slot value"*; coverage now demotes the bare pattern, so for most grammars it means *"this shape could be authored to win by more"*. It still emits `Debug.LogWarning` on every parser construction — in runtime builds, and again on every `SetActiveSets` rebuild.

A warning that fires on grammars that now behave correctly is the kind that gets globally suppressed, taking the residual case with it.

**The residue is real** and is pinned by `Coverage_ResidualHazard_WhenTheStrandedValueBeginsAPattern`: when the stranded value's first word could itself begin a pattern, the orphan run terminates there, the bare form keeps its `1.0`, and the value strands exactly as before. So the detector should not simply be deleted.

**Options:**
- Narrow the trigger to that residue — `_startLiterals`/`_startSlots` can now answer whether the stranded value's first word begins a pattern. Would require making the detector an instance method; it is currently `static` and runs before those caches are read.
- Demote it to authoring guidance (editor-only, or once per `Configure` rather than per rebuild — see the existing 'Validation warnings re-emit on every active-set switch' limitation, which this compounds).
- Leave it and accept the noise.

Arguably a design question rather than a straight issue-lane fix, since it turns on what the detector is now *for*.

jinwoo1601:
Closes #81

## What

`WarnOnDroppableRequiredLiteral` gains `[System.Diagnostics.Conditional("UNITY_EDITOR")]`, matching `AssertCoverageTablesMatch`'s idiom in the same file. `Conditional` rather than `#if` so the call site and the four helpers below it stay one piece of code — they remain compiled but unreachable in a player build.

**The trigger stays at full breadth.** That is the half of this that is a decision rather than a cleanup, and it is the option the issue framed as a design question. The other two options — narrowing to the residue, or leaving it — were declined for the reasons below.

The only other C# edit is comment text — 19 of the 20 added lines in that file are comment, in two hunks: the rationale block above the attribute, and the reworded justification above `MaxWarningExpansion`, whose value is unchanged. Everything outside that file is Markdown.

## Why

The detector's contract changed under #65 §5.2 without its trigger changing. It used to flag a shape that always discarded a spoken slot value; coverage demotes the bare pattern, so for most grammars it now flags one that could merely be authored to win by more — while still emitting on every parser construction, in shipped builds, and again on every `SetActiveSets` rebuild (compounding `KNOWN_LIMITATIONS.md`'s "Validation warnings re-emit on every active-set switch"). A warning that loud on grammars that now behave correctly is the kind that gets globally suppressed, taking the residual case with it.

So it becomes what it now is — authoring guidance — delivered where authoring happens.

**Why not narrow the trigger to the residue**, which the issue offered as the alternative:

- **#82 has since widened the residue.** It merged as #88 after this issue was filed. The orphan run now terminates at `IsAdmissibleStart` — a pattern that can begin here *by missing leading elements* — which is strictly wider than `CanStartPattern`. A construction-time narrowing can only consult the cheap start-set half (`_startLiterals`/`_startSlots`), because the real test needs an actual token array. It would therefore have gone silent on exactly the cases #82 added.
- **At `coverageWeight: 0` the hazard is not residual at all** — it reverts wholesale, pinned by `Coverage_WeightZero_RevertsBothSidesTogether`. Any narrowing has to stay full-breadth there anyway.
- **The remedy is the better authoring either way**, and it reaches the residue coverage alone does not — though it is *not* the only edit that does: removing the literal outright (`["decelerate", "{burn_level}"]`) reaches it too, scoring the same `1.0` and taking the same span tie-break, at the cost of the phrasing. That rationale is already recorded in the code and in four doc pages; narrowing would have meant reversing it.

One correction to the issue text: the detector is `static` but already runs at ctor line 394, *after* `_startLiterals`/`_startSlots` are built — "runs before those caches are read" is no longer true, so the instance-method conversion it anticipated was not the obstacle.

## Docs

The "deliberately not narrowed" rationale is asserted in five places, none of which said *where* the warning lands. `scoring.md`, `command-recognition.md`, `troubleshooting.md` and `KNOWN_LIMITATIONS.md` each now say Editor-only, so nobody hunts for it in a device log. The re-emit limitation drops this pass from its root cause — in a player build only the slot-value/alias and optional-expansion passes still re-emit.

`CHANGELOG.md` **amends the existing #42 entry** rather than adding a new one: that entry is still under `[Unreleased]`, so the warning has never shipped and there is nothing for a user of the next release to notice changing.

## Validation

- [x] Unity **EditMode 124/124**, **PlayMode 424/424**, both `failed="0"`, on this working tree via the host project (`VoXR TestGround`, 6000.4.7f1). Fresh run — result XMLs post-date the run mark and the log shows a genuine `Csc … Jinwoo1601.VoXR.Runtime.dll [CacheMiss]`, not a cache replay.
- [x] **All 14 tests that `LogAssert.Expect` this warning still pass**, so it still fires under `UNITY_EDITOR`. 14 expectation sites across 3 files: 11 in `VoxrCommandParserTests.cs` (including both residue pins, `Coverage_ResidualHazard_WhenTheStrandedValueBeginsAPattern` and `Coverage_LeavesTheStrandedArgumentHazard_WhenTheValueBeginsAPattern`, and `Coverage_WeightZero_RevertsBothSidesTogether`), 2 in `VoxrEagerCommitTests.cs`, and 1 in `VoxrCommandRecogniserInjectionTests.cs:731` — the last inside the shared helper `ConfigureForUnanalysableGrammar`, which carries an explicit **emission-order** coupling to the modified method ("the ctor runs the #42 scan before the optional-expansion one — so this pair must stay in this order") and is reached by 3 tests. The ctor's pass order at `VoxrCommandParser.cs:393-395` is untouched, and no `Tests~/` file is in the diff.
- [x] `NativeBridge~/` untouched, so the bridge gates do not apply and were not run.
- [ ] **Not covered by any gate here:** the thing this change actually does — the warning's *absence* in a built player — is unobservable to an editor-run suite by construction. No automated check in this project can cover it; it is a one-line reading of the attribute.
- [ ] On-device (Quest) check: not required. The edit is conditional compilation, and its only runtime effect is one fewer log line.


jinwoo1601 (commented):
## Independent review — PR #89

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE`, **plus `XFILE`**, added because the change is a conditional-compilation boundary whose other end lives in the test assemblies. No angle was dropped.
**Head reviewed:** `f65d2feedd6086e98f0761fa126e37cb3ea46858` · **base (merge-base):** `29dbecf`.

**No blockers.** The code change itself verified clean on every angle: `[Conditional("UNITY_EDITOR")]` is a legal target here (`void`, no `ref`/`out`, not an override), the decision is made at the call site's compilation unit — `VoxrCommandParser.cs:394`, same assembly — the elided argument is a bare identifier so nothing evaluable is lost, and the method's only escape is one `Debug.LogWarning`. No `Application.logMessageReceived` hook exists in the package, so nothing captures the warning into tooling. All four `VoxrCommandParser` construction sites keep it in the Editor. Invariants 1, 2, 4 and 6 hold; the detector's scan body and the ctor's pass order at `:393-395` are untouched.

All four confirmed findings are in **prose**, not code — three of them in the PR's own description of itself.

---

### 1 · The "all five tests" box names 5 of 14 dependent tests — CONFIRMED, minor

The second ticked box reads *"All five tests that `LogAssert.Expect` this warning still pass"* and names five, all in `VoxrCommandParserTests.cs`. The real dependent set is **14 test methods across 3 files** (14 lexical `LogAssert.Expect` sites: 11 + 2 + 1). Unlisted: `RequiredLiteralBeforeSlot_WarnsAtConstruction` (`:2052`) — the warning's own primary test — `RequiredLiteralDropped_SlotFilledPatternNowWins` (`:2074`), `OptionalSlotAfterTheLiteral_AlsoWarns` (`:2166`), `Coverage_WeightZero_RevertsBothSidesTogether` (`:2652`, cited by name elsewhere in the body), `Coverage_LeadingTerm_ReordersSiblingsAtTheSameStart` (`:2726`); `VoxrEagerCommitTests.cs:1266`; and three tests reaching it via `VoxrCommandRecogniserInjectionTests.cs:731`, whose helper carries an explicit **emission-order** coupling to the modified method (`:729` — *"the ctor runs the #42 scan before the optional-expansion one — so this pair must stay in this order"*). All 14 do pass — box 1's full-suite green covers them — but this box is the only evidence offered for the PR's central invariant, and a reader sizing blast radius from it inspects 5 of 14 and never sees the ordering coupling.

### 2 · "the only thing that reaches the residue" is false — CONFIRMED, minor

`VoxrCommandParser.cs:518-519` and `CHANGELOG.md:14` both assert the optional-literal swap is the sole remedy reaching the residue. `KNOWN_LIMITATIONS.md:501-502` — inside the residual-hazard entry this same diff edited — says otherwise: *"Removing the literal outright (`["decelerate", "{burn_level}"]`) works too."* Verified by running the real parser, not by reading: on the residue fixture (`hard stop` registered, "decelerate hard burn"), the stripped pattern scores **1.0 and extracts `burn_level`**, bit-identical to the `?by` remedy — same score, same span, same literal count, so `IsBetterCandidate:1148` resolves it on `ConsumedEndIdx 3 > 1` either way. The base comment was correctly scoped (*"it **also** wins in the residual case"*); the exclusivity is new here. The accurate claim is "the only remedy that reaches the residue **without losing a phrasing**" — the swap is genuinely better (`"decelerate by hard burn"`: `?by` gives one command at 1.0 with the slot, the stripped form splits into two with the primary at 0.50 and no slot), but better is not only.

### 3 · The diff falsified `command-recognition.md:126` and left it standing — CONFIRMED, minor

That line: *"a pattern carrying more than six optional elements is compared unexpanded, since this scan runs on every parser rebuild."* In a player build the scan now runs on **zero** rebuilds. The diff updated precisely this sentence's in-source twin — the `WarningForms` comment gained *"an editor session makes"* — so the qualifier was judged necessary for exactly this claim and then applied to one of its two copies. Worse, the identical clause also survives **on a line this diff itself rewrote**: `CHANGELOG.md:14` says *"the scan runs on every parser rebuild"* in the same paragraph that now says *"**The warning is Editor-only**"* and *"not worth … re-emitting on every `SetActiveSets` rebuild."* That paragraph contradicts itself. `KNOWN_LIMITATIONS.md:507` states the cap without the rationale and is fine.

### 4 · "Everything else in the diff is Markdown" — CONFIRMED, nit

`git diff --numstat` on `VoxrCommandParser.cs` is `20 3` across two hunks 86 lines apart. Exactly **1 of the 20 added lines** is the attribute; the other 19 are C# comment, including a second hunk at `:613-615` rewriting the rationale above `MaxWarningExpansion` — the constant's justification, which is the kind of hunk a reviewer skips after being told the rest is prose. Content is accurate and the constant is unchanged from base; only the body's description of the artifact is wrong.

---

### Refuted

- **`scoring.md:368`'s new clause ("keeping the trigger wide is what preserves the residue") reads backwards and doesn't follow.** Refuted on the technical half: `_orphanRun` terminates on `IsAdmissibleStart` (`:1605`), which needs a token array and is strictly wider than `CanStartPattern` (`:1748`) — the only half `_startLiterals`/`_startSlots` can answer, and the exact mechanism issue #81 proposed narrowing with. So a construction-time narrowing *would* go silent on the admissible-only part of the residue, and breadth is literally what keeps coverage of it. The wording objection survives only as the copy-edit nit in Notes.

### Notes — unverified, one line each, below the materiality floor

- `Tests~/Runtime/Jinwoo1601.VoXR.Tests.Runtime.asmdef` has `includePlatforms: []` (only Android excluded), so the code comment's "pins it in editor Play Mode" is convention, not enforcement — latent only, since the bound verification is `-testPlatform EditMode|PlayMode` and `release.yml` strips `Tests~` entirely.
- `VoxrCommandParser.cs:489` and `KNOWN_LIMITATIONS.md:499` still say the parser "now has the information to narrow it" via `_startLiterals`/`_startSlots` — post-#88 that reads against finding 4's own mechanism. Pre-existing, not introduced here, but this PR edited both paragraphs.
- The new Editor-only sentence at `KNOWN_LIMITATIONS.md:500` lands two sentences *before* the paragraph re-introduces the warning unqualified — right fact, awkward position.
- `scoring.md:368` flips "residue" from fix-target to preserve-target across two sentences; `:76` and `:351` still say "warns … at construction" unqualified but route the reader onward, so neither asserts anything false.

### Not verifiable from here

- **Any Unity compile or test run.** This is a bare UPM package; nothing builds standalone. The EditMode 124/124 and PlayMode 424/424 counts remain an author claim to this review — no angle could re-run them. Finding 1's "all 14 still pass" is a static argument (no `Tests~/` file is in the diff, ctor order untouched), not an observation.
- **On-device (Quest) behaviour**, including that a player build actually omits the log line — which is the one thing this PR changes, and is unobservable to an editor-run suite by construction, as the PR states.

jinwoo1601:
Pushed `15b14a8`, closing all four CONFIRMED findings from the review above.

- **Finding 1** ("all five tests" named 5 of 14) — validation box rewritten with the real dependent set: 14 expectation sites across 3 files, naming the emission-order coupling in `VoxrCommandRecogniserInjectionTests.cs:731`.
- **Finding 2** ("the only thing that reaches the residue") — withdrawn in both places that carried it (`VoxrCommandParser.cs`, `CHANGELOG.md`) and replaced with the reason that actually holds: a construction-time trigger can consult only `_startLiterals`/`_startSlots` → `CanStartPattern`, while the run terminates on the wider `IsAdmissibleStart` (#82). The PR body's third bullet is corrected the same way.
- **Finding 3** (falsified `command-recognition.md:126`, and the self-contradicting `CHANGELOG.md:14`) — both copies now carry the "an Editor session makes" qualifier the source comment already had.
- **Finding 4** ("everything else in the diff is Markdown") — body now says the only other C# edit is comment text, 19 of 20 added lines.

Nothing changed for the refuted finding. The Notes are left as posted.

Re-verified: Unity **EditMode 124/124**, **PlayMode 424/424**, counts identical to the previous run, `Jinwoo1601.VoXR.Runtime.dll` rebuilt after the edit (so the comment change was compiled, not served stale). `NativeBridge~/` still untouched. The attribute and the method body are unchanged by this commit — the diff is comment and prose only.

Issue #91, opened 2026-08-15
Sibling warning's element number can index a different word than the pattern it quotes


jinwoo1601:
Found by the code review of the `feat-sibling-set-detection` branch (#74 backlog item 1); deferred by ruling because the fix reshapes `SiblingSet`, which backlog item 2 touches anyway.

## What happens

`SiblingSet` carries **one** `DiscriminatorIndex`, an index into the collapsed frame, but the message quotes each member's pattern **as authored**. When the members' authored patterns differ in length, one number cannot index all of them.

```
mode_weapons    : ["?please", "switch", "to", "weapons"]
mode_navigation : ["switch", "to", "navigation"]
```

The only shared bucket is the 3-element frame, so the message says *element 3* while element 3 of the first pattern as authored is `to`, not `weapons`.

The longest-frame collapse fixes the **symmetric** case — where both patterns carry the optional, a full-length expansion is also a sibling and wins the collapse — but it cannot help here, because no full-length expansion of the first pattern is a sibling of anything.

## Current mitigation

The message appends `" (with its optional elements omitted)"` to any member whose authored length differs from the frame, so the reader is told the quoted text is not the form being indexed. Requirements §4.4 accepts this as a stated residual rather than a defect. It is still a worse message than it needs to be.

## Fix

A per-member discriminator index (or per-member form length) on `SiblingMember`, so the message can index each quoted pattern correctly. That is a shape change to the primitive, which is why it waits for item 2.

## Severity

Low — cosmetic, Editor-only, and the note flags it. Filed because a warning whose element number points at the wrong word is exactly the kind of small wrongness that erodes trust in the whole diagnostic.

## Where

`Runtime/Commands/VoxrCommandParser.cs` — `SiblingSet.DiscriminatorIndex` and `BuildSiblingWarning`.

jinwoo1601:
## The longest-frame carve-out no longer holds (found by the PR #94 review)

This issue currently scopes itself out of one shape:

> The longest-frame collapse fixes the **symmetric** case — where both patterns carry the optional, a full-length expansion is also a sibling and wins the collapse — but it cannot help here, because no full-length expansion of the first pattern is a sibling of anything.

That carve-out was true when a set's members were normalised to one per discriminating value. **Issue #90's fix (PR #94) removes that normalisation**, so the symmetric case is no longer covered either.

### Why

[...]

Issue #90, opened 2026-08-15
Sibling sets drop duplicate-valued members instead of grouping them


jinwoo1601:
Found by the code review of the `feat-sibling-set-detection` branch (#74 backlog item 1); deferred by ruling because the fix reshapes `SiblingSet`, which backlog item 2 (`feat-tie-aware-selection`) touches anyway.

## What happens

`FindSiblingSets` keeps a member only if it brings both a `(command, pattern)` and a discriminating value the set does not already have:

```csharp
if ((kept[j].CommandIndex == members[i].CommandIndex
        && kept[j].PatternIndex == members[i].PatternIndex)
    || string.Equals(kept[j].Value, members[i].Value, StringComparison.Ordinal))
```

The value arm **drops** the later member rather than grouping it under the value it shares. So with three intents:

```
a : ["mode", "on"]
b : ["mode", "on"]
c : ["mode", "off"]
```

the emitted set is `{a:"on", c:"off"}` and `b` is never named — even though `b`↔`c` is exactly the same hazard as `a`↔`c`. An author who fixes what the warning names leaves the other half live.

## Why it was not fixed in item 1

The value arm is what keeps author-duplicated patterns out of the report (requirements F8), so it cannot simply be removed. The clean fix is for a set's members to be *values with the patterns carrying them* rather than a flat list deduplicated by value — i.e. a shape change to `SiblingSet`.

Item 2 promotes `FindSiblingSets` from an Editor-only call to a runtime consumer keyed by `(commandIdx, patternIdx)`, which is the natural moment to settle the shape with a real consumer in hand. Doing it earlier would guess at what that consumer needs.

## Severity

Low in practice — it needs two patterns whose *normalized* forms coincide across intents, which is already broken authoring. But it is a silent under-report in a diagnostic whose whole job is to not be silent.

One sharper variant found by the same review: because the dedup is greedy and first-wins, a dropped member can leave the survivors sharing a single intent, which the same-intent filter then suppresses entirely — turning an under-report into no report at all.

## Where

`Runtime/Commands/VoxrCommandParser.cs`, the emission-gate loop in `FindSiblingSets`.

Issue #74, opened 2026-08-13
Design: route an ambiguous sibling tie to confirmation instead of firing the first-registered pattern


jinwoo1601:
Design-lane candidate, split out of #65 at G2. **Not a bug** — the current behaviour is deterministic and documented (see the new *"Sibling patterns that differ only in their last word fire the first-registered one"* entry in `KNOWN_LIMITATIONS.md`). This is about whether there is a better answer than guessing.

## The situation

Two patterns share every element but the last — `switch to weapons` / `switch to navigation`, or `set auto pilot on` / `... off`. The speaker says one, VOSK drops the final discriminating word, and the surviving evidence fits **both** siblings equally. They tie on score, on consumed span and on literal count, so selection falls through to its last key: registration order.

```
transcript "switch to"  ->  both patterns score (1+1+0)/3 = 0.67
                        ->  mode_weapons fires, because it was registered first
```

The word that would have decided is exactly the one that went missing, so **no scorer can recover the intent**. Reducing the miss cost (#65) extended this from four-element patterns down to three, which is where two-word-prefix grammars live.

## Why this was not fixed as a bug

Refusing is not obviously better than guessing: on the flush path the transcript is final, so refusing means firing *nothing* — the exact class #65 exists to rescue. The eager gate can refuse cheaply because the missing word may still arrive; a final transcript has no such option. That asymmetry is why #70's tail guard does not generalise here.

## The idea worth designing

Rather than refuse or guess, **detect the tie and route it to the existing confirmation/pending machinery** — `requiresConfirmation` and `OnCommandPending` already exist. "Did you mean weapons or navigation?" is a better answer than a coin flip, and the parser already knows it is a coin flip: the tie is visible at the point selection falls through to registration order.

Open questions for the design:

- What exactly constitutes the tie — equal score *and* equal span *and* equal literal count, or something narrower (the tie-breaking element specifically being the one that missed)?
- Does this need a new verdict/reason on the result, or can it reuse the pending path as-is?
- Should it be opt-in per command, or default-on for grammars that carry the shape? The parser already detects sibling hazards at construction (#42's warning), so it could warn rather than act.
- Interaction with backlog item 2 (`feat-coverage-in-selection`), which changes the score selection compares on — coverage does **not** break the tie here, since both siblings have identical coverage.

## Meanwhile

[...]

PR #92, opened 2026-08-15
Report sibling patterns that tie on a dropped word (#74)

jinwoo1601:
Backlog item 1 of the G1-locked sibling-tie-disambiguation design (issue #74). **Construction-time only — no runtime behaviour change, no public API change.**

## The problem

Two patterns identical but for one required literal tie on *every* selection key when the recogniser drops that word:

```
mode_weapons    : ["switch", "to", "weapons"]
mode_navigation : ["switch", "to", "navigation"]

speaker says "switch to navigation", VOSK drops "navigation"
  → both score (1+1+0)/3 = 0.667, same span, same literal count
  → selection exhausts its keys and falls through to registration order
  → mode_weapons fires, because it was declared first
```

The word that would have decided is precisely the word that went missing, so no scorer can recover the intent. That behaviour is not changed here. **What is fixed is that the author is never told** — the shape is plainly visible at construction, and until now nothing looked.

`WarnOnDroppableRequiredLiteral` (#42, narrowed by #81) does not reach it: that scan needs a strictly-longer pattern, an element-prefix relation, and a stranded *slot*. This shape has none of the three. It detects a dropped argument; this detects a coin-flipped intent.

## What landed

- **`FindSiblingSets`** — the sole definition of the relation (design DR-2): equal length, equal at every position but one, a required literal in each at that position, at **any** position. Buckets each `(form, position)` by the frame left behind, so two forms share a bucket exactly when they are siblings and an n-way set is collected in one pass.
- **An Editor-only author warning**, appended last among the constructor's scans, naming the intents, the patterns as authored, the element position and the discriminating values.
- **A discriminator/cancel collision report** — a discriminating value that is also cancel vocabulary would be swallowed by cancel if the ambiguity is ever routed to the speaker, making that choice unreachable.
- **Two tests confirming design §2.8**, which was load-bearing and unobserved (below).

## §2.8 confirmed

The design reasoned from source that a *medial* discriminator slips every `TryEagerCommit` condition, and rested DR-5 entirely on that without ever observing it. It reproduces:

```
set_mode  : ["set", "{ship}", "mode",  "on"]
set_level : ["set", "{ship}", "level", "on"]
buffer    : "set alpha on"          → EagerCommitVerdict.Commit
```

The miss is a literal, so the slot condition is blind to it; it is medial, so `"on"` matches afterwards and #70's tail counter resets; the buffer is spanned; the score clears the gate; neither pattern is a prefix of the other, so the extendability precompute allows it. The gate fires on evidence that cannot distinguish the siblings. The companion test shows the flush then picks by registration order, and that **reversing the two declarations flips the intent on an identical utterance**.

One design correction: the score is `3.0/4.0 = 0.75`, not §2.8's `2.0/2.5 = 0.8` — that constant belongs to a five-element example already in the suite. Both clear the gate, so the reasoning is unaffected.

## Measured (design §7.3 gates DR-7's "warning on by default")

Over the shipped demo grammar — 11 commands, 32 patterns:

| | |
|---|---|
| Sibling sets the relation admits | **11** — 5 cross-intent, 6 same-intent |
| Same-intent | **all 6 synonym authoring** — the same command fires either way, so they never warn |
| Cross-intent that are **reachable** | **1** — `switch to weapons` / `switch to navigation`, the #74 pair itself |
| Warnings actually emitted | **1** |
| Existing tests needing amendment | **1** — the one that builds the #74 grammar |

Two narrowings get from 11 to 1, and both are ruled rather than assumed.

**Intent.** A set whose members share an intent cannot fire the wrong intent — the same command is dispatched either way. All six are ordinary synonym authoring.

**Reachability.** Losing one required element from a frame worth `D` leaves `(D−1)/D`. Four of the five cross-intent sets have two-element frames, so they drop to `0.5` — under the default `minScore`, which rejects **both** siblings. Nothing fires, rather than the wrong thing, and this repo already pinned that twice (`MissedLiteral_TwoElementPattern_StillRejected` and its recogniser-level counterpart `_DoesNotFire`). Warning "the wrong intent can fire" there would be false, and `requiresConfirmation` is inert when no command dispatches.

The check weighs elements rather than counting them: a matched slot credits `MatchScore` whether optional or required, and only an optional literal credits `OptionalLiteralScore`. So `switch ?to weapons` reaches exactly `1.5/2.5 = 0.60` and warns, while `cease fire` sits at `0.5` and does not.

**Residual, bound for `KNOWN_LIMITATIONS.md`:** `minScore` is user-configurable and the parser constructor cannot see it, so the gate judges against the default. An author who lowers it below `(D−1)/D` makes the shorter ties live and gets no warning. Erring the other way would put a knowingly false claim in front of every author who left the knob alone.

Both counts are pinned — `SiblingSets_DemoGrammar_VolumeAndOrderAreStable` for the relation, `SiblingWarning_DemoGrammar_WarnsOnlyOnTheReachableTie` for what reaches the console — and both read `DemoGrammar.AllCommands()` directly rather than a transcription.

## Review

Two independent passes — a `review-cycle` on the branch and a full-profile `review-pr` on the diff — 23 finder angles and 5 adversarial verifications between them. Findings that changed the code:

- **Optional literals are no longer folded when comparing frames.** `?to` and `to` consume the same token but do not *score* the same — a matched optional literal credits `0.5` to both sides against a required literal's `1.0` — so those patterns never tie. Optional *slots*, which are genuinely score-neutral, still fold.
- **The reachability gate was incomplete**, in the three ways described above. Fixed at two layers: the admission-rule half in the primitive (knob-free), the `minScore` half in the warning (judged against a default the constructor cannot see).
- **The cancel-collision report no longer inherits the same-intent filter**, which rests on a harm model that does not apply to it.
- **Dead code and a vacuous test removed** — the per-`(command, pattern)` dedup guarded against a mechanism that disappeared with optional-literal folding.
- **The volume test now calls the shipped `IsSingleIntent`** instead of its own copy of the rule, and `AllCommands()` flattens by activation order rather than coincidentally agreeing with it.
- **Allocation:** frames were built for every bucket before anything knew whether it would survive, and then again for every duplicate set before the collapse discarded it. Both deferred.

One finding was **refuted** and correctly produced no change: that the collision report should also test confirm vocabulary. Design §5.5 rules confirm out during disambiguation, so a confirm-valued discriminator stays reachable.

An earlier revision of this PR body carried allocation figures ("38.6 MB per construction", "−39% time, −24% allocation") taken from a review harness that was never committed. An independent simulation put the same quantity at roughly 100 KB — a ~400× disagreement — and attributed most of the real saving to hoisting normalization rather than to the frames. The saving is real; the magnitudes were not reproducible from the branch and have been withdrawn rather than restated.

Deferred to #90 and #91 — both reshape `SiblingSet`, which item 2 touches anyway.

## Verification

EditMode **124/124**, PlayMode **451/451** (Unity 6000.4.7f1 via `VoXR TestGround`). `NativeBridge~` untouched, so no bridge build and no harness re-baseline. No on-device verification required and none claimed.

## Scope

Backlog items 2–4 are deliberately absent: no three-state comparator, no eager refusal, no pending machinery, no opt-in flag. `IsBetterCandidate` keeps its `bool` return and exact key list. Issue #74 stays open — item 3 is where the behaviour it asks for arrives.

Product docs follow after G2, before merge, per the workflow.


jinwoo1601 (commented):

**Head reviewed:** `d8101344d1b81e21ff11352eb8c470508da8c71b` · **Base:** `ef84690` (merge-base) · **Profile:** full — `CLAIM` plus all 11 roster angles, no angles dropped. Every numbered finding below carries a verifier verdict.

A pre-PR `review-cycle` already ran on this branch and its findings are closed in `d810134`; this pass had fresh eyes on the final diff and was told not to re-report those. What follows is new.

**Headline:** the feature is sound, verified green (EditMode 124/124, PlayMode 444/444), and §2.8 is genuinely confirmed. But the *evidence for DR-7's default-on ruling is overstated*: the reachability gate this feature added models whether a frame leaves a remainder, not whether that remainder is enough to produce a tie. Three angles found three faces of that one omission, and all three verified CONFIRMED. Finding 1 is the one that matters at G2.

---

### 1 · MAJOR · CONFIRMED — four of the five demo-grammar warnings assert a harm unreachable at default settings

The PR states *"all 5 genuine"* and *"a 0% false-positive rate"*. Of the five cross-intent sets the demo grammar warns about, four have **2-element frames** — `cease/resume` (`* fire`), `stop/resume` (`* firing`), `weapons/navigation` (`* mode`), `enable/disable` (`* all`). Only `switch to *` is 3-element.

A dropped required literal credits the denominator but not the numerator (`VoxrCommandParser.cs:1743`, `RequiredLiteralMissPenalty = 0f`), so an L-element pattern losing one scores `(L−1)/L`. For L=2 that is **0.5**, against `minScore = 0.6f` (`VoxrCommandRecogniser.cs:31`). The gate lives at the recogniser (`VoxrCommandRecogniser.cs:756`), and this repo already pins the outcome twice:

- `VoxrCommandParserTests.cs:3119` `MissedLiteral_TwoElementPattern_StillRejected` — *"'cease fire' heard as 'fire' scores (0 + 1) / 2 = 0.5 … and stays under the gate"*
- `VoxrCommandRecogniserInjectionTests.cs:834` `MissedLiteral_TwoElementPattern_DoesNotFire` — `Assert.AreEqual(0, recognisedCount)`, the text routed to `OnUnrecognisedSpeech`

So on those four, the discriminator dropping makes **nothing** fire — both siblings score 0.5 and both are rejected. The emitted message says *"selection falls through to registration order, so the wrong intent can fire"* (`VoxrCommandParser.cs:1148`), which is false for them, and the remedy it offers (`requiresConfirmation`) is inert when no command fires. `Documentation~/troubleshooting.md:81` independently rules that lowering `minScore` is the wrong fix for exactly this shape.

**Threshold:** `(L−1)/L ≥ 0.6 ⟺ L ≥ 3`. At default settings the demo grammar contains **one** reachable sibling tie, not five.

This is the same reachability standard the feature itself used to suppress empty frames — but the parallel is not exact, and the difference is worth stating: the empty-frame suppression rests on `Score <= 0f` in `IsBetterCandidate` (`:1571`), which is knob-free, whereas `minScore` is user-configurable and the parser constructor — `(slots, commands, coverageWeight, additionalGrammarWords)` — cannot see it. An author who lowers `minScore` to 0.5 makes these ties live. That is a real argument for warning conservatively; it is simply nowhere recorded, and the PR presents the count as though every warning were reachable.

Neither the warning volume nor the DR-7 ruling is necessarily wrong. What is wrong is the claim that all five are genuine — which is precisely the evidence the human is being asked to rule on.

### 2 · MINOR · CONFIRMED — a remainder of only optional elements warns, but the admission rule refuses both siblings

The one reachability gate is `if (form.Length < 2) continue;` (`:855`), which counts elements rather than elements that credit evidence. The optional-literal branch (`:1727-1740`) credits `rawScore`, `denominator`, `literalCount` and `consumedEndIdx` but **never** `matchedRequired`.

So `["?please","weapons"]` vs `["?please","navigation"]` buckets (identical text on both sides, since optional literals are no longer folded), passes every member gate, is cross-intent, and warns — yet on `"please"` both members reach `matchedRequired=0, missedRequired=1` and both are refused by the DR-7 admission rule at `:1603`, before any comparison key. `ParseInternal` has no fallback (`:1279-1280`), so nothing fires and the message's claim is unreachable.

### 3 · MINOR · CONFIRMED (narrowed) — `{?x}` → `{x}` folding is score-equivalent but not admission-equivalent

The fold is justified as score-neutral, and that half is correct: the matched-slot branch does `rawScore += MatchScore; denominator += MatchScore;` unconditionally (`:1705-1706`). But the next line is `if (!isOptional) matchedRequired++;` (`:1707-1708`), and `MatchedRequired`/`MissedRequired` are the admission key, not score bookkeeping.

**The fold is not generally unsound** — `SiblingSets_OptionalSlotInTheFrame_StillMatches` (`:3554`) pins a genuine 0.75/0.75 tie, because there the remainder still credits required elements on both sides. The defect is narrower: where the remainder credits **zero** required elements to one member. `a ["{?ship}","mode"]` vs `b ["{ship}","level"]` warns, but on `"alpha"` a is `0/1` (refused) while b is `1/1` (admitted) — equal score, equal span, equal literal count, separated only by admission, so registration order is never reached.

**Findings 2 and 3 share one fix:** require each member's remainder to carry at least one element that credits `matchedRequired`. Implementation note — this cannot be computed from `SiblingSet.Frame`, since normalization erases `{?ship}` vs `{ship}` (`:993`); it must be per member against that member's own form, which `SiblingBucket` currently discards. Cheapest shape is an int captured at collection time (`:893-895`), where the form is in hand.

### 4 · MINOR — the allocation deferral stops one step short of the case it was built for

`var set = new SiblingSet(order[o].Discriminator, order[o].BuildFrame(), kept.ToArray());` (`:947`) runs **before** `IndexOfSameMembers` (`:963`), and the tiebreak that follows reads only `set.Frame.Length` (`:966`) — a value already available as `order[o].Form.Length` without allocating. When one hazard surfaces under F expansion frames — exactly what the collapse exists for — F−1 frames and F−1 member arrays are built and immediately dropped. Simulated on 200 patterns × 6 optionals: 1,440 pairs built, 45 sets survive, **96.9% discarded**. That is the same defect class the headline fix closed, relocated from bucket-creation to emission.

### 5 · MINOR — the self-pairing guard is now dead code, and its test passes vacuously

`SiblingSets_OnePatternsOwnForms_AreNotSiblingsOfEachOther` (`:3666`) documents the collapse `[one, one, two, three]` vs `[one, two, two, three]` — which required the optional-literal folding that `d810134` removed. As shipped, those two forms normalize to `["?one","one","two","three"]` and `["one","two","?two","three"]` and never share a bucket, so `kept` never receives a second member and the `(CommandIndex, PatternIndex)` half of the dedup at `:926-927` is unreachable. Two independent derivations failed to construct any input that reaches it: with slot-only folding, two same-length forms of one pattern can differ only at optional positions, and an optional position can never be the discriminator. The test asserts `0` and passes for the wrong reason.

### 6 · MINOR — `AllCommands()` flattens by declaration order while claiming activation order

`DemoGrammar.AllCommands()` documents itself as *"flattened in the order a parser with all three sets active sees them"* but derives that from `Build`'s array literal, whereas the parser's order comes from `CommandSetManager.Activate` concatenating in **`SetActiveSets` argument order** (`CommandSetManager.cs:64-70`). They agree today and the pinned order is correct — but nothing couples them, so reordering only the `SetActiveSets` arguments would leave the order-pinning test green against an order no parser builds. Given this helper exists specifically because the previous measurement ran against a grammar the parser never sees, the failure class survives in reduced form.

### 7 · MINOR — the volume test re-implements the filter it claims to measure

`SiblingSets_DemoGrammar_VolumeAndOrderAreStable` computes the cross/same split with its own inline loop (`:3776-3785`) rather than calling `IsSingleIntent` (`:1099`), which is `private` with one call site and no test of its own. So the `5 warn / 6 suppressed` split — the evidence DR-7's ruling rests on — is asserted against a *copy* of the rule, and the test's stated purpose ("a later change that makes this scan noisier cannot pass unnoticed") does not hold for a change to the filter itself. This is the same defect class the pre-PR review fixed one level down: the grammar transcription was removed, the filter transcription was not.

### 8 · MINOR — the allocation figures in the PR body are unsubstantiated and do not reconcile

*"86% allocated and never read, 38.6 MB per construction on a 200-pattern grammar … −39% time, −24% allocation"* and *"~0.097 ms per construction"* appear nowhere a reader can check — no benchmark, no allocation pin, no committed harness; `grep` finds none of the figures in the repo. An independent simulation puts the removed per-bucket frames at roughly **100 KB** for that grammar, not 38.6 MB — a ~400× disagreement — and attributes most of the real saving to the `NormalizeElement` hoist rather than the frames. The saving is real; the stated magnitude and its attribution are not supported. The repo already has the idiom for pinning this (`ZeroAllocPollPathTests.cs:27`, `GC.GetAllocatedBytesForCurrentThread`). Separately, the `~0.097 ms` figure was taken *before* the rework the same PR credits with −39% time.

### 9 · MINOR — superseded counts left in a test comment (and in the gitignored architecture doc)

`VoxrCommandParserTests.cs:3701-3704` still reads *"all five same-intent sets were ordinary synonym authoring and all four cross-intent ones were real hazards"* — the pre-fix 9/4/5 figures — 100 lines above the assertions that pin 11/5/6 in the same file, added in the same commit. The gitignored `architecture.md` "Measured outcome" block carries the same superseded numbers while its own §7.1 table and `requirements.md` §8.3 carry the corrected ones.

### 10 · MINOR — requirements F14 and F15 no longer describe the shipped behaviour

F14 still states frame comparison ignores optional decoration and that `["switch","?to","weapons"]`/`["switch","to","navigation"]` **must** be detected — the shipped `SiblingSets_OptionalLiteralInTheFrame_IsNotASet` asserts the opposite, deliberately. F15's acceptance text states *"items 2–3 still see every set the relation admits"*, which is now false: `form.Length < 2`, the `(ci,pi)` dedup, the value dedup and `kept.Count < 2` all remove sets inside the primitive. The two-layer model (DR-1 relation, then harm-reachability) is the right shape and DR-1 is **not** being challenged — but the routine advertises one contract in its doc-comment and delivers another three lines later, and item 2 is routed through it by DR-2.

### 11 · MINOR — the "(with its optional elements omitted)" branch is reached by no test

`grep "optional elements omitted" Tests~/` returns nothing. The branch (`:1126-1127`) is easily reachable — `["?x","switch","to","weapons"]` vs `["?y","switch","to","navigation"]` yields only the length-3 bucket, so both members get the suffix and the element number points at `to` in both quoted texts. That is exactly the F6/§4.4 limitation the suffix exists to soften, and nothing verifies the softening is emitted or attached to the right member.

---

### Refuted, and correctly produced no change

- **That the collision report should also test the confirm vocabulary.** Design §5.5 rules confirm out during disambiguation (*"cancel vocabulary continues to apply; confirm vocabulary does not"*), so a confirm-valued discriminator stays reachable. Only cancel is swallowed. The cancel-only scope is right.

### Verified clean

- **Invariant 4** — the whole `WarnOnDroppableRequiredLiteral` block (method, `WarningForms`, `MaxWarningExpansion`, `BuildDroppableLiteralWarning`, `IsElementPrefix`) is **byte-identical** to base; ctor call order preserved with the new scan appended last.
- **Invariants 1, 2** — the parser diff is four purely additive hunks; `IsBetterCandidate` byte-identical to base; every added `public` member sits on an `internal` struct or a private nested class.
- **Invariant 3** — `SiblingScan_IsEditorOnly` genuinely bites (`ConditionalAttribute` is emitted into metadata; deleting it fails the assert), and `FindSiblingSets` has no Runtime caller outside the elided one.
- **Invariant 9** — the ordered `CollectionAssert` would fail under hash iteration order. (The 6 same-intent sets' relative order is unpinned.)
- **`DemoGrammar` refactor** — grammar data byte-identical to base; `Configure` passes the same slots, same set order, same `SetActiveSets` arguments; `VoxrWavReplayTests` behaviourally unchanged.
- **Claim 6** — two independent hand-derivations reproduce **11 sets, 5 cross-intent, 6 same-intent** with the pinned emission order matching element-for-element.
- **Claim 13** — the archived result XMLs contain head-only test names, so the 124/444 run is at `d810134`, not an earlier commit.
- **PIT** — no findings. `SiblingBucket.Form` does alias the pattern array for zero-optional patterns, but is safe (`VoxrCommandDefinition` deep-copies; nothing writes). The `normalized` scratch has no stale-read path. No new construction-time throw.
- **`MaxWarningExpansion` fallback** — the over-cap raw pattern is exactly the all-optionals-included form, so it still keys consistently; the cost is recall only, no false positives.

### Notes (unverified, one line each)

- `SiblingSet.Frame`'s contents are never read — only `.Length`, at three sites.
- The comment at `:852-854` points at "the emission loop" for a rationale that moved with the gate in `d810134`.
- `JoinWith` returns `" and A"` for one part and throws for zero; all call sites are provably ≥2.
- Nothing pins that `requiresConfirmation`/`cancelVocabulary` still exist under those names, though the repo pins names by reflection elsewhere.
- `FindSiblingSets`'s `commands == null` guard is unreachable; the ctor throws first.
- `WarningForms` is walked twice per pattern per construction — once by each scan.
- Remaining allocation: a `List<SiblingMember>` per bucket (84% never reach two members), a key string per bucket *hit*, and a second `ExtractSlotName` sweep in the discriminator loop.
- `Sib`/`SibP` duplicate `Cmd`/`P` from the sibling test file; the `" (with its optional elements omitted)"` literal now exists in three places.
- `DemoGrammar.cs`'s `Purpose:` header still says "for acoustic replay tests" though it now also serves grammar analysis.
- The 19-line #74 hazard doc sits above two `const char` declarations rather than above the scan it explains.
- `EagerFlush_PatternStillOwingItsLastWord_FiresTheRightCommand` configures a cross-intent sibling grammar and now warns without declaring it — harmless, since it makes no absence assertion.

### Not verifiable from here

On-device Quest behaviour. Nothing in this diff touches capture, the native bridge, or MonoBehaviour lifecycle, so none is required — and none is claimed. The Unity results quoted above were read from the archived XMLs rather than re-run.

jinwoo1601:
### Follow-up: the review's confirmed findings are closed in `b3cb1c8`

The reachability defect the review found is fixed, and the fix is in two layers because the two halves are not the same kind of rule.

**In the primitive (knob-free).** A form with no required element outside the discriminator is never collected. DR-7's admission rule refuses such a candidate before any comparison key, so it cannot be a rival in a tie whatever it scores. That closes both the both-refused case (`["?please","weapons"]` / `["?please","navigation"]` — an optional literal credits score and span, never `MatchedRequired`) and the asymmetric one (`["{?ship}","mode"]` / `["{ship}","level"]` — folding `{?ship}` onto `{ship}` is right about *score* and wrong about *admission*).

**In the warning (default-relative).** A tie is reported only if it would clear the default `minScore`. Losing one required element from a frame worth `D` leaves `(D−1)/D`, so a two-element pattern drops to `0.5` and both siblings are rejected — which this repo already pinned twice. The check weighs elements rather than counting them, so `switch ?to weapons` lands on exactly `0.60` and warns while `cease fire` stays at `0.50` and does not.

The split matters: `minScore` is configurable and the constructor cannot see it, so judging against the default is an audience call, not a property of the relation. A later runtime consumer applies its own threshold and should not inherit an assumption about someone else's.

**Result: the shipped demo grammar emits one warning, not five** — the #74 pair itself. Pinned by `SiblingWarning_DemoGrammar_WarnsOnlyOnTheReachableTie`.

Also closed from the same review: the dead `(command, pattern)` dedup and its vacuously-passing test; the volume test now calling the shipped `IsSingleIntent` rather than its own copy; `AllCommands()` flattening by activation order rather than coincidentally matching it; and frame/member-array construction deferred past the duplicate check.

The PR body's allocation figures have been **withdrawn** rather than restated — they came from an uncommitted harness and an independent simulation disagreed by ~400×.

**Residual, and deliberately not fixed here:** an author who lowers `minScore` below `(D−1)/D` makes the shorter ties live and gets no warning. Bound for `KNOWN_LIMITATIONS.md` in the post-G2 doc pass.

Findings deferred by ruling remain #90 and #91. EditMode 124/124, PlayMode 451/451.

Issue #93, opened 2026-08-16
Sibling-set edge branches and lookup invariants have no executable coverage


jinwoo1601:
Found by the `review-cycle` pass on `feat-tie-aware-selection` (#74 backlog item 2), deferred by ruling 2026-08-16 so the feature's confirmed findings could land first. Three related gaps, all "nothing fails if this is deleted".

## 1. The two dedup branches issue #90 forced are untested

Retaining duplicate-valued members made two display branches reachable, and both were added blind:

- `BuildSiblingWarning`'s value-list dedup (`if (!values.Contains(quoted))`)
- the cancel-collision loop's `alreadyReported` guard

Delete either and both suites stay green. Every exact-message assertion in the suite uses a distinct-valued fixture; the one fixture that *does* carry a duplicate value asserts only `new Regex("differ only at element 3")`, which matches whether the message reads `("mode" or "level")` or `("mode", "level" or "level")`. No fixture anywhere pairs a duplicated value with cancel vocabulary, so the collision dedup has no exercise at all.

A third branch of the same family — the pattern-text dedup — *is* now pinned, by `SiblingWarning_TwoIntentsSharingPatternText_NameThatTextOnce`. These two should match it.

## 2. The per-round tie reset is unpinned

`ParseInternal` declares its tie locals inside the extraction loop, and the comment names its own failure mode: hoisting them out "compiles and passes any single-command test". Every test added is a single-command test — all four Editor diagnostic tests assert `diag.Length == 1`. The only multi-round diagnostic test parses a grammar with no sibling sets, so no rival is ever recorded in either round.

Wanted: one Editor test where a sibling-tied command is followed by an unambiguous second command in the same utterance, asserting `TiedSiblingIntent` is null on the second entry.

## 3. Invariants 8 and 9 have no executable guard

- **Allocation-free selection.** `CompareCandidate` and `AreSiblingRivals` run in the innermost loop of both selection paths. Nothing would catch an allocation introduced beside them. The repo already has the instrument — `Tests~/Runtime/ZeroAllocPollPathTests.cs` uses `GC.GetAllocatedBytesForCurrentThread()`.
- **Lookup built once per parser.** Requirements F13's acceptance is "parsing the same utterance N times calls `FindSiblingSets` once". Deleting the `if (_siblingLookupComputed) return;` guard leaves both suites green.

[...]

PR #94, opened 2026-08-16
Refuse the eager commit on a sibling tie (#74 item 2, closes #90)

### LINKED ISSUE #90: Sibling sets drop duplicate-valued members instead of grouping them

jinwoo1601:
Backlog item 2 of the G1-locked `sibling-tie-disambiguation` design (issue #74). Teaches the candidate comparator to report an exact tie, and makes the eager-commit gate refuse to fire early when the tie is between two *sibling* patterns — two intents separated by one required literal that the recogniser dropped.

**The fired intent does not change. Only the timing does.**

## Why

Item 1 landed the test proving design §2.8: a **medial** discriminator slips every condition the eager gate has.

`set {ship} mode on` against `set {ship} level on`, heard as `set alpha on`, scores `3/4 = 0.75`, fills every slot, spans the buffer, and both patterns are terminal and unextendable. So the gate commits — on evidence that cannot tell the two intents apart — and registration order picks the winner. Reverse the two declarations and the same utterance fires the other intent.

The trailing case is milder: #70's tail condition already refuses it, so the parser waits and a late discriminator still lands. The medial case had no such wait. `KNOWN_LIMITATIONS.md` put it exactly right — *"both paths guess; the medial one guesses sooner"*. This removes the "sooner".

## What changed

| | |
|---|---|
| **DR-3** | `bool IsBetterCandidate` → `CandidateOrder CompareCandidate` returning `Better`/`Tied`/`Worse`. `Better` is returned in exactly the cases the old bool returned `true`, so selection is bit-identical |
| **DR-5** | `TryEagerCommit` records the first tied sibling rival and returns `None` when it has one. Not behind any flag |
| **#90** | `FindSiblingSets` keeps every member carrying a hazard instead of dropping duplicate-valued ones — a correctness prerequisite, not a cleanup (see below) |
| **§9 row 2** | The flush loop records the tied rival into the Editor-only `ParseDiagnosticEntry`. Inert here; item 3 consumes it |
| **F16** | The warning's remedy no longer says "Diverge earlier", which does not work |

**The condition sits at the very end of `TryEagerCommit`, after both `HoldExtendable` returns.** Placed beside #70's it would also convert `HoldExtendable` into `None`, lengthening a wait on a buffer that was never going to commit early. The only transition introduced is `Commit → None`.

**Why #90 had to land first.** Its gate dropped members that repeated a discriminating value. Where the dropped member was the only one carrying a second intent, the survivors shared one intent, the same-intent filter suppressed the set entirely, and the hazard was invisible to the runtime lookup. `TryEagerCommit_ShadowedCrossIntentMember_Refuses` fails on `main`.

## Measurements (design §7 item 2)

Rig: `Planning~/features/coverage-in-selection/ab-rig`, before = `f482154`, after = this branch. Validated against the committed grammar pin first (133 entries, exact match).

| | before | after |
|---|---|---|
| 699-corpus `Parse` | — | **699/699 rows identical** — no change to intents, slots or scores |
| `--prefix-verdicts` (4014 rows) | 435 `Commit` / 96 `HoldExtendable` / 3483 `None` | **identical — zero rows moved** |
| steady `TryEagerCommit` | ~0.020 ms/call | ~0.020 ms/call — unchanged (independently corroborated at review) |

**The corpus is a control, not evidence.** Its one reachable sibling tie — line 148, `switch to` — is a *trailing* discriminator, already refused by #70. The delta lives in purpose-built medial fixtures, where the flush intent is unchanged in every one.

> **Corrected after review.** An earlier version of this table carried a `FindSiblingSets` speedup and a cold-start figure, and claimed the unchanged `HoldExtendable` count of 96 was "the direct check on the condition's placement". All three are withdrawn: two harnesses disagree on whether the speedup exists, the cold figure came from a probe not reproducible from the branch, and the review *built* two deliberately-misplaced variants that produced byte-identical corpus output — so the 96 is blind to the error it was offered against. The real guard on placement is `TryEagerCommit_SiblingTieOnAnExtendablePattern_StillHoldsExtendable`. See the review and follow-up comments.

## Review

Two independent review passes ran and **both are posted on this PR** — read them rather than this summary. The branch pass (`review-cycle`) found four confirmed issues; the PR pass (`review-pr`, full profile, 12 angles) found sixteen, including a requirements-F10 violation introduced by the branch pass's own fix. All are resolved at `38ac145`; the follow-up comment maps each finding to its resolution.

Two findings are worth the reviewer's attention:

- **A hole this feature opened.** `WarningForms` caps expansion at 6 optionals; `ComputeCanCommitEarly` abandons at 12 and fails *conservatively*. So a pattern with 7–12 optionals kept a live eager commit while its sibling relations went unanalysed. New to this change — at `f482154` `FindSiblingSets` had one caller and it was `[Conditional("UNITY_EDITOR")]`. Now fixed: truncation is recorded per pattern and the gate refuses rather than claiming no hazard from an analysis it never ran.
- **Canon contradicts itself.** Design §5.3 has the Editor diagnostic see *every* tie; §9 row 2 has it see *sibling* ties only. The build had silently picked a side. Kept as built by ruling, with the conflict and its consequence recorded in requirements §4.8 rather than papered over. **This is the one item I'd most like a second opinion on at G2.**

Deferred by ruling → #93 (edge-branch coverage and the two lookup invariants). Issue #91 stays deferred to item 3, which rewrites the same message.

## Verification

EditMode **128/128**, PlayMode **477/477**. Baseline at `f482154` was measured, not assumed: 124/124 and 451/451.

Product docs are **not** in this PR — they are written after G2 per the workflow. The sites needing updates are catalogued (`command-recognition.md`, `troubleshooting.md`, `scoring.md`, `KNOWN_LIMITATIONS.md`, `CHANGELOG.md`), including four in `scoring.md` and the unshipped `[Unreleased]` entry that the review found and the docs had missed.

Issue #74 stays **open** — item 3 delivers the behaviour it asks for.


jinwoo1601 (commented):
## Independent review — PR #94 @ `54ff43f`

**Profile:** full (feature lane, G2 at stake) — `CLAIM` + all 11 `angle-roster` angles, no angles dropped or added. Three reasoned findings then put through adversarial verification; the rest carry their own executed evidence.

**Disclosure:** the branch and this review were produced by the same assistant. Independence comes from the finder agents inheriting nothing but a written brief — they did not see the authoring session. Two findings below contradict claims in the PR body, and one contradicts a fix made earlier on this branch.

**What was re-run rather than trusted:** the 699-corpus `Parse` A/B, the 4014-row `--prefix-verdicts` A/B, the grammar pin, the `FindSiblingSets` timing, both build configurations (with and without `UNITY_EDITOR`), and the "fails on `main`" claim. The Unity suites could not be run from the review environment.

---

## Major

### 1. The truncation arm refuses on non-siblings, in player builds — violates F10 (Must) · CONFIRMED, severity raised

`VoxrCommandParser.cs:1453`. The conservative arm added for the previous review's WRAP-1 finding:

```csharp
if (_siblingFormsTruncated[ci1][pi1] || _siblingFormsTruncated[ci2][pi2])
    return true;
```

sits before the membership lookup, so it answers "these are sibling rivals" for **any** cross-intent tie touching an over-cap pattern — the two need share no set and need not be set members at all.

Observed on head, rig staged from `54ff43f`, `alpha = ["go","?p1"…"?p7","north"]` vs `bravo = ["go","north"]` (identical discriminator ⇒ duplicates, not siblings; `FindSiblingSets` returns **0 sets** at both 6 and 7 optionals):

```
optionals=6   sets=0   TiedSiblingIntent=null    eager("go north")=Commit
optionals=7   sets=0   TiedSiblingIntent=bravo   eager("go north")=None
```

The only difference is `_siblingFormsTruncated`, and **both rows reproduce in the player configuration**. This is therefore not the Editor-only diagnostic mismatch it was filed as:

- **F10 (Must)** — *"A non-sibling tie does not trigger the refusal … authoring errors stay out of the runtime path"* — is false past 6 optionals.
- **Requirements §4.8**, written on this branch, states only the false-*negative* direction (*"all leave `TiedSiblingIntent` null"*). The false positive is unstated.
- **`VoxrCommandParser.cs:2089`** — *"set only when the winner was decided by registration order over a rival differing on one dropped word"* — `bravo` differs from `alpha`'s all-dropped form at **zero** elements.

The tension is structural, not an oversight: "unknown ⇒ refuse" cannot be narrowed without performing the analysis that was skipped. Closing the 7–12-optional hole conservatively necessarily over-refuses. Harm is bounded — latency only, the same intent still fires via flush — but the requirement is a Must and the record is wrong. Ruling needed on whether to amend F10, narrow the arm to the eager path only, or revert to documenting the hole.

No test covers a non-sibling tie above the cap: `TryEagerCommit_NonSiblingTie_StillCommits` uses zero optionals, and `TryEagerCommit_SiblingAnalysisTruncatedByTheExpansionCap_RefusesAnyway` uses a real sibling pair.

### 2. "The unchanged `HoldExtendable` count of 96 is the direct check on the condition's placement" is false · CONFIRMED by construction

PR body, mirrored at `architecture.md:443`. Two scratch variants of the head parser were built making exactly the mistake `architecture.md:202` describes — the refusal inserted above the `HoldExtendable` returns, once immediately after `if (bestHasUnmatchedRequiredTail)` and once immediately before it. **Both produce `--prefix-verdicts` output byte-identical to the shipped placement: 435 / 96 / 3483, zero rows differing.**

No corpus prefix pairs a sibling tie with an extendable pattern, so the 96 is blind to the very error it is offered as the check on. The real guard is the unit test `TryEagerCommit_SiblingTieOnAnExtendablePattern_StillHoldsExtendable`. The claim appears in the PR body, `architecture.md:443`, and a commit message; all three should name the test and describe the 96 as a null control.

### 3. `EagerFlush_SiblingTie_LateDiscriminatorSelectsTheRightIntent` is vacuous · CONFIRMED by execution (2 angles)

`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs`. It injects the complete utterance in one result:

```csharp
_recogniser.InjectText("set alpha level on");
_recogniser.FlushPendingBuffer();
```

`set_level` matches all four elements at 1.0 while `set_mode` misses a required literal, so `CompareCandidate` returns `Worse`, not `Tied`. No rival is recorded, the gate returns `Commit`, the command fires inside `InjectText`, and `FlushPendingBuffer()` is a no-op. **The test passes unchanged on `main`.**

Its comment claims the opposite — *"the window the refusal buys is where the missing word can still land … the flush selects on evidence the eager scan never had"* — and that sequence is exercised by no test in the change, so the half of DR-5 that justifies paying the latency is unpinned. This is the fixture trap `architecture.md` §8 states in its own words (*"a fixture that cannot reach the new condition cannot evidence it"*).

Note for the fix: `HandleResult` only **appends** to the buffer, so a two-result fixture must be one where a continuation can actually reach the discriminator.

### 4. #90's retention re-opens a case issue #91 records as fixed — wrong element number · CONFIRMED by execution

`VoxrCommandParser.cs:1044` (gate) / `:1062-1080` (collapse). `architecture.md` §6.4 accepts "two warnings for one hazard" as the cost of retention. The accepted package is larger: the second warning also carries a **wrong element number**, because once `kept` stops being value-normalised the two frames hold different member lists, `IndexOfSameMembers` returns `-1`, and the longest-frame preference at `:1073` never fires.

Executed on §6.4's own fixture at both revisions, both sets clearing the reachability gate:

```
base  el3  frame=[set|?now|*|on]  → "…"set ?now mode on" and "set ?now level on"
                                     that differ only at element 3"                    ✓
head  el2  frame=[set|*|on]       → "…"set ?now mode on" (with its optional elements
                                     omitted), … and "set mode on"
                                     that differ only at element 2"                    ✗
                                     element 2 of "set ?now mode on" is "?now"
```

That branch's retained comment states the rule it exists to enforce: *"Keeping the first-seen instead would report the discriminator's position within a form that silently dropped an optional, so an author counting elements in their own pattern would land on the wrong word."* Issue #91 explicitly carves this shape out as already handled (*"The longest-frame collapse fixes the symmetric case"*), so #91's deferral does not cover it.

### 5. The `FindSiblingSets` speedup does not reproduce across harnesses · DISPUTED — recommend withdrawing the claim

PR body: *"`FindSiblingSets` ~0.077 ms → ~0.038 ms — #90's gate replaced a per-member string compare with an index compare"*; `architecture.md:440` adds *"~2× faster … real and reproducible across four paired runs"*.

Two independent harnesses disagree on both the absolute and the existence of a delta:

```
review harness (6 interleaved passes):   before 0.0104–0.0116 ms   after 0.0105–0.0134 ms   (no delta)
author harness (6 interleaved passes):   before 0.0618–0.0743 ms   after 0.0373–0.0525 ms   (~40% drop)
```

The **causal story is wrong regardless of which harness is better**: the gate did replace a string compare with two int compares, but `DistinctValueCount` (`:1115`) reinstates the same `string.Equals` scan one line later, over a *larger* member list. The string work moved; it did not vanish. A residual mechanism exists for multi-member buckets, but it is a different claim from the one published.

A figure that two harnesses cannot agree on should not be a headline result, and the feature does not need it. Recommend dropping the row and its causal sentence.

---

## Minor

### 6. The sibling lookup is built even when it can never be consulted · measured

`VoxrCommandParser.cs:2904`. `TryEagerCommit` opens `EnsureCanCommitEarly(); EnsureSiblingLookup();` but short-circuits 168 lines later on `if (_canCommitEarly == null) return HoldExtendable`. With an abandoned precompute no path can reach the sibling condition, so the lookup is dead work for the parser's lifetime. Measured on a grammar forcing the abandon (50 families × 2 patterns × 3 optionals + one 13-optional pattern), first `TryEagerCommit` on a fresh parser:

```
base  0.0349 ms /       219 B        head  1.2034 ms / 1,195,795 B     verdict HoldExtendable both sides
```

Guarding the call with `if (_canCommitEarly != null)` changes nothing observable — `AreSiblingRivals` already returns `false` on a null lookup, and the Editor's other callers build it independently.

### 7. New CS0649 in every downstream player build · reproduced by 2 angles

`VoxrCommandParser.cs:268`. `List<SiblingSet> _siblingSets;` is declared unconditionally but assigned only under `#if UNITY_EDITOR`. Building the head worktree without `UNITY_EDITOR` (rig `NoWarn` reduced) emits:

```
VoxrCommandParser.cs(268,26): warning CS0649: Field 'VoxrCommandParser._siblingSets'
is never assigned to, and will always have its default value null
```

`f482154` emits none for this type, and `Runtime/Jinwoo1601.VoXR.Runtime.asmdef` sets no `noWarn` — so every consumer's Quest build gains a warning from a shipped package. Harmless at runtime (the sole reader's only call site is `[Conditional]`-elided).

### 8. Only one operand of the truncation check is exercised

`VoxrCommandParser.cs:1453` — `_siblingFormsTruncated[ci1][pi1] || _siblingFormsTruncated[ci2][pi2]`. The one fixture registers the 7-optional pattern **first**, so it is always the incumbent (`ci1`); `||` short-circuits and the right operand never evaluates. Delete it and both suites stay green, while a truncated pattern arriving as the *rival* of an analysed incumbent would commit on an analysis never performed.

### 9. The per-pair intent test's second arm is unpinned

`VoxrCommandParser.cs:1433` — `ci1 == ci2 || string.Equals(_commands[ci1].Intent, _commands[ci2].Intent, …)`. Both Editor fixtures place the same-intent rival in the *same command*, exiting on `ci1 == ci2`; nothing registers two distinct commands sharing one intent string inside a cross-intent set. Separately, no Runtime test distinguishes per-pair from per-set: `TryEagerCommit_SameIntentSiblingTie_StillCommits` builds a single-intent set that the `IsSingleIntent` pre-filter removes, so it passes under either rule.

### 10. The truncation predicate is stated twice and can drift · 3 angles

`VoxrCommandParser.cs:1357` re-derives what `WarningForms` (`:707`) already decided. Verified aligned today, but invariant 12 holds only while they agree, and the agreement is asserted in prose. A change to how `WarningForms` bounds expansion leaves the runtime gate believing an expansion happened that did not — re-opening finding 1 of the previous cycle. `FindSiblingSets` already calls `WarningForms` per pattern and could report the truncation it observed.

### 11. Three timing figures are not reproducible from the branch

The PR presents steady/cold `TryEagerCommit` and `FindSiblingSets` figures as measurements. The staged rig's only timing mode times `Parse`; nothing in the rig, repo or `Planning~/` times `TryEagerCommit` or `FindSiblingSets`. The branch's own issue #93 says so — *"taken with throwaway probes … not reproducible from the branch"* — but the PR body's one-line summary of #93 drops that clause.

### 12. The review tally is unevaluable by a reader

*"11 angles, 5 findings adversarially verified: 4 confirmed, 1 refuted"* — `gh pr view 94` shows 0 comments and 0 reviews, and the only record is under `Planning~/`, which `.gitignore:44` excludes. Two of the five are described; the other three, including two whose confirmation is asserted, are named nowhere reachable.

### 13. `BuildSiblingWarning` still takes the parameter its caller just dropped

`VoxrCommandParser.cs:1472`, sole caller `:1224`. The comment three lines above the caller gives the reason to drop it — *"resolving those indices against a different array passed by a future caller would silently mis-quote patterns or throw inside the message builder"* — and `BuildSiblingWarning` **is** that message builder, still dereferencing `commands[members[i].CommandIndex]`.

### 14. Three comments cite the wrong requirement

`:1010`, `:1113`, `:1403` cite "requirements F8" meaning **item 1's** F8 (author-duplicated patterns). This feature's F8 is *"The refusal honours the configured `minScore`"*. Canon disambiguates as "item 1 requirements F8"; the file's practice pairs an ambiguous id with an issue number.

### 15. The truncation cliff has no author-facing report

`WarnOnExcessiveOptionalExpansion` exists precisely because the `MaxOptionalExpansion` consequence is knowable from the assets at construction. The new 6-cap consequence is knowable the same way and gets no message — an author whose 7-optional pattern silently changed eager behaviour has no construction-time signal.

### 16. Allocation, not time, is the dominant new cost, and it recurs

First-probe base→head on larger grammars: `17.11 ms / 120,152 B` → `20.82 ms / 1,962,406 B` (10 commands × 6 optionals). `VoxrCommandRecogniser` rebuilds the parser on every `SetActiveSets`/slot change, so "once per parser" is not once per session. The megabyte is one key `string` per (form, discriminator position) in `FindSiblingSets`.

---

- **The eager tie locals should collapse to a `bool`** (previous cycle) — design §5.3 requires recording *which pattern* tied; `architecture.md` §4.1 had already rejected it.
- **The `HoldExtendable` path shortens the window on a detected tie** — REFUTED twice over. `architecture.md:202` states the consequence in those exact words and a shipped test pins it; and on any *reachable* sibling tie the deciding word cannot arrive, because `bestHasUnmatchedRequiredTail` returns `None` first, so elements after the discriminator already matched and appended speech can never fill a passed position.
- **Loser-loser sibling pairs among tied candidates are never tested** — the shape was constructed and reaches, but is inert: removing the sibling pair leaves the same winner and the same `Commit`. What decides the intent is a *non-sibling* tie, which §5.3 routes to registration order deliberately. Refusing there would be protection by coincidence, and would have item 3 prompt with values belonging to a command that is not firing.
- **`AreSiblingRivals`' bounds guards are dead code** — unreachable, but they match `CanCommitEarly`'s stated convention ("route through the bounds-checked path rather than indexing raw"). Left alone by two independent angles.

## Verified clean

Selection is bit-identical: `CompareCandidate == Better` ⟺ base `IsBetterCandidate == true`, re-derived key by key against `f482154` including NaN and signed-zero paths. The `bestScore <= 0f` sentinel cannot yield `Tied`. Only `Commit → None` is introduced structurally. `EagerCommitVerdict`'s single external consumer handles the new `None` correctly. `ParseDiagnosticEntry`'s two new fields are safe for both readers. #90's emission *criterion* is unchanged (old `kept.Count < 2` ⟺ new `DistinctValueCount < 2`), so no set newly appears or disappears, and the demo-grammar pins hold unedited at 11 / 5 / 6. Both build configurations compile and produce identical corpus output across all 4014 prefix rows and 699 parse rows. `AreSiblingRivals` allocates 0 bytes; steady `TryEagerCommit` is unchanged within noise. `closes #90` is substantively correct — including the sharper same-intent-suppression variant — though the citation offered for it (*"fails on `main`"*) discriminates nothing, since every new refusal fixture fails on `main`; the with-DR-5-without-#90 counterfactual was run and does support it.

## Notes

- `JoinWith`'s `parts.Length == 0` arm is unreachable from all three callers.
- `kept.Count < 2` is subsumed by `DistinctValueCount(kept) < 2`.
- Three `.ToArray()` round-trips exist only to satisfy `JoinWith(string[])`; all three producers are now `List<string>`.
- Comment volume: +247 comment lines to +248 code lines (1.00) against the base file's 0.54; the truncation rationale is stated four times, the Editor-record rationale three.
- A comment block was inserted between `[Conditional]` and the method it decorates.
- Two object-initializer members and one nested `for` are indented against the surrounding style.
- `BuildSiblingWarning` can now name N intents but fewer than N patterns — readable, but worth a glance.
- `VoxrCommandRecogniser.cs:872` calls the eager rules "the two `TryEagerCommit` already enforces (#66, #70)"; there are now three.
- The source comment *"the word that would have decided may still be coming"* is inherited from design §5.8 and is apt for the trailing case, not the medial one this feature targets — see the second refutation above.

## Post-G2 product-doc sites (list is a deliverable, not a finding)

The branch's list has four gaps, marked ★:

| Site | False after merge |
|---|---|
| `Documentation~/scoring.md:177` ★ | *"an eager verdict always names the command the subsequent flush will actually fire"* — F15 restated this in source only |
| `Documentation~/scoring.md:276` ★ | same claim |
| `Documentation~/scoring.md:284`/`:293` ★ | the "verdict above `None` requires all of" list omits the sibling condition |
| `Documentation~/scoring.md:290` ★ | *"A medial miss is fine and still commits"* |
| `Documentation~/command-recognition.md:147-151`, `:356` | medial drop "commits EARLY with the wrong sibling" |
| `Documentation~/troubleshooting.md:91` | same claim |
| `KNOWN_LIMITATIONS.md:532-538` | *"Both paths guess; the medial one guesses sooner"* |
| `KNOWN_LIMITATIONS.md:560` | needs the eager-vs-warning `minScore` asymmetry |
| `CHANGELOG.md:14` ★ | `[Unreleased]` still says "same eager timing" and offers "diverge earlier" |
| `Documentation~/api/command-recogniser.md:16` ★ | eager flush now has a second refusal reason |

- **Unity test execution.** The claimed EditMode 128/128 and PlayMode 477/477 could not be run; the host project cannot be driven from a review agent. Test *existence* and *content* were audited — the diff adds exactly 4 EditMode and 26 PlayMode `[Test]` methods, arithmetically consistent with 124→128 and 451→477.
- **On-device Quest behaviour** — no device binding exists in this project.
- **`VoxrCommandRecogniser.cs` in the player configuration** — the rig stages only 9 Runtime files and does not include it, so no automated check compiles it without `UNITY_EDITOR`. Pre-existing.

jinwoo1601:
## Review findings addressed — now at `38ac145`

All sixteen findings from the review above are resolved. Suites re-verified: **EditMode 129/129, PlayMode 479/479**; player configuration builds with **no new CS0649**; the 699-corpus A/B is unchanged in both directions.

### Fixed in code

| Finding | Resolution |
|---|---|
| **1 — truncated arm violated F10 in player builds** | The arm no longer answers `true` blindly. It compares the two patterns' **required elements** — their all-optionals-omitted readings, which is the one reading that survives truncation, since DR-1 needs the discriminator to be a required literal and required elements appear in every expansion. Zero differing positions ⇒ duplicates ⇒ no refusal. **Partial by construction**: a relation existing only in a *mid* expansion is still missed. Verified: the 7-optional duplicate pair now returns `Commit`; genuine truncated siblings still return `None`. |
| **6 — lookup built when unreadable** | Guarded on `_canCommitEarly != null`. Measured 1.2 ms / 1.2 MB → 0.03 ms / 219 B on a grammar that abandons. |
| **7 — new CS0649** | `#pragma` at the declaration. Player build now emits only the five pre-existing `VoxrPendingCommand` warnings, matching base exactly. |
| **8 — only the left `\|\|` operand exercised** | `TryEagerCommit_TruncatedPatternAsTheRival_RefusesToo`. |
| **9 — per-pair intent second arm unpinned** | `LastParseDiagnostics_SameIntentAcrossTwoCommands_IsNotTheRecordedRival`. It had to be a *diagnostic* test: the eager verdict is `None` either way, so the arm is only observable in which rival gets recorded. |
| **10 — truncation predicate duplicated** | One `ExpansionTruncated`, shared by `WarningForms` and `EnsureSiblingLookup`. |
| **13 — `BuildSiblingWarning` kept the dropped parameter** | Now an instance method reading `_commands`. |
| **14 — wrong F8 citations** | Now "item 1's requirements F8". |
| Notes | `JoinWith` takes a `List` and drops its unreachable zero-arm; the subsumed `kept.Count < 2` disjunct removed; three `.ToArray()` round-trips gone; `VoxrCommandRecogniser`'s "the two conditions" comment corrected. |

### Test rewritten

**3 — `EagerFlush_SiblingTie_LateDiscriminatorSelectsTheRightIntent` was vacuous.** Deleted, not repaired: the scenario cannot occur. Reaching the sibling condition means #70 passed, so an element *after* the dropped word already matched, and `HandleResult` only appends — the word cannot land in a position the match has gone past. Replaced with the boundary that does occur.

That also corrected a rationale. Design §5.8's *"the missing word may still arrive"* is true of the **trailing** shape (which #70 refuses) and false of the **medial** shape this feature targets. What refusing buys is that item 3 can **ask**. The source comment now says so; §5.8 keeps its wording, since amending locked design is a G1 matter rather than a review outcome.

### Claims corrected

- **2 — "the unchanged `HoldExtendable` count of 96 is the direct check on placement"** — withdrawn. Your construction proved it: two deliberately-misplaced variants produced byte-identical corpus output. The actual guard is `TryEagerCommit_SiblingTieOnAnExtendablePattern_StillHoldsExtendable`; the 96 is a null control. Corrected in the PR body and `architecture.md` §9.3.
- **5 — the `FindSiblingSets` speedup** — withdrawn. Re-running the original harness interleaved still reproduced the drop, so the two harnesses disagree on both the absolute *and* whether a delta exists. A figure two harnesses cannot agree on is worth less than none, and the feature does not need it. The causal story was wrong regardless: `DistinctValueCount` reinstates the string comparison the gate was said to remove.
- **11 — non-reproducible timings** — the cold figure is withdrawn for the same reason #93 already records. The steady-state "unchanged" claim is kept, since your review independently corroborated it.
- **12 — unevaluable review tally** — the PR body now points at this review rather than asserting a count from a gitignored doc.

### Deferred with a record

- **4 — #90's retention re-opens the wrong element number.** Real, and #91's carve-out ("the longest-frame collapse fixes the symmetric case") is now false — [commented on #91](https://github.com/jinwoo1601/VoXR-Speech-Recognition/issues/91#issuecomment-5305621051) with the executed before/after. Item 3 rewrites that message anyway.
- **15 — the truncation cliff has no author-facing report.** For `KNOWN_LIMITATIONS.md` in the post-G2 pass; the narrowing in finding 1 shrinks the cliff considerably.
- **16 — allocation is the dominant new cost and recurs per rebuild.** Finding 6's fix removes the worst case; the remainder is recorded in the architecture doc.

### Refuted, deliberately not applied

Collapsing the eager tie locals to a `bool` (design §5.3 asks for *which pattern* tied), and testing loser-loser sibling pairs (the shape reaches but is inert — removing the pair leaves the same winner, so refusing would be protection by coincidence, and would have item 3 prompt with values from a command that is not firing).

### Doc list

Your post-G2 list was more complete than mine — `scoring.md` ×4, `CHANGELOG.md`, and `api/command-recogniser.md` are now on it.

jinwoo1601:
## G2 accepted — product docs written, ready to merge

Post-G2 close-out complete at `2426754`. Ten sites across five files corrected, two new limitation entries added.

| File | What changed |
|---|---|
| `Documentation~/command-recognition.md` | The one-word-hazard section and the eager-flush verdict list — the medial shape no longer commits early; new bullet for the ambiguity refusal |
| `Documentation~/scoring.md` | §3's eager/flush invariant restated; §6's intro, condition 5, and the `Commit`-additionally clause, which now lists the sibling condition |
| `Documentation~/troubleshooting.md` | The medial-drop paragraph |
| `Documentation~/api/command-recogniser.md` | `eagerFlushOnCompleteMatch` row |
| `KNOWN_LIMITATIONS.md` | The sibling entry updated; **two new entries** — the six-optional analysis bound, and the warning-versus-gate asymmetry below the default `minScore` |
| `CHANGELOG.md` | New `[Unreleased]` entry under **Changed**; item 1's entry corrected where this feature superseded it |

**The correction worth flagging.** The docs — and design §5.8 — sold the deferral as a chance for the dropped word to arrive. It is not: speech only appends, and for a medial drop the position that word would have occupied is already behind the match. What deferring buys is that the guess is made once, on a final transcript, which is where item 3 can ask. That's now stated plainly in three pages rather than left implied, and it came out of a review finding that was *refuted* — the finding was wrong, and checking it surfaced that the rationale underneath was.

**Verification.** EditMode 129/129, PlayMode 479/479 at `38ac145`; docs-only since. All 50 `DocCheck` claims still pass against the real parser, including both `scoring.md` §6 eager-verdict pins.

**Still open by design:** #74 (item 3 delivers the disambiguation this defers to), #91 (widened — its longest-frame carve-out no longer holds), #93 (edge-branch coverage and two lookup invariants). #90 closes with this merge.

Ready for merge whenever you are — I won't merge it.

Issue #96, opened 2026-08-16
Entering a confirmation pending also raises OnUnrecognisedSpeech when it is the only result


jinwoo1601:
## What

In `ProcessParsedResultsCore`'s Step 7, the `RequiresConfirmation` arm enters pending and `continue`s **without** setting `anyThresholdFiltered`. If that command was the utterance's only result, the loop ends with `acceptedCount == 0` and `anyThresholdFiltered` still clear, so the tail raises `OnUnrecognisedSpeech(text)`.

The integrator is told the speech was not understood in the same frame they were asked to prompt for a confirmation about it.

## Repro

- One command with `requiresConfirmation: true`.
- Say it.
- `OnCommandPending` raises, and `OnUnrecognisedSpeech` raises for the same utterance.

`Runtime/Commands/VoxrCommandRecogniser.cs` — the `RequiresConfirmation` arm, and the `acceptedCount == 0` tail.

## Why it is filed rather than fixed

Found while building `feat-disambiguation-pending` (#74 item 3), whose new disambiguation arm has the identical shape one branch above. That arm **does** set `anyThresholdFiltered`, because raising "not understood" alongside "which did you mean?" is the specific confusion the feature exists to prevent.

The confirmation arm is pre-existing behaviour on a path that feature does not own, so it was left alone under the surgical-change rule rather than fixed in passing.

## Fix

Set `anyThresholdFiltered = true` in the `RequiresConfirmation` arm, matching the confidence, debounce and disambiguation arms. Needs a test asserting `OnUnrecognisedSpeech` does not raise when a confirmation pending is the only outcome — and a check of whether any integrator is relying on the current behaviour as a "nothing fired" signal.

Issue #95, opened 2026-08-16
The parse diagnostic records only sibling ties, so a non-sibling tie is indistinguishable from no tie


jinwoo1601:
Design §5.3 of `sibling-tie-disambiguation` promises that restricting the *action* to sibling rivals keeps non-sibling ties out of the runtime paths **"while leaving them visible to the editor diagnostic"**. That second half was never built, and after backlog item 3 it is the only part of §5.3 still outstanding.

## What is true today

`ParseDiagnosticEntry.TiedSiblingIntent` / `.TiedSiblingPatternIndex` (`Runtime/Commands/VoxrCommandParser.cs`) are written only when `AreSiblingRivals` answers true. So all of these leave the record null, indistinguishable from "nothing tied":

- a tie between two patterns that are not siblings at all (authoring error);
- a tie between two members carrying the **same** discriminating value (duplicates);
- a tie inside a wholly same-intent sibling set.

No other Editor surface reports two unrelated patterns tying at selection. An author whose grammar contains a genuine duplicate-pattern hazard sees a healthy-looking diagnostic.

## Why it was left

Item 2 (#90, PR #94) built the record sibling-only per design §9 row 2, which says the loops record a tied *sibling* rival. §9 row 2 and §5.3 are two prose sentences that drifted apart; neither is a ratified decision record (DR-1..DR-7 are silent on what the diagnostic stores). The human ruled 2026-08-16 to keep the build as-is, on the grounds that the field was inert and item 3 was its only prospective consumer.

Item 3 (`feat-disambiguation-pending`) is now built, and the ruling was revisited there: the **runtime** record must stay sibling-only, because DR-4 makes the discriminating values the choice vocabulary — a non-sibling tie has no discriminating values, so there is nothing the speaker could be asked. The Editor record was kept identical to the runtime one rather than widened, so that one concept has one shape.

## What closing this needs

Not a relaxed gate. It needs the record to distinguish **"a rival tied"** from **"a *sibling* rival tied"** — a second field, or a small enum on the existing one — plus the message/inspector surface that renders the difference, plus updates to the two tests that currently assert null for a non-sibling tie.

## Severity

Low. Editor-only, diagnostic-only, no runtime consequence in either configuration of `disambiguateSiblingTies`. Filed so that §5.3's unimplemented half is on the board rather than lost in a gitignored design doc.

PR #97, opened 2026-08-16
Ask the speaker which sibling command was meant (#74)

### LINKED ISSUE #74: Design: route an ambiguous sibling tie to confirmation instead of firing the first-registered pattern

### LINKED ISSUE #91: Sibling warning's element number can index a different word than the pattern it quotes

jinwoo1601:
Backlog item 3 of the G1-locked sibling-tie design. **This is the item where the tie stops being deferred and starts being answered.**

Closes #91. Filed on the way: #95, #96.

**Deliberately does not close #74.** That issue carries the whole G1-locked design, and its §9 backlog item 4 — `feat-disambiguation-docs`, the guide-level narrative spanning all three items plus the `KNOWN_LIMITATIONS.md` rewrite — is not built here. Closing it on merge would retire the only tracker for work that has not happened.

## What changes

Two patterns identical but for one required literal — `["set","{ship}","mode","on"]` and `["set","{ship}","level","on"]` — are siblings. When the decoder drops that one word, the surviving evidence fits both *exactly* equally: same score, same consumed span, same literal count. Selection exhausts every key it has and falls through to registration order, so the intent is a coin flip.

Item 1 warned the author at construction. Item 2 made the parser see the tie and made the eager gate refuse. Neither changed the flush, which still fired the first-registered sibling — so item 2 bought a pause and spent it on nothing. **This makes the flush ask.**

With `disambiguateSiblingTies` on, a flush winner carrying tied sibling rivals enters pending with the new `AwaitingDisambiguation` reason and raises `OnCommandPending`. The speaker says `mode` or `level` — already in the decoder's grammar, because they are pattern literals — and the right intent fires. Cancel works; "yes" does not, and deliberately does not abandon the question either. On timeout nothing fires, even under `FireAsIs` (DR-6).

One exception to the headline, stated rather than buried: when a rival's sibling analysis was **truncated** (a pattern past `MaxWarningExpansion`), it ties provably but cannot be *named*, so no question can be phrased about it. If it is the only rival the flush fires the winner as it would with the flag off; if another rival makes a question happen, it is absent from the choices and `IsTruncated` says so.

Off by default. With the flag clear, **parse and fire behaviour** is byte-identical to `337e758` — same intents, slots, scores, events and timing. Three Editor-only *messages* do change regardless of the flag, which is the carve-out requirements F2 grants explicitly (*"no edited expected values except the tests F13–F15 explicitly change"*): #91's element numbers, F15's flag clause, and F13's narrowing of the cancel-collision report.

## Measurements (F18)

Measured **interleaved**, before `337e758` → after `d5aaa66`, rig validated against the committed grammar pin first.

| # | Figure | Result |
|---|---|---|
| a | 699-corpus `Parse` A/B, flag off | **699/699 rows identical** — no change to intents, slots or scores |
| b | Disambiguation-flow fixtures | **29 tests**; 10 drive the full two-turn exchange, medial *and* trailing discriminators |
| c | Construction warning volume | **unchanged: 2** on the demo grammar, 0 collisions; text changed by F15 only |
| d | Per-parse cost, flag off | **below the harness's noise floor**; allocation **exactly unchanged at 492.8 bytes/utterance** |

**(a) is cited as a control, and what it is blind to is part of the number.** The corpus is single-utterance and contains no follow-up answers, so it is structurally incapable of exercising a two-turn exchange. It is evidence for F2 — the flag-off path did not move — and for nothing else.

**(d), stated honestly.** Twelve interleaved rounds of 1,398,000 parses: before 19.362 µs mean (18.92–20.05), after 19.442 µs (19.14–20.11). Mean paired delta +0.08 µs, but the delta changes sign 4 times in 12 and spans −0.92 to +0.76 µs. **No point estimate is claimed** — the harness cannot resolve it. Allocation it resolves exactly, and that is the claim that matters: a preallocated record adds none.

Also pinned: the generated grammar JSON is **identical across the flag** (1531 bytes). F7's claim that discriminating values need no grammar addition was *reasoned*, and `GetFollowUpGrammarWords` is not flag-aware — if the reasoning were wrong the flag would change what VOSK can hear in both configurations, the one way F2 could break that (a) cannot localise.

## Verification

- Unity **EditMode 130/130**, **PlayMode 519/519** (Unity 6000.4.7f1, host `VoXR TestGround`).
- A/B rig **player configuration** (the only build here without `UNITY_EDITOR`) compiles clean with `<NoWarn>` stripped.
- **`PlayerCheck.cs` 6/6** — new, and the durable half of this PR. See below.
- On-device Quest verification is human-only and **has not been run**.

## The review found a crash the 519 tests could not

`review-cycle` ran 11 finder angles and 3 adversarial verifiers. Five defects survived; all are fixed in `d5aaa66`.

The one worth reading: the rival buffers are indexed by `_resultCount`, but `ParseInternal` tested "result buffer full" at the *bottom* of the extraction round. A parse that filled the result buffer ran one more full scan with `_resultCount == _resultBuf.Length` and wrote exactly one slab past the end — `IndexOutOfRangeException`, on `"set alpha on set alpha on set alpha on"` against a two-command sibling grammar. **Live in the Editor regardless of the flag.** Until this feature the round's tie state was two `int` locals, so the extra round was harmless and the guard's position did not matter.

The others: Step 7 read `_parser` live where the loop deliberately snapshots it (NRE if a subscriber reconfigures from an event — with one result, not several); the debounce gate tested only the winner while any rival can fire; `Truncated` was set by the cap alone though two other paths drop answerable rivals; and the Editor diagnostic went null on a truncated sibling tie, where the coin flip is still real.

One claim was **refuted** and no change made: the matching `minConfidence` leak cannot occur, because a rival's span differs from the winner's only by `[unk]`, which `ComputeConfidence` skips.

**`PlayerCheck.cs`** is the response to the standing hazard. This design has twice hit "a runtime path gated on a symbol every test defines" — Unity always defines `UNITY_EDITOR`, so no Unity test can see it. Six assertions run in the A/B rig's player configuration: the flag gates recording, a flag-on player is *not* inert, both discriminating words survive, the overrun stays fixed, and the n-ary record and its cap behave.

**It is not part of this diff, and that is worth being plain about.** It lives in the A/B rig under `Planning~/`, which is gitignored, so like the rest of the rig it is local-only — the six files above are the whole merge. The PR review caught an earlier draft of this section calling it "the durable half of this PR"; it also caught that `stage.sh` did not copy it, so the documented invocation could not build and the first 6/6 came from a hand-copy. `stage.sh` now stages it, and the run above is from a `stage.sh`-produced tree.

## What is deliberately not here

- **#96**, filed not fixed: the pre-existing `RequiresConfirmation` arm has the same shape as the new branch and does not set `anyThresholdFiltered`, so a confirmation pending that is an utterance's only result also raises `OnUnrecognisedSpeech`. Pre-existing, on a path this feature does not own. The new branch *does* set it.
- **#95**, filed not built: design §5.3's promise that non-sibling ties stay visible to the Editor diagnostic. Forced, not chosen — DR-4 makes the discriminating values the choice vocabulary, and a non-sibling tie has none, so there is no question to ask.
- **#93** untouched. The eager path's *condition* untouched (F17). No scoring arithmetic. `sibling-tie-disambiguation.md` unedited (F16) — design §5.8's rationale is wrong for the medial shape, and by ruling that correction lives in source comments and the post-G2 product docs, not in the locked design.
- **Product docs.** Written after the G2 ruling, before merge, per the workflow.

## Two questions carried to G2, not decided here

1. **Is `OnCommandConfirmed` the right event for a disambiguation answer?** Resolving a choice goes through `PendingResolution.Confirmed`, so it raises before `OnCommandRecognised`. Consistent — but that event is documented in terms of `requiresConfirmation`, and an integrator may read it as "the user approved a destructive action".
2. **Precedence when a discriminating value is also a whole command.** The choice check runs before the parse, so while a question is on the table the answer wins over a standalone command of the same word. Probably right; not discussed in canon. It does not occur in the demo grammar.


jinwoo1601 (commented):

**Head reviewed:** `d5aaa669512f22b4bfd5fb444bf0115ffd39c285` · **base (merge-base with `main`):** `337e758`
**Angles run:** `CLAIM` + all 11 of the review-cycle roster (LINE, GONE, XFILE, PIT, WRAP, REUSE, SIMP, EFF, ALT, CONV, GAP). None dropped, none added.

The finder agents inherited nothing from the session that wrote the code. A `review-cycle` pass already ran on this branch and its five findings are fixed in the head commit — those were declared out of scope for this review, so what follows is what that cycle missed, **including three defects introduced by its own fixes**.

Fourteen findings. Three were adversarially verified with executable repros against the head worktree; all three came back CONFIRMED. Nothing was refuted this round.

---

### 1 · The rival cooldown gate degrades the flag-on path to the coin flip it exists to prevent · moderate · CONFIRMED

`Runtime/Commands/VoxrCommandRecogniser.cs:1100-1107`. Introduced by the prior review cycle to fix a real finding — Step 7's debounce tests only the tie's *winner*, while any rival can fire. The fix is at the wrong place and is worse than the disease.

```csharp
if (commandCooldown > 0f
    && _debouncer.IsOnCooldown(parser.RivalIntent(i, n), now, commandCooldown))
{ truncated = true; continue; }
```

On a **two-way** set this drops the only rival, `if (_choiceBuf.Count < 2) return false;` (`:1116`) falls through, and the **winner fires** — with `truncated` discarded on the `false` return, so no signal survives. `anyThresholdFiltered` and the Editor `attempts.Add` are inside the same short-circuited block, so it logs as an ordinary accepted fire.

And the exclusion is **provably always wrong at shipped defaults**. `commandCooldown = 0.3f`, `bufferWindow = 0.5f`, `pendingTimeout = 5.0f`. Eager flush is skipped while pending (`:574`), so an answer always waits the full buffer window: the earliest it can fire is `now + 0.5`, while exclusion requires `now − lastFire < 0.3`. The cooldown has certainly expired before the rival could have fired. The premise the gate was built on is unreachable at shipped settings — which is why the only test of it sets `CommandCooldown = 30f`, 100× the default.

The bias also runs backwards: the intent on cooldown is the one the speaker *just* used, so it is arguably the **most** likely one they meant.

`Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:1694-1720` pins the harm as intended behaviour. The speaker says "set alpha **level** on", then a corrupted "set alpha on", and the package fires **`set_mode`** — a command they never uttered — *because* they just used `level`.

The counter-argument was weighed and does not save it: the pre-existing `RequiresConfirmation` arm has the same ask-time staleness, which *supports* gating at ask time. But when the confirm path's intent is on cooldown, **nothing fires**; when the disambiguation rival is on cooldown on a two-way set, **the sibling fires**. The gate borrowed the rule and inverted its outcome.

**Recommended:** remove the gate. A deliberate answer to a question the recogniser asked is not the duplicate-VOSK-result fire `CommandDebouncer` exists to suppress, and the confirmation path already accepts exactly this. Delete the test with it — it pins the wrong behaviour — and state the property in the docs instead.

---

### 2 · A truncated sibling rival is dropped without setting the truncation flag, when a question *is* asked · moderate · CONFIRMED

`Runtime/Commands/VoxrCommandParser.cs:2247-2262`. `!nameable` falls through both statements, so `tiedTruncated` is never set. The in-source defence — *"the integrator's remedy … is what happens anyway when no question is asked at all"* — holds only when the truncated rival is the **sole** tie. A question is asked whenever some *other* rival is nameable.

Reproduced at head, no `UNITY_EDITOR`:

```
C1  W + A(truncated)         RivalCount=0  Truncated=False   -> no question asked, fine
C2  W + A(truncated) + B     RivalCount=1  Truncated=False   rival[0]=standby
```

C2's grammar is `set {ship} mode on` / `set {ship} level on ?p ?q ?r ?s ?t ?u ?v` / `set {ship} standby on` on `set alpha on`. A question is asked, `set_level` is **absent** from `ChoiceValues`, and `IsTruncated` reads **false**. Saying `level` matches nothing, and DR-6 then fires nothing at timeout — while `TiedSiblingIntent` names `set_level` as the coin-flip rival in the same parse whose choice list omits it.

Architecture §2.8 states the rule this breaks: *"It reads **an answerable rival was dropped**, not 'the cap was hit'."* The truncated arm is a fourth such path and does not set it. Nothing warns at construction either — 7 optionals is `> MaxWarningExpansion` (6) but `<= MaxOptionalExpansion`, so `WarnOnExcessiveOptionalExpansion` is silent too.

**Recommended:** set `tiedTruncated` on the `!nameable` path once any rival has been recorded.

---

### 3 · Two patterns of the *same intent* are offered as separate choices · moderate · CONFIRMED

`Runtime/Commands/VoxrCommandParser.cs:2380-2390`. The dedup compares `Value` only. `IsAnswerableRival` — the predicate F13 requires both consumers to share — is applied winner-vs-rival only, never rival-vs-rival.

Reproduced at head with `alpha:[set {ship} mode on]`, `beta:[set {ship} level on],[set {ship} report on]` on `set alpha on`:

```
RivalCount=2  WinnerValue=mode
  rival[0] value="level"   intent="beta"
  rival[1] value="report"  intent="beta"
```

One `VoxrCommandDefinition` with two patterns is enough. The speaker is asked to choose between `level` and `report`, which dispatch the **identical command** — the "round trip to choose between identical outcomes" that `IsAnswerableRival`'s intent half exists to prevent.

The cap harm was reproduced too — `beta` with four sibling patterns plus `gamma:[set {ship} standby on]`:

```
RivalCount=4  Truncated=True
  level/beta, report/beta, trim/beta, gain/beta     <- standby/gamma never offered
```

All four slots go to one intent and the only genuinely different alternative is squeezed out. F19's *letter* holds (`Truncated` fires); its point does not. Latent rather than live: the demo grammar's five cross-intent sets are all 2-member — but it records six same-intent sets including a 4-member one, so one cross-intent member joining it reproduces this.

**Recommended:** reject a rival whose intent already appears among the recorded rivals, alongside the value test.

---

### 4 · `PlayerCheck.cs` is neither in this PR nor reachable by the documented command · moderate · verified directly

The PR body calls it *"new, and the durable half of this PR"*. Both halves are false.

- It is **not in the PR**: `git show d5aaa66 --name-only` lists six files, none of them `PlayerCheck.cs`. It lives under gitignored `Planning~/`, so `git ls-files` returns nothing for it.
- It is **not staged**: `stage.sh:39` copies a fixed list — `UnityStub.cs`, `Program.cs`, `Check.cs`, `DocCheck.cs`, `Rig.csproj` — and `Rig.csproj` compiles `*.cs` in the *output* directory. So the documented `-p:StartupObject=AbRig.PlayerCheck` cannot build from a `stage.sh`-produced tree. The recorded 6/6 came from a hand-copy.

The check's substance is sound — `RivalCount == 1` in a no-`UNITY_EDITOR` build does prove `EnsureSiblingLookup()` ran outside the Editor gate. But as wired, the guard against a failure class this design has already hit **twice** will not fire again unless someone remembers both to copy the file and to override `StartupObject`.

**Recommended:** add `PlayerCheck.cs` to `stage.sh`'s `cp` list and update `Rig.csproj`'s "two entry points" comment; and correct the PR body, since the file genuinely is not part of what gets merged.

---

### 5 · The crash the prior cycle fixed has no Unity test · minor · high confidence

`Runtime/Commands/VoxrCommandParser.cs:2085`. `_tiedSiblingBuf[_resultCount] = …` (`:2303`) is written **unconditionally** — not under `_recordSiblingTies`, not under `#if UNITY_EDITOR` — so the overrun the moved guard prevents is not player-specific: without it, an utterance with more extraction rounds than `_resultBuf.Length` throws in the Editor, in a player, and with the flag off. Its only regression test is in `PlayerCheck.cs`, which per finding 4 does not run. No Unity test constructs the shape; the longest fixtures are two-round.

**Recommended:** an EditMode/PlayMode test parsing more rounds than registered commands (`"set alpha on set alpha on set alpha on"` against a two-command grammar), leaving `PlayerCheck` to carry only genuinely player-only claims.

---

### 6 · `VoxrPendingAmbiguity.IsTruncated`'s public doc is false for three of its four producers · minor · high confidence

`Runtime/Commands/VoxrPendingAmbiguity.cs:52-57` still reads *"True when the sibling set held more values than the runtime offers as choices, so the remaining intents are reachable only by saying the whole command again."* The flag now also fires for the set-coherence rejection (a *second, different question*), a rival with no definition (not fireable at all), and a rival on cooldown (temporary — re-uttering hits it again). Architecture §2.8 anticipated the drift and widened the *code*; the public doc kept the cap's wording **and** the remedy the widening invalidates. §2.8/§8b also enumerate three paths where the code has four.

---

### 7 · "Byte-identical with the flag clear" overclaims · minor · high confidence

`Runtime/Commands/VoxrCommandParser.cs:1489-1492`. In the Editor, and independently of the flag, the collision report gained `if (!HasAnswerableCoMember(set, m)) continue;` and now tests `_effectiveCancelVocabulary` instead of `DefaultCancel`. This PR's own `SiblingWarning_SameIntentDiscriminator_ReportsNoCancelCollision` asserts `NoUnexpectedReceived()` on a grammar that emitted a warning at base. Message text and element numbers also move (F15, #91). Requirements F2 carries the carve-out — *"no edited expected values except the tests F13–F15 explicitly change"* — the PR body drops it.

---

### 8 · The headline claim has two unstated exceptions · minor · high confidence

*"a flush winner carrying tied sibling rivals enters pending … and raises `OnCommandPending`"* — two shipped paths do the opposite and still coin-flip silently: a rival on cooldown (finding 1) and a truncated sibling analysis (`VoxrCommandParserTests.cs:4941`, *"a tie we cannot phrase a question about is not one we ask about"*). In both the integrator gets no `OnCommandPending` and no signal that registration order decided. Neither appears in the body's "What is deliberately not here".

---

### 9 · `Closes #74` retires the only tracker for unbuilt work · minor · medium confidence

#74's §9 backlog item 4, `feat-disambiguation-docs` — the guide-level narrative spanning all three items, plus the `KNOWN_LIMITATIONS.md` rewrite — is explicitly scoped beyond this feature's own post-G2 docs and is not built here. Merging with `Closes #74` retires it. (#91 by contrast is fully delivered.) **Recommended:** file a successor issue for item 4 before merge, or leave #74 open.

---

### 10–14 · Smaller, all high confidence

10. **Duplicated header comment** — `VoxrCommandRecogniser.cs:1013-1035`. "Builds the choice list for result i…" and the whole six-line "Allocates —" paragraph each appear **twice**, verbatim; only the middle `parser` paragraph is unique. A merge artefact of the review-fix commit. Reported independently by eight of twelve angles.
11. **Rival buffers allocated regardless of the flag** — `VoxrCommandParser.cs:612-616`. The comment concedes it is "purely for constructor simplicity". `_rivalSlotBuf` alone is `commands × 4 × maxSlots × 16` bytes — ~75 KB on a 200-command grammar — re-allocated on every `NotifySlotChanged()` and never read when the flag is off. DR-7's "costs a flag-off player nothing" holds per-*parse*; per-*rebuild* it does not, and claim (d) measures per-utterance allocation, which structurally cannot observe it.
12. **Sibling analysis recomputed on slot-only rebuilds** — `VoxrCommandParser.cs:643`. `FindSiblingSets` is a pure function of the *commands*, but `RebuildParser` passes the same `_activeCommands` with only slots changed, so `NotifySlotChanged()` re-runs the whole expansion walk. The file's own note measures a comparable build at "1.2 ms and ~1.2 MB".
13. **The two-set test never asserts the thing it describes** — `VoxrCommandParserTests.cs:4871` comments *"it is stated rather than hidden"* but asserts only `RivalCount`, `WinnerValue` and the rival's value/intent, never `record.Truncated`. Hard-coding that assignment away leaves the suite green.
14. **`PlayerCheck.cs`'s own comment overclaims** — it says it would notice "if the recogniser stopped passing the flag through", but `stage.sh`'s `SOURCES` does not include `VoxrCommandRecogniser.cs`, so that plumbing is never compiled in the player configuration.

---

### Verified clean — worth recording

- **GONE returned no findings.** The buffer-full guard move is behaviour-equivalent: at base the extra round appended neither a result nor a diagnostic entry, and its only writes were round-local.
- **F17 holds.** `AreSiblingRivals` → `TryFindSiblingRival` is answer-identical: the added intent conjunct in `IsAnswerableRival` is already forced true by the unconditional intent guard above it, and the truncated arm and `TryEagerCommit`'s call site are byte-identical to base.
- **`BuildRivalCommand` is sound.** Intent, slots, pattern index and confidence-span end all come from the rival; taking `Score` and `RawText` from the winner is legitimate, since `Tied` requires exact score equality and `RawText` is the whole utterance.
- **No test lines were deleted anywhere in the diff** — no coverage was dropped to reach green.
- **No mutable-struct-copy bugs**, the `.meta` GUID is unique and correctly formatted, and the parser's new ctor params are appended so `VoxrBatchTestRunner`'s two sites compile untouched.
- Claim 7 (29 tests / 10 two-turn), 12 (130 + 519), 16 and 17 all check out against the diff.

### Not verifiable from here

Unity EditMode/PlayMode counts and the rig measurements were re-run by this session but not independently reproduced by the finder agents — they can only check the diff is consistent with them, which it is. **On-device Quest verification has not been run** and is correctly not claimed.

### Notes — unverified, one line each

- `TiedSiblingRecord.SetId` is write-only in shipped code; the set identity is only ever needed round-locally.
- `RecordTiedRival`'s four `ref` parameters are the four fields of `TiedSiblingRecord` — the round's tie state wants to be that struct.
- `SiblingSet.DiscriminatorIndex` lost its last production reader in this PR; only tests read it now.
- `RivalIntent`/`BuildRivalCommand` restate the flat-index arithmetic the `TiedRival` accessor one method above already expresses.
- The cancel-vocabulary fallback rule is written twice, in two files, with no shared symbol — the comment claims "one rule".
- `if (nameable && !sameQuestion)` immediately followed by `if (nameable && sameQuestion)` is an if/else wearing two ifs, in the hot loop.
- `_choiceBuf`/`_choiceValueBuf`/`_choiceDefBuf` are declared ~950 lines below the file's other instance fields, and retain references after the question resolves.
- `VoxrCommandRecogniser.cs`'s `Depends:` header does not list `VoxrPendingAmbiguity`, though the parser's header was correctly updated in the same PR.
- `Complete`'s comment says "the two existing callers"; there is one.
- `PendingAmbiguity` guards on `Choices` while `TryHandleConfirmCancel` guards on `ChoiceValues`.
- `VoxrCommandParser.cs:1451` still says the constructor "is handed only (slots, commands, coverageWeight, additionalGrammarWords)" — it now takes six.
- A chosen rival never passes the flush path's `IsIncomplete`/#73 completeness gate; only definition-lookup and cooldown are re-tested.
- Re-entrant `InjectText` from an event handler overwrites the parser's pooled buffers mid-loop — pre-existing in shape, but a corrupted read now reaches `EnterPending` with choices rather than only mis-firing.
- `Rig.csproj`'s comment still names two entry points; there are four `Main`s in that directory.

---

**Findings 1–4 are the ones that bear on G2.** 1, 2 and 3 are behavioural and all three were reproduced; 4 means the PR's stated durable safeguard is not part of the merge. Nothing here is merged or approved by this review — the ruling is the human's.

jinwoo1601:
Pushed `df87929` — the four findings ruled fix-now are addressed. Three of them were introduced by the earlier `review-cycle`'s own fixes, which is the argument for having run both passes.

**1 · Rival cooldown gate reverted.** It degraded the flag-on two-way case to the coin flip: dropping the only rival left one choice, the branch fell through, and the winner fired. And at shipped defaults it could never have fired correctly anyway — an answer always waits out `bufferWindow` (0.5s) while exclusion requires being inside `commandCooldown` (0.3s), so the cooldown has always expired before a choice could fire. The test that pinned the wrong behaviour is replaced by `Disambiguation_AnswerFiresEvenWhenItsIntentIsOnCooldown`, which asserts the question is still asked and the answer still fires.

The principle was already settled elsewhere: the confirmation path enters pending after the debounce check and fires on confirm without re-checking. A deliberate answer to a question the recogniser asked is not the duplicate VOSK result the debouncer exists to suppress.

**2 · A truncated rival beside an answerable one now sets `Truncated`.** §2.4's ruling held only when the truncated rival was the *sole* tie; when another rival makes a question happen, it is an answer the speaker could have given that is missing from the list. Tracked separately and resolved at record-write time so arrival order cannot change the answer. Pinned by `TiedSiblingRecord_TruncatedRivalBesideAnAnswerableOne_ReportsTruncation`.

**3 · Rival-vs-rival dedup now uses `IsAnswerableRival`.** Two patterns of one intent no longer occupy two choice slots — which also stops four sibling patterns of a single intent consuming the whole cap and squeezing out the only genuinely different alternative. Pinned by `TiedSiblingRecord_TwoRivalsOfOneIntent_OfferOnlyTheFirst`.

**4 · `PlayerCheck.cs` is staged.** `stage.sh` now copies it and `Rig.csproj` lists the entry points, so `./stage.sh WORKTREE <out>` runs it with no hand-copy — 6/6 from a freshly staged tree. It remains local-only because `Planning~/` is gitignored, and the PR body now says that plainly instead of calling it "the durable half of this PR".

The PR body is also corrected for findings 7–9: the byte-identical claim now carries the Editor-message carve-out requirements F2 actually grants, the headline claim states its truncated-analysis exception, and `Closes #74` is removed so backlog item 4 keeps its tracker.

**Verification after the fixes:** Unity EditMode 130/130, PlayMode 521/521. Player-configuration checks 6/6.

Findings 5, 6 and 10–14 are left unfixed for your ruling — notably **5** (the overrun crash the earlier cycle fixed is not player-specific and still has no Unity test, only the rig check) and **6** (`IsTruncated`'s public doc describes one of its four producers).

jinwoo1601:
Pushed `f918de7` — every remaining finding and Note is now addressed, either fixed or filed.

**5 · The overrun crash has a Unity test.** `PlayerCheck` was the wrong home for it: the per-round record write is not behind `#if UNITY_EDITOR`, so an utterance with more extraction rounds than results throws in the Editor too, flag or no flag. `TiedSiblingRecord_MoreExtractionRoundsThanResults_DoesNotOverrun`.

**6 · `IsTruncated`'s doc rewritten.** It now says what the flag *means* — an intent that also matched is not on this list — with the causes in a `<remarks>`, since none of them is actionable by the speaker and the old text's remedy was false for three of the four.

**11 · The rival buffers are left null when nothing will record**, and the record write is under the same gate. DR-7's "costs a flag-off player nothing" held per parse but not per rebuild, where `_rivalSlotBuf` is tens of KB on a large grammar, remade on every `NotifySlotChanged()` and never read. `PlayerCheck` asserts the null directly — the Editor forces `_recordSiblingTies` true and structurally cannot see that branch.

**13 · The two-set test asserts the `Truncated` its comment promised.**

**14 · `PlayerCheck` no longer claims** to cover the recogniser plumbing that `stage.sh` does not stage.

**12 · Filed as #98, not fixed.** With the flag on, a slot-only `RebuildParser` recomputes a commands-only analysis (~1.2 ms measured). Real — but where the cache should live is a design question, not a cleanup-pass decision.

**Notes.** All addressed: the duplicated `TryBuildAmbiguity` header removed; its three pooled `List` fields are now locals (it allocates arrays at the end regardless, and the pooling only kept commands alive past the question); `TiedSiblingRecord.SetId` removed as write-only, the set id being round state rather than a property of the answer; the two mutually exclusive `if`s became `if/else`; `RivalIntent`/`BuildRivalCommand` use the `TiedRival` accessor instead of restating the buffer layout; `VoxrFollowUpVocabulary.Resolve` is now the single definition of the empty-override fallback, which the collision report's comment already claimed; `PendingAmbiguity` guards the same field `TryHandleConfirmCancel` does; the `Depends:` header lists `VoxrPendingAmbiguity`; `Complete`'s comment says one pre-existing caller, because there is one; and two unreachable-by-construction properties are now stated at their sites — a chosen rival is never completeness-checked, and re-entrant `InjectText` from an event handler is unsupported.

`SiblingSet.DiscriminatorIndex` is left as-is: it lost its last production reader when both messages moved to the per-member authored index, but it is part of what a sibling set *is* and the tests read it.

**Verification:** Unity EditMode 130/130, PlayMode 522/522. Player-configuration checks 7/7 from a `stage.sh`-produced tree with no hand-copy.

One honest caveat carried forward from the run: the Editor forces `_recordSiblingTies` true, so the Unity suites do not exercise the new flag-off null branch at all. That branch is covered by `PlayerCheck`, which is exactly why it exists — and `PlayerCheck` is local-only, because `Planning~/` is gitignored.

jinwoo1601:
Product docs pushed as `9413b9f`. This is the post-G2 close-out step, so the PR now carries the full feature.

**What was written.** `api/data-types.md` gains a `VoxrPendingAmbiguity` section with a worked `OnCommandPending` handler; `api/command-recogniser.md` gains `disambiguateSiblingTies` and `PendingAmbiguity`; `command-recognition.md` gains **Ask instead of guessing**; `KNOWN_LIMITATIONS.md`'s sibling entry becomes a limitation *with a supported remedy*, while stating plainly that the limitation is what remains with the flag off — which is the default. `CHANGELOG.md` gets entries for #74 and #91.

**Scope held deliberately.** The guide-level narrative spanning items 1–3, and the full `KNOWN_LIMITATIONS` rewrite, belong to `feat-disambiguation-docs`. What is corrected here is what this feature made **false**, which is a different thing from what it made incomplete.

**F16 is discharged.** `command-recognition.md` used to say the deferral was worth paying for because "a future release can ask you which command you meant". It stops saying that. The paragraph exists to explain why deferring helps *at all* — the missing word can never arrive on a medial drop — and the answer was always "because that is where we could ask you". Now we do.

**Verifying the prose against source found four inaccuracies, three of them in what I had just written, and one of them a code gap:**

1. **`OnCommandConfirmed` does not always fire when a disambiguation is answered.** If the chosen command sets `requiresConfirmation`, `Complete` re-enters and raises `OnCommandPending` again — "which?" then "are you sure?". That is not an edge case: the guide's own remedy #2 tells authors to mark the destructive sibling `requiresConfirmation`, so taking both remedies together *is* this two-stage path, and nothing described it. Now documented in the guide, the API reference and the changelog.
2. **`OnUnrecognisedSpeech`'s row said "or a match was diverted to pending".** The new branch deliberately suppresses it, so that is now true of two diversions out of three. Corrected there and in `scoring.md`, whose "those two are the only filters that suppress it" this feature falsified.
3. **`cancelPendingOnRelease` is not a name an integrator can use** — the field is private; the Inspector label is **Cancel Pending On Release**.
4. **A code gap in the rival dedup**, from the fix two commits ago. Dropping a rival that shares an offered choice's *intent* but carries a *different word* was silent. Same word is genuinely nothing lost — saying it picks the choice already there. A different word reaching the same command is not: only one of the two is offered, so a speaker who says the other gets no match and the pending times out. The two cases are now distinguished and both directions are asserted.

**Verification:** Unity EditMode 130/130, PlayMode 522/522.

Still not run and not claimed: on-device Quest.

PR #99, opened 2026-08-16
Make the three sibling-tie features read as one story (#74 item 4)

jinwoo1601:
Backlog item 4 of the G1-locked sibling-tie design — the last one. **Documentation only; no code changed.**

## Why this is smaller than the backlog imagined

Item 4 was scoped as *"`command-recognition.md`, the `KNOWN_LIMITATIONS.md` entry rewritten from 'documented limitation' to 'documented limitation with a supported remedy', and `CHANGELOG.md`."* Item 3's own post-G2 close-out wrote all three — not to poach scope, but because leaving them saying "no remedy exists" would have been **false** the moment the flag shipped.

What was actually left is the part item 4 is really for: items 1–3 each documented themselves in isolation, so a reader met one hazard three times in three voices, and the authoritative reference still described the world before item 3.

## What two coherence reads found

The most damaging finding was in `scoring.md`, not in the new pages. Its `Commit` rule closed with *"the flush then fires the same command it always would have"* — the last word on the hazard, in the page the guide sends you to for the full model, telling a reader the thing is unresolvable. Two other pages had the same shape, including the `KNOWN_LIMITATIONS` entry aimed squarely at the reader running below the default `minScore`.

Contradictions resolved:

- **`scoring.md` forbade building behaviour on registration order** while three pages advised "register the safer one first". The reference now carves out the one defensive use and names it what it is.
- **The >6-optionals entry claimed the eager gate commits early** on such a pair. It does not — the runtime pair test falls back to required elements and refuses, which two tests pin. The same entry conflated two different mechanisms five lines apart.
- **`OnUnrecognisedSpeech` had two-filter and three-filter versions on the same page**, 43 lines apart.
- **The pipeline diagram put Pending Command Check before Debounce**, which is backwards.

**Four of the nine findings in the second read were damage the first pass's own fixes introduced**, including one factual error: I had written that the runtime path does not consult `minScore`. It does — the disambiguation branch sits *after* that gate, so a tie only reaches it if its winner cleared the **configured** threshold. That is what makes the advice right for that entry's low-threshold reader and wrong for anyone on the default.

## Discoverability

For a reader who knows neither "sibling" nor "disambiguate": the hub now describes `command-recognition.md` and `troubleshooting.md` in symptom terms; the section owning the fix is titled for the symptom; `scoring.md`'s session-log table gains the two rows a reader working from evidence needs, including the `rejectReason` string the runtime actually writes; and the flag's independence from `eagerFlushOnCompleteMatch` is stated, since eager flush is off by default and the narrative order implied a dependency.

Also: **"sibling" names two different hazards with opposite fixes**, in consecutive sections. A callout now tells them apart by the Console message each produces.

## Verification

- **No code changed** — every file in the diff is markdown, so the Unity suites are untouched and were not re-run. Last green at `53b0156`: EditMode 130/130, PlayMode 522/522.
- **All 217 intra-repo markdown links verified to resolve**, file and anchor. The section heading was renamed twice during the pass, so every inbound link was rewritten and re-checked.
- Every claim added about runtime behaviour was checked against source: the filter order and its rationale, the `rejectReason` shape, the flag's independence from eager flush, `pendingTimeout` continuing to run after a push-to-talk release, the debug window's reason string, and both Console message fragments the new callout quotes.

## Scope

Closes out the #74 backlog. #74 itself was closed at item 3's merge, so there is no issue to close here.


jinwoo1601 (commented):

**Head reviewed:** `2a247ba` · **base (merge-base with `main`):** `53b0156`
**Angles run:** CLAIM + all 11 roster angles. **Disclosure:** PIT, WRAP and EFF were run as one agent rather than three, because the diff contains no code — that agent's primary task was to *verify* that claim independently rather than review source outside the diff. It did: all 8 paths are `.md`, all mode `100644 M`, no renames out of code directories, no submodule or binary blobs, and zero code fences added or removed. Every other angle ran individually.

Twenty-two findings. The diff is documentation-only, so severity here means "how wrong a belief does this give a reader".

**The pattern worth stating up front:** this branch has now been through three correction rounds, and **each round's own fixes were the next round's defects**. Two statements written in the last pass are simply false (findings 4 and 5 below). That is not an argument against the passes — it is the argument for them.

---

### 1 · `## Pending Commands` still documents two kinds of pending · major · reported by 4 angles

`command-recognition.md:479` — *"The pending command system handles **both cases**"* — and `### Timeout Behaviour` (`:526-531`) states `FireAsIs` unconditionally: *"the pending command fires with whatever slots were filled"*. That is **false** for a disambiguation: `PendingCommandHandler.cs:347-351` returns `Confirmed` only when `pending.Reason != AwaitingDisambiguation`.

This is not a corner. `api/command-recogniser.md:45` documents `OnCommandPending`'s three reasons and then sends every reader here *"for the full treatment"*. `### Programmatic Control` still names only `HasPendingCommand` and `PendingCommand`, where `PendingAmbiguity` is the property `api/data-types.md` calls not optional to check. The PR fixed the identical undercount 90 lines away (`:571`, "two situations" → "three") and in `push-to-talk.md:65`, but not in the section that owns pending semantics.

### 2 · The session-log section describes a two-way arm that is now three-way · major

`scoring.md:447-454`: *"**Four paths** short-circuit before the parse"*, and the confirm/cancel row says *"The confirm case is the same entry with `accepted: true`"*. Item 3 made that arm three-way — `VoxrCommandRecogniser.cs:676` emits `chosen via vocabulary, now awaiting confirmation` with `accepted: false`. `grep` finds that string nowhere in the repo's documentation. It is the log line produced by the two-stage exchange this very PR documents in prose.

The same section's timeout row says *"Under `FireAsIs` the same entry carries `accepted: true`"* — false for a disambiguation, and contradicted by three other pages plus the code.

### 3 · The independence-from-eager-flush correction landed on one page of three · major

`command-recognition.md:193` now states the flag is independent of `eagerFlushOnCompleteMatch`. But `troubleshooting.md:93` and `KNOWN_LIMITATIONS.md:545-547` introduce the flag through the same *"the gate buys the decision one thing… and that is where the recogniser can ask"* narrative, with no such correction. Eager flush is **off by default**, so a default-configured reader — arriving via `index.md:34`, which this PR made the recommended first stop — is told the prerequisite is a feature they have not enabled.

### 4 · The "two hazards" callout's distinguishing test is false · minor

`command-recognition.md:131-137` says hazard 1 is *"one intent"* and that hazard 2 *"names two **intents**"*. `BuildDroppableLiteralWarning` names two intents whenever they differ (`VoxrCommandParser.cs:927-931`), and `WarnOnDroppableRequiredLiteral` has no same-intent gate. The same page says so eight lines above the callout (`:123`, *"Across commands, not just within one… It is warned about"*). The offered discriminator fails on exactly the case the doc calls out as real. The two message fragments the callout also quotes — `in front of slot "…"` vs `differ only at element N` — *are* exclusive and would carry the test alone.

### 5 · "It therefore builds no sibling set for that pattern" is false · minor

`KNOWN_LIMITATIONS.md:591-593`, written in the last correction pass. `WarningForms` returns the raw pattern past the cap (`VoxrCommandParser.cs:905-906`) and `FindSiblingSets` buckets whatever it gets. Two patterns each carrying the same ≥7 optionals and differing at one required literal normalise to the same key, form a set, and **do** warn. The true statement is narrower: the set is built from one reading only, so a relation visible in any other reading is not warned about.

### 6 · "(The eager-flush gate is *not* affected)" is an unqualified exemption its own entry retracts · nit

`KNOWN_LIMITATIONS.md:584`, also from the last pass. Two bullets later the same entry says a middle-reading relation *"stays invisible to **both**"*. The gate refuses on the all-optionals-omitted reading and commits otherwise — so it is affected, just not in the case the repro uses.

### 7 · "those forms now feed a runtime gate" contradicts the entry it points at · minor

`command-recognition.md:127` says the six-optional bound costs more now *"because those forms now feed a runtime gate"*. The entry it links says the gate is not what suffers (`KNOWN_LIMITATIONS.md:584`), and the same PR reworded that entry's own root cause from "the eager gate" to "runtime behaviour". What the bound actually costs is the warning and the question.

### 8 · "the runtime asks about **any** tie" is over-broad · minor

`KNOWN_LIMITATIONS.md:641`. The load-bearing half is correct and verified — the disambiguation branch sits after the **configured** `minScore` gate (`VoxrCommandRecogniser.cs:850` then `:916`), which is what makes the advice right for this entry's low-threshold reader. But "any tie" is contradicted 36 lines earlier by the same file's *"such a rival is never offered as a choice"*, and by three further runtime conditions between the gate and the question.

### 9 · "which of **two** indistinguishable commands" · nit

`push-to-talk.md:65`. Sibling sets are n-ary — `MaxDisambiguationRivals = 4`, so a question carries up to five choices, as this same changeset states at `command-recognition.md:224`.

### 10–15 · Structure and reach

10. **`push-to-talk.md` names the flag three times, twice as an instruction, and never links to it** — its See Also omits `command-recognition.md` entirely. Every other page that mentions the flag links it.
11. **`README.md` mirrors the docs index and was left behind** (`:26`, `:163`) — the mirror is a maintained convention, not an oversight I am inventing; commit `84a31fa` updated both in lockstep for the same reason.
12. **`KNOWN_LIMITATIONS.md:518`'s heading and Symptom still state the pre-item-3 absolute** ("fire the first-registered one"), while two bullets inside the same entry were carefully qualified with "with the flag off".
13. **The push-to-talk interaction is written twice at full depth**, in two pages authored by this PR, neither linking to the other, in different words. Same for the off-by-default rationale, which now appears four times.
14. **`KNOWN_LIMITATIONS.md:640-644` re-derives the bullet immediately above it**, reusing its exact phrase.
15. **`scoring.md:274` and `command-recognition.md:577`** carry the same rationale sentence in near-identical words, where the guide already defers to the reference for the full table.

### 16–22 · House conventions

16. `KNOWN_LIMITATIONS.md:541-542` — a 126-character line in a file hard-wrapped at 74–85 throughout.
17. `editor-testing.md:21` — the first em dashes in a file with 30 instances of ` -- ` and zero `—`.
18. `KNOWN_LIMITATIONS.md:601-604` — **cites two internal test method names**, the only test identifiers anywhere in the shipped doc set, for tests `editor-testing.md:92` says are stripped from the published package. A consumer cannot open what is named.
19. `command-recognition.md:131-137` — the callout is hard-wrapped in a file whose other blockquotes are single 260–357 character lines.
20. `command-recognition.md:189` — sentence-case heading in a file where all 13 other H2s are Title Case, including the one that also carries a colon. Inherited, but this PR is the rename.
21. `command-recognition.md:127` — em dash inside the section that uses ` -- ` exclusively.
22. `index.md:34` — a second ` -- ` reading as a second separator, plus the file's only emphasis markup.

---

### Two structural questions, not defects

**The word "sibling" names two hazards with opposite fixes**, and the callout added here teaches the reader to live with that rather than removing it. In *code* only hazard 2 is called sibling (`SiblingSet`, `AreSiblingRivals`, `disambiguateSiblingTies`); hazard 1 is `WarnOnDroppableRequiredLiteral` with no `Sibling` identifier anywhere. The collision is prose-only and therefore reachable — retiring the word for hazard 1 would make the callout unnecessary. It also does not travel: `scoring.md:478-480` now has rows for both senses, adjacent, with no callout.

**`KNOWN_LIMITATIONS.md` carries four adjacent entries in this family**, split by internal cause (expansion cap, `minScore` blind spot) where the reader arrives by symptom — which is why each needed its own separately-worded flag-interaction paragraph in this PR.

---

### Verified clean

- **GONE found nothing.** Every deletion carries its content forward; traced individually against the base. No worked example, number or remedy lost its replacement.
- **The "no code changed" claim holds** under the checks listed at the top.
- **Link integrity confirmed independently**: 228 intra-repo links, zero unresolved, file and anchor. The PR's "217" is the count excluding `Samples~/`, which resolve anyway.
- **Reader journey (a) works**: `index.md` → troubleshooting symptom entry → the fix, without the reader knowing "sibling" or "disambiguate".
- The filter order and rationale, the `rejectReason` string, the flag's independence from eager flush, `pendingTimeout` after release, the debug window's reason string, and both Console fragments were each re-traced to source and hold.

### Not verifiable from here

The Unity suites were not re-run and did not need to be — no code changed. Last green at `53b0156`: EditMode 130/130, PlayMode 522/522. On-device Quest verification is human-only and is not claimed.

---

**Findings 1–3 are what bear on merge**: each gives a reader a false belief about runtime behaviour, and 1 and 2 sit on the pages the docs themselves route readers to. Nothing here is merged or approved by this review.

jinwoo1601:
Pushed `2dc2166` — all 22 findings addressed, plus the two structural rulings.

**The three majors.** `## Pending Commands` now handles three kinds and carries the `FireAsIs`-always-cancels exception where a reader predicts timeout behaviour; its Programmatic Control paragraph names `PendingAmbiguity`. `scoring.md`'s session-log section gains the `chosen via vocabulary, now awaiting confirmation` row and its timeout row carries the disambiguation exception ("Four paths" → five). The eager-flush independence note now appears wherever the flag is introduced, not just in the guide.

**The two false statements from the previous pass** are corrected: the callout's one-intent-versus-two discriminator (`BuildDroppableLiteralWarning` names two intents whenever they differ), and "builds no sibling set for that pattern" (it builds one from a single reading; what is lost is every other reading).

**"Sibling" is retired for the bare-pattern hazard** across five pages, per the ruling. In code only the one hazard is ever called that — the other is `WarnOnDroppableRequiredLiteral`, with no `Sibling` identifier anywhere — so the collision was prose-only and is now removed rather than documented. **The callout is deleted**: it existed to explain a collision that no longer exists, and its own test was the finding above.

**Also:** the unqualified "eager gate is not affected" its own entry retracted; "any tie" narrowed to what the runtime can name; "which of two" → several, since sets are n-ary; `push-to-talk.md` now links the flag it instructs you about and owns the release interaction outright instead of restating it; `README.md`'s index line and pending-system bullet, which mirror `Documentation~` by convention; the `KNOWN_LIMITATIONS` symptom that still asserted the pre-item-3 absolute; and all seven house-style breaches — including removing two internal test method names from a shipped doc, for tests that are stripped from the published package.

**Filed not fixed: #100.** `KNOWN_LIMITATIONS.md` carries four adjacent entries on this hazard, split by internal cause where readers arrive by symptom — which is why each needed its own near-duplicate flag paragraph. Merging them restructures a document this branch only needed to correct, and would bury the corrections in a reorganisation.

**Verification:** 234 intra-repo links resolve, file and anchor (up from 228 — the retirement added cross-links). Still documentation only; no code changed, so the Unity suites remain green at `53b0156` (EditMode 130/130, PlayMode 522/522) and were not re-run.

PR #101, opened 2026-08-16
Record a non-sibling tie in the parse diagnostic (#95)

### LINKED ISSUE #95: The parse diagnostic records only sibling ties, so a non-sibling tie is indistinguishable from no tie

jinwoo1601:
Closes #95

## What

The Editor parse diagnostic now records a tie whether or not the rival was a *sibling*, and says which kind it was.

**Record** (`ParseDiagnosticEntry`, `Runtime/Commands/VoxrCommandParser.cs`)
- `TiedSiblingIntent` / `TiedSiblingPatternIndex` → `TiedRivalIntent` / `TiedRivalPatternIndex`. Renamed rather than kept, because the field now sometimes holds a non-sibling and the old name would assert the opposite.
- New `bool TiedRivalIsSibling`, plus `DescribeTiedRival()` — one phrasing (`intent (pattern N)`) shared by both Editor surfaces so they cannot drift.
- The flush loop's tie branch runs on any `CandidateOrder.Tied` instead of only when `TryFindSiblingRival` succeeds. Everything below that — the choice vocabulary, the truncation accounting, `RecordTiedRival` — stays inside the sibling arm and is unchanged.
- Exemplar precedence: first *sibling* rival wins; failing that, first rival of any kind. A sibling rival displaces a non-sibling exemplar however late it arrives, and a sibling exemplar is never displaced. This is load-bearing, not tidy — a winner's own second phrasing routinely ties it first and is not a sibling (same intent), so naming whatever tied first would shadow the cross-intent hazard behind it, which is exactly what `LastParseDiagnostics_ShadowedRival_IsTheCrossIntentOne` forbids.

**Surface**
- `VoxrMatchAttempt` gains `TiedRival` (string, null when nothing tied) and `TiedRivalIsSibling`, as two *optional* trailing constructor parameters — the synthetic attempts (no match, confirm/cancel, follow-up, timeout) have no parse round behind them and are left untouched.
- Populated in `VoxrCommandRecogniser.BuildAttempt` and `VoxrBatchTestRunner.MakeResult`.
- `VoxrDebugWindow` last-match breakdown draws a tie line: plain for a sibling tie, amber for a non-sibling one ("check for duplicate or overlapping patterns"). Never red — the command fired, and by every rule the parser has it was the right winner.
- `VoxrBatchTestWindow` prints the same distinction per result. A corpus run is where a duplicate pattern is most likely to be caught, and the pass/fail column cannot show it: the case passes whenever the coin lands on the expected intent.

**Tests / docs** — the base suite held two null assertions on this field family, and only one of them sat on a real tie, so they move in opposite directions: `LastParseDiagnostics_SameIntentTie_RecordsNoRival` is the one genuine non-sibling tie and its expectation **flips** (renamed `…_RecordsANonSiblingRival`), while `LastParseDiagnostics_NoTie_RecordsNoRival` **keeps** asserting null — nothing ties there — and is touched only for the rename plus new `IsSibling` / `DescribeTiedRival` pins. (This corrects the wording this PR originally carried, "the two tests that asserted null for a non-sibling tie", which is inherited from issue #95's own phrasing and miscounts: one such test, not two.) `IsSibling` assertions added to the shadowed-rival and truncated-tie tests so the displacement rule and the "unnameable ≠ not a sibling" distinction are pinned; new `LastParseDiagnostics_DuplicatePatterns_RecordANonSiblingRival`. `CHANGELOG.md` under `[Unreleased]` → Added, and the `editor-testing.md` bullet that enumerates what the breakdown shows.

## Why

Design §5.3 promised that restricting the *action* to sibling rivals keeps non-sibling ties out of the runtime paths "while leaving them visible to the editor diagnostic". Only the first half was built. Three shapes therefore left the record null, indistinguishable from "nothing tied": a tie between patterns that are not siblings at all, a tie between two members carrying the same discriminating value, and a tie inside a wholly same-intent set.

The headline case is a grammar with two intents on the same pattern. The construction-time sibling warning is silent there — siblings differ at exactly one position and duplicates differ at none — so the author sees a clean `1.00` PASS on a grammar containing a command that can never fire.

The runtime record stays sibling-only, and that is not a compromise: DR-4 makes the discriminating values the choice vocabulary, so a tie with no such values is a tie there is no question to ask. What issue #95 asked for was the diagnostic distinguishing "a rival tied" from "a *sibling* rival tied", which is what the new bool does.

## Validation

- [x] Unity EditMode — **131/131 passed**, 0 failed, 0 skipped (`Jinwoo1601.VoXR.Tests.Editor.dll`)
- [x] Unity PlayMode — **522/522 passed**, 0 failed, 0 skipped (`Jinwoo1601.VoXR.Tests.Runtime.dll`)
- [x] Both run against the checked-out working tree in the `VoXR TestGround` host project (Unity 6000.4.7f1), per the project's verification bindings. Green confirmed genuine rather than a stale assembly: the result XML contains the newly added and renamed test names, which exist only in this branch's test file, and neither log carries a `CS` error — so no reference to the old field names was missed.
- [x] No runtime behaviour change: `RecordTiedRival` and the `TiedSiblingRecord` buffer are untouched, and two tests assert `RivalCount == 0` on the non-sibling shapes. The pre-existing `TiedSiblingRecord_*` PlayMode family still passes, which is independent evidence the rename stayed inside the Editor diagnostic.
- NativeBridge build and desktop harness not run — `NativeBridge~/` is untouched, and the binding gates them on that directory changing.
- No on-device Quest check needed: Editor-diagnostic-only, no mic capture and no MonoBehaviour lifecycle path.

## Deliberately out of scope

`VoxrDebugSessionLog`'s `AttemptDto` does not carry the new fields. The auto-exported JSON has a documented `schemaVersion` and a `readme` prose contract, so adding fields is a shape change requiring a version bump — a different question from "render the difference to the author", which is what the issue asked for. Worth a follow-up issue if the session log should stay in step with the windows.


jinwoo1601 (commented):
## Independent review — PR #101

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`, plus **`XFILE` added** because the diff renames public fields on a struct and appends two optional parameters to a constructor with call sites in two assemblies. No angles dropped.
**Head reviewed:** `8eace59b5bd011a41c268cbab315672222b7a656` — base merge-base `9db32eb`.
**Result:** one confirmed finding, at nit severity, against the PR body rather than the code. `LINE` and `GONE` each returned zero findings; the correctness of the restructured branch and the ticked validation boxes both hold up under audit.

---

### 1. The PR body miscounts which base tests it changed — nit, CONFIRMED

The body says *"the two tests that asserted null for a non-sibling tie updated"*. Only **one** base test did. `9db32eb:Tests~/Editor/VoxrCommandParserDiagnosticTests.cs:261` (`LastParseDiagnostics_SameIntentTie_RecordsNoRival`) is the genuine non-sibling tie, and its expectation correctly flips. The other null assertion, `9db32eb:…:231-232` in `LastParseDiagnostics_NoTie_RecordsNoRival`, is a shape where **nothing ties at all** — its null is deliberately preserved, not flipped.

That it truly does not tie is proved by the head, not by the test's name: the widened branch at `Runtime/Commands/VoxrCommandParser.cs:2226` now records an exemplar for *any* rival, yet that test still asserts null and passes.

The wording is inherited verbatim from issue #95's own "updates to the two tests that currently assert null for a non-sibling tie", so the PR reproduces the issue's miscount rather than inventing it — but the substance shipped correctly either way. Accurate wording: one expectation flipped, one null deliberately preserved.

---

### Considered and killed

- **`DescribeTiedRival()` could silently drop the tie line on a null intent** — refuted: `VoxrCommandDefinition.cs:29` is `Intent = intent ?? throw`, so a null intent cannot reach the record, and `TiedRivalIntent == null` is an exact proxy for "nothing tied".
- **The widened branch could record a phantom rival against a nonexistent winner** — refuted: `CompareCandidate` returns `Better`, never `Tied`, while `bestScore <= 0f` (`VoxrCommandParser.cs:2884`), so `TryFindSiblingRival(-1, -1, …)` is unreachable.
- **Optional trailing parameters could silently rebind arguments at an existing call site** — refuted: all eight `new VoxrMatchAttempt(...)` sites pass ≤ 9 positional arguments and the two new parameters are appended after `isAccepted`, not inserted.
- **State could leak across extraction rounds or adopted incumbents** — refuted: `diagRivalIsSibling` is declared, written and cleared only alongside the two existing exemplar locals, including in the clear-on-adopt block.

### What was verified rather than taken on trust

The exemplar-precedence ternary was re-derived over all four states: a sibling still wins first-and-never-displaced exactly as the base's `if (diagRivalCommandIdx < 0)` did, so the final exemplar is identical on every grammar the base could see. The sibling-only choice path (`sawUnnameableRival`, `sameQuestion`, `tiedTruncated`, `RecordTiedRival`) is byte-identical to base and equally reachable — the runtime record does not widen. The rename is total: zero stale references outside `Planning~/features/*`, which are immutable process docs recording the pre-#95 state. Assembly visibility lines up (`InternalsVisibleTo` + `#if UNITY_EDITOR` on every new consumer). The ticked test counts match the retained artifacts — `verify95-editmode.xml` carries `passed="131" failed="0"`, `verify95-playmode.xml` `passed="522" failed="0"`, both containing the branch-only test names, with zero `error CS` in either log and every source file's mtime preceding both runs.

### Notes — unverified, not blocking

- `VoxrCommandParser.TiedRival(int, int)` (the sibling-**only** runtime choice accessor) and the new `VoxrMatchAttempt.TiedRival` (explicitly *any* kind) now share one identifier with opposite guarantees in one namespace. Nothing binds wrongly; by the PR's own rename rationale the runtime accessor is the one that should keep `Sibling` in its name.
- `LastParseDiagnostics_SameIntentAcrossTwoCommands_IsNotTheRecordedRival` and its comment describe a mechanism that changed: that rival now *is* recorded as an interim exemplar and then displaced. The assertion still holds; the name and comment no longer explain why.
- No test covers the second half of the stated precedence rule — that a sibling exemplar is never displaced by a later sibling. `ShadowedRival` exercises only non-sibling-then-sibling.
- The tie-line suffix wording is duplicated verbatim in `Editor/VoxrDebugWindow.cs:441,451-452` and `Editor/VoxrBatchTestWindow.cs:283-287`; only the `intent (pattern N)` half is centralised in `DescribeTiedRival()`.
- The amber "check for duplicate or overlapping patterns" wording also fires on a single command's own second phrasing, which is a dead-phrasing hazard rather than a duplicate across intents.
- `Runtime/Commands/VoxrCommandParser.cs:2383-2393`: the initializer's new members sit four spaces deeper than `PatternString` above them. This is the repo's formatter hook's doing — a manual de-indent was reverted by it — and is left as reviewed rather than pushed, so this review's line citations stay true to the head SHA.
- `KNOWN_LIMITATIONS.md:518,584` still name the construction-time console warning as the only detection route for sibling ties; nothing there is now untrue, but the new tie line would fit.

### Not verifiable from here

- **On-device (Quest) behaviour** — no headset reachable from WSL. The PR claims none is needed, and that claim audits clean: the diff touches no mic capture, no `MonoBehaviour` lifecycle hook, and no main-thread-only path.
- **The Unity suites were not re-run** for this review. The ticked counts were audited against the retained NUnit XML and logs rather than reproduced.
- The `NativeBridge~/` build and desktop harness were correctly not run — that directory is untouched, and the project binding gates them on it changing.

jinwoo1601:
Addressed review finding 1 (the PR body's miscount). The **Tests / docs** bullet now states what the base suite actually pinned: one expectation flipped (`SameIntentTie`, the only genuine non-sibling tie) and one null deliberately preserved (`NoTie`, where nothing ties), rather than "the two tests that asserted null for a non-sibling tie".

Body edit only — no code change, so the review above still stands against head `8eace59`. The seven Notes remain unruled.

jinwoo1601:
Filed the naming Note from the review above as #102 — `VoxrCommandParser.TiedRival(...)` is sibling-only while `VoxrMatchAttempt.TiedRival` deliberately is not. Low severity, internal-only, no public API change.

The remaining six Notes stay unruled.

PR #103, opened 2026-08-17
Name the sibling-only rival family for what it restricts (#102)

### LINKED ISSUE #102: `TiedRival` names both the sibling-only choice accessor and the any-kind diagnostic field
Raised by the independent review of PR #101 (issue #95) and left unruled at merge, so it is on the board rather than lost in a review comment.

`VoxrCommandParser.TiedRival(...)` is strictly **sibling-only**. `VoxrMatchAttempt.TiedRival` is explicitly **not**. One identifier, two opposite guarantees, both `internal` to `VoXR.Commands`.

## What is true today

The runtime side indexes the choice buffer that DR-4 restricts to sibling rivals — a non-sibling tie has no discriminating values, so it can never be offered as a choice and never enters this buffer:

```csharp
// Runtime/Commands/VoxrCommandParser.cs:2527
internal TiedSiblingRival TiedRival(int resultIdx, int n) =>
    _tiedRivalBuf[resultIdx * MaxDisambiguationRivals + n];
```

The diagnostic side is the half issue #95 deliberately widened. It may name a rival that is **not** a sibling — which is the entire reason `TiedRivalIsSibling` had to be added beside it:

```csharp
// Runtime/Commands/VoxrMatchDiagnostics.cs:58
public readonly string TiedRival;
```

So a maintainer reading `TiedRival` in `Editor/VoxrDebugWindow.cs:436` and then in `VoxrCommandRecogniser.cs:1115` (`parser.TiedRival(i, n).Value`, which feeds the vocabulary the speaker answers with) gets two different contracts from one name, with nothing at either site saying which is in force.

The mismatch is not confined to the accessor. The whole sibling-only family drops the word: `RecordTiedRival` (`:2414`), `_tiedRivalBuf` (`:375`, whose element type is `TiedSiblingRival`), and `RivalIntent` (`:2549`) are all sibling-only. Their neighbours say so and they do not — `TiedSiblingBuffer` (`:2523`), `TiedSiblingRecord`, `TiedSiblingRival` (`:64`), `_tiedSiblingBuf`.

## Why it happened

Before #95 there was no ambiguity to resolve: every tie the parser recorded anywhere was a sibling tie, so `TiedRival` and "tied sibling rival" were the same thing and the shorter name was free. #95 widened the *diagnostic* record to cover non-sibling ties and renamed `TiedSiblingIntent`/`TiedSiblingPatternIndex` to `TiedRivalIntent`/`TiedRivalPatternIndex` for exactly the stated reason — the field now sometimes holds a non-sibling and the old name would assert the opposite.

That rename is right, and it is what left the runtime accessor as the odd one out: `TiedRival` is now the name for "any kind" everywhere except the one place where it means the opposite. The review that caught it put the standard plainly — the PR applied its own rename rule to one side of the boundary and not the other.

## What closing this needs

Rename the sibling-only family so the word `Sibling` appears where the sibling restriction applies, leaving `TiedRival` to mean "a rival of any kind" consistently. Everything involved is `internal`, so **no public API changes** and no `CHANGELOG` entry — `Runtime/AssemblyInfo.cs` grants `InternalsVisibleTo` to the Editor and both test assemblies, which is the full blast radius.

Call sites are few: `VoxrCommandRecogniser.cs:1115`, the parser's own `:2549` and `:2569`, and the `TiedSiblingRecord_*` family in `Tests~/Runtime/VoxrCommandParserTests.cs` (roughly lines 4700–5080).

Worth settling in the same pass: whether `_tiedRivalBuf` should match its own element type (`TiedSiblingRival`), since the private field is where the mixed vocabulary starts.

## Severity

Low, and readability-only. Nothing binds wrongly today — the two are members of different types, so no call site can resolve to the wrong one, and PR #101's review verified every `VoxrMatchAttempt` call site binds correctly. This is about a maintainer inferring the wrong contract from a name, in code where the sibling/non-sibling distinction is load-bearing enough that #95 existed to make it visible.

jinwoo1601:
Closes #102

## What

`TiedRival` named two opposite contracts. `VoxrCommandParser.TiedRival(...)` indexed a buffer DR-4 restricts to **sibling** rivals; `VoxrMatchAttempt.TiedRival` is the diagnostic field issue #95 deliberately widened to name a rival of **any** kind. Both `internal` to `VoXR.Commands`, nothing at either call site saying which was in force.

This applies the rename rule #95 established — say `Sibling` where the sibling restriction applies — to the parser's whole choice-buffer family, leaving `TiedRival` to mean "a rival of any kind" consistently:

| before | after |
|---|---|
| `TiedRival(i, n)` | `TiedSiblingRivalAt(i, n)` |
| `RecordTiedRival(…)` | `RecordTiedSiblingRival(…)` |
| `_tiedRivalBuf` | `_tiedSiblingRivalBuf` |
| `RivalIntent(i, n)` | `SiblingRivalIntent(i, n)` |
| `CopyRivalSlots(i, n)` | `CopySiblingRivalSlots(i, n)` |
| `BuildRivalCommand(…)` | `BuildSiblingRivalCommand(…)` |
| `_rivalSlotBuf` | `_siblingRivalSlotBuf` |

`VoxrMatchAttempt.TiedRival` / `TiedRivalIsSibling` and their two Editor readers are untouched — that is the point.

## Why

Two judgement calls the reviewer should rule on rather than assume:

**The `At` suffix.** Inserting `Sibling` mechanically gives `TiedSiblingRival(i, n)` — a method sharing its name with the struct it returns. C# permits it, but it is a worse readability trap than the one being fixed, so the accessor takes `At`. The doc comment on it says why, and also states the sibling-only contract and its contrast with `VoxrMatchAttempt.TiedRival`.

**Three names beyond the issue's list.** The issue enumerates four members. `CopyRivalSlots`, `BuildRivalCommand`, and `_rivalSlotBuf` are sibling-only and drop the word by exactly the same criterion, and index the same flat buffer. Renaming four of the seven would reproduce the fault the issue itself names — applying a rename rule to one side of a boundary and not the other. Trim this back if you disagree; it is the only part of the diff not literally enumerated in #102.

Rename only. No behaviour change. Four are `internal` and three (`_tiedSiblingRivalBuf`, `_siblingRivalSlotBuf`, `RecordTiedSiblingRival`) are `private`, all inside an `internal class VoxrCommandParser` — **none is `public`**, so no public API change and no `CHANGELOG` entry. `Runtime/AssemblyInfo.cs` grants `InternalsVisibleTo` to `Jinwoo1601.VoXR.Editor` and both test assemblies, which is the full blast radius for the four `internal` ones. No `Documentation~/`, `Samples~/`, or root-markdown references existed.

## Validation

- [x] **Unity EditMode: 131/131 pass** (`failed="0"`, `inconclusive="0"`, `skipped="0"`) via host project `VoXR TestGround`, Unity 6000.4.7f1.
- [x] **Unity PlayMode: 522/522 pass** (`failed="0"`, `inconclusive="0"`, `skipped="0"`) — where the parser/command tests live; an EditMode-only run would have missed them.
- [x] **Zero compile errors across all four assemblies** (Runtime, Editor, Tests.Runtime, Tests.Editor). This is the load-bearing check for a rename of `internal` members: a missed call site is a compile error, and there were none.
- [x] Post-rename sweep for stale `_tiedRivalBuf` / `_rivalSlotBuf` / `RecordTiedRival` / `RivalIntent` / `CopyRivalSlots` / `BuildRivalCommand` / parser-side `TiedRival(` references. **Corrected after review** — as originally worded ("anywhere in the repo") this box was false. `grep` in this shell is a wrapper passing `--ignore-files`, so it honours `.gitignore` and silently skipped `Planning~/`, which contains compiled C# that stages and compiles these very Runtime sources. One stale call site was there (`ab-rig/PlayerCheck.cs:77`) and broke that rig's build. It is now fixed, but `Planning~/` is gitignored so **that fix is not in this diff**. Tracked content is confirmed clean via `command grep`.
- [x] **Grammar A/B rig green on the renamed parser** — `./stage.sh WORKTREE` then `PlayerCheck` **7/7, exit 0** and `DocCheck` **50/50 doc claims, exit 0**. `PlayerCheck` is the only build in this project that compiles *without* `UNITY_EDITOR` defined, so this is the one check that reaches the shipped player configuration — which all 653 Unity tests structurally cannot.
- [x] `Tests~` edited via `sed` rather than the editor tooling, to keep the CSharpier hook from reformatting the whole (not-currently-CSharpier-clean) file; only the four lines pushed past 100 chars were rewrapped by hand. Diff is 10 renamed call sites plus that rewrapping.

No on-device (Quest) check applies: nothing here changes runtime behaviour, and the rename is proven by compilation plus 653 passing tests.

### DIFF
--- a/Runtime/Commands/VoxrCommandParser.cs
+++ b/Runtime/Commands/VoxrCommandParser.cs
@@ -372,8 +372,8 @@ namespace VoXR.Commands
         // Parallel to _resultBuf: index i describes the result at index i. Rival detail lives in
         // flat preallocated arrays rather than inside the struct so nothing allocates per round.
         TiedSiblingRecord[] _tiedSiblingBuf;
-        TiedSiblingRival[] _tiedRivalBuf; // _resultBuf.Length * MaxDisambiguationRivals
-        VoxrSlotMatch[] _rivalSlotBuf; // ...that, times _maxSlotsPerPattern
+        TiedSiblingRival[] _tiedSiblingRivalBuf; // _resultBuf.Length * MaxDisambiguationRivals
+        VoxrSlotMatch[] _siblingRivalSlotBuf; // ...that, times _maxSlotsPerPattern
 
         // The cancel words the recogniser will actually match against, resolved once here so
         // the construction-time collision report tests the same array TryHandleConfirmCancel
@@ -606,7 +606,7 @@ namespace VoXR.Commands
             // allocate there.
             //
             // Left NULL when nothing will record, which is what DR-7's "costs a flag-off player
-            // nothing" asks for at the rebuild as well as at the parse. _rivalSlotBuf is
+            // nothing" asks for at the rebuild as well as at the parse. _siblingRivalSlotBuf is
             // commands x 4 x maxSlots VoxrSlotMatch — tens of KB on a large grammar, re-made on
             // every NotifySlotChanged() and never read. Every reader is downstream of
             // _recordSiblingTies (the recording branch tests it first, and the recogniser only
@@ -615,8 +615,10 @@ namespace VoXR.Commands
             if (_recordSiblingTies)
             {
                 _tiedSiblingBuf = new TiedSiblingRecord[_resultBuf.Length];
-                _tiedRivalBuf = new TiedSiblingRival[_resultBuf.Length * MaxDisambiguationRivals];
-                _rivalSlotBuf = new VoxrSlotMatch[
+                _tiedSiblingRivalBuf = new TiedSiblingRival[
+                    _resultBuf.Length * MaxDisambiguationRivals
+                ];
+                _siblingRivalSlotBuf = new VoxrSlotMatch[
                     _resultBuf.Length * MaxDisambiguationRivals * _maxSlotsPerPattern
                 ];
             }
@@ -2291,7 +2293,7 @@ namespace VoXR.Commands
                                         tiedTruncated = true;
                                     }
                                     else
-                                        RecordTiedRival(
+                                        RecordTiedSiblingRival(
                                             ci,
                                             pi,
                                             rivalSetId,
@@ -2411,7 +2413,7 @@ namespace VoXR.Commands
         //
         // matchResult is the RIVAL's match — its slots are in _matchSlotBuf right now, which is
         // why the capture has to happen here and not later.
-        void RecordTiedRival(
+        void RecordTiedSiblingRival(
             int ci,
             int pi,
             int rivalSetId,
@@ -2448,7 +2450,7 @@ namespace VoXR.Commands
             int firstRival = _resultCount * MaxDisambiguationRivals;
             for (int r = 0; r < tiedRivalCount; r++)
             {
-                var kept = _tiedRivalBuf[firstRival + r];
+                var kept = _tiedSiblingRivalBuf[firstRival + r];
                 if (
                     IsAnswerableRival(
                         kept.Value,
@@ -2488,7 +2490,7 @@ namespace VoXR.Commands
             }
 
             int slot = firstRival + tiedRivalCount;
-            _tiedRivalBuf[slot] = new TiedSiblingRival
+            _tiedSiblingRivalBuf[slot] = new TiedSiblingRival
             {
                 CommandIndex = ci,
                 PatternIndex = pi,
@@ -2507,7 +2509,7 @@ namespace VoXR.Commands
                 Array.Copy(
                     _matchSlotBuf,
                     0,
-                    _rivalSlotBuf,
+                    _siblingRivalSlotBuf,
                     slot * _maxSlotsPerPattern,
                     matchResult.SlotCount
                 );
@@ -2522,34 +2524,44 @@ namespace VoXR.Commands
         // recogniser already walks the result buffer by index in its Step 7 loop.
         internal TiedSiblingRecord[] TiedSiblingBuffer => _tiedSiblingBuf;
 
-        // Rival n of result i. Flat indexing rather than a jagged array so nothing allocates per
-        // round; the caller has already checked n against TiedSiblingBuffer[i].RivalCount.
-        internal TiedSiblingRival TiedRival(int resultIdx, int n) =>
-            _tiedRivalBuf[resultIdx * MaxDisambiguationRivals + n];
+        // Sibling rival n of result i. Flat indexing rather than a jagged array so nothing
+        // allocates per round; the caller has already checked n against
+        // TiedSiblingBuffer[i].RivalCount.
+        //
+        // Every member of this family says Sibling, and that is not decoration (issue #102):
+        // this buffer holds ONLY sibling rivals, because DR-4 makes the discriminating values
+        // the choice vocabulary and a non-sibling tie has none. VoxrMatchAttempt.TiedRival is
+        // the opposite — the Editor diagnostic that issue #95 widened to name a rival of ANY
+        // kind. One name for both contracts is what this family was renamed away from.
+        //
+        // The At suffix is only there because C# would otherwise have this method and the
+        // struct it returns share a name.
+        internal TiedSiblingRival TiedSiblingRivalAt(int resultIdx, int n) =>
+            _tiedSiblingRivalBuf[resultIdx * MaxDisambiguationRivals + n];
 
         // That rival's own slot matches, copied out fresh because they cross into a public
         // VoxrCommand the subscriber can retain — the PendingCommandHandler precedent that
         // anything reaching a public event is allocated, never pool-borrowed. Once per
         // ambiguity, never per candidate.
-        internal VoxrSlotMatch[] CopyRivalSlots(int resultIdx, int n)
+        internal VoxrSlotMatch[] CopySiblingRivalSlots(int resultIdx, int n)
         {
             int slot = resultIdx * MaxDisambiguationRivals + n;
-            int count = _tiedRivalBuf[slot].SlotCount;
+            int count = _tiedSiblingRivalBuf[slot].SlotCount;
             if (count <= 0)
                 return Array.Empty<VoxrSlotMatch>();
 
             var slots = new VoxrSlotMatch[count];
-            Array.Copy(_rivalSlotBuf, slot * _maxSlotsPerPattern, slots, 0, count);
+            Array.Copy(_siblingRivalSlotBuf, slot * _maxSlotsPerPattern, slots, 0, count);
             return slots;
         }
 
-        // Rival n's intent, so the recogniser can resolve its definition before offering it as
-        // a choice — and drop it from the list if that lookup fails.
-        internal string RivalIntent(int resultIdx, int n) =>
-            _commands[TiedRival(resultIdx, n).CommandIndex].Intent;
+        // Sibling rival n's intent, so the recogniser can resolve its definition before offering
+        // it as a choice — and drop it from the list if that lookup fails.
+        internal string SiblingRivalIntent(int resultIdx, int n) =>
+            _commands[TiedSiblingRivalAt(resultIdx, n).CommandIndex].Intent;
 
-        // The command that fires if the speaker picks rival n — built here rather than in the
-        // recogniser because _slotNames and the confidence span both live on this side.
+        // The command that fires if the speaker picks sibling rival n — built here rather than
+        // in the recogniser because _slotNames and the confidence span both live on this side.
         //
         // Confidence is computed over the RIVAL's own span. ComputeConfidence takes
         // (tokens, startIdx, endIdx): startIdx comes from the record, because CompareCandidate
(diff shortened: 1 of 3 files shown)


jinwoo1601 (commented):
## Independent review — PR #103

**Depth profile:** issue lane → **slim** (`CLAIM`, `LINE`, `GONE`), **plus `XFILE` added**. `XFILE` was added because the renamed members are consumed across assembly boundaries via `InternalsVisibleTo`, making "did every consumer follow the rename" a cross-file contract question. That addition is what found the one substantive item below; the three slim angles would have missed it. No angles were dropped.

**Head reviewed:** `64b6359a32a70b663a87fc8c412d401bbaad6a3a` · **merge-base:** `80dae14`

**The rename itself is clean.** `LINE` and `GONE` each independently reconstructed head from base by applying only the seven claimed renames, then compared normalized token streams. All three files are token-identical, with one exception: a trailing comma added inside an array initializer at `Tests~/Runtime/VoxrCommandParserTests.cs:4800`, which is a C# no-op. `[Test]` attributes 197/197 and assertions 495/495, base vs head. That forecloses any altered literal, operator, bound, argument order, or dropped comment across the whole diff — invariant 1 (pure rename) holds mechanically, not by inspection. All three findings below are about the PR's *description* or *tooling*, not the shipped code.

---

### 1. A stale call site survives outside the diff, breaking the project's only player-side compile check — `minor`, CONFIRMED

`Planning~/features/coverage-in-selection/ab-rig/PlayerCheck.cs:77` still called the removed accessor:

```csharp
&& on.TiedRival(0, 0).Value == "level",
```

This is compiled C#, not a doc reference: `stage.sh:16` stages `Runtime/Commands/VoxrCommandParser.cs` — in `WORKTREE` mode, literally the file under review — and `Rig.csproj:22` compiles it alongside `PlayerCheck.cs` into one assembly. The verifier reproduced the break end-to-end: `PlayerCheck.cs(77,27): error CS1061: 'VoxrCommandParser' does not contain a definition for 'TiedRival'`, `Build FAILED`. The same staging at the merge-base succeeds, so this PR is the cause. One assembly means the failure takes `DocCheck` (the #83 documentation-number pin) and `Program` (the #65 A/B replay) with it, whichever `StartupObject` is selected.

This matters more than "gitignored scratch tooling": `PlayerCheck.cs:3` calls itself "the ONLY build in the project that compiles without `UNITY_EDITOR`", written after this design shipped player-side breakage twice that the Unity suite structurally could not see.

**This falsifies validation box 4** ("no stale … references remain **anywhere in the repo**"). The mechanism is worth recording: `grep` in this shell is a wrapper passing `--ignore-files`, so it honours `.gitignore` and silently skips `Planning~/`. Any post-rename sweep run the obvious way had the same blind spot.

**Disclosure — already fixed, and the fix is NOT in this PR.** `Planning~/` is gitignored, so the repair cannot be committed here. The line now reads `on.TiedSiblingRivalAt(0, 0).Value` in the local checkout; the rig builds clean and `PlayerCheck` passes **7/7, exit 0**. Reverse it freely if you disagree with my fixing it unruled — it touches no tracked file. Bonus: that green run is positive evidence for the pure-rename invariant on the *player* side, which all 653 Unity tests structurally cannot reach.

### 2. "All seven are `internal`" is false — `nit`, CONFIRMED

Three of the seven carry no accessibility modifier and are therefore `private`: `_tiedSiblingRivalBuf` (`VoxrCommandParser.cs:375`), `_siblingRivalSlotBuf` (`:376`), `RecordTiedSiblingRival` (`:2416`). Only the four methods at `:2539`, `:2546`, `:2560`, `:2574` are `internal`. The backticked keyword next to the `InternalsVisibleTo` sentence pins this to the C# modifier, not a colloquial reading.

The conclusion is unaffected — `private` is narrower than `internal`, the containing type is `internal class VoxrCommandParser` (`:137`, sole declaration, not `partial`), and none is `public` — so **the no-public-API / no-`CHANGELOG` decision stands**. Accurate wording: "four `internal` and three `private`, all inside an `internal` class."

### 3. `VoxrMatchDiagnostics.TiedRival` names a member that does not exist — `nit`, CONFIRMED

`VoxrMatchDiagnostics` (`Runtime/Commands/VoxrMatchDiagnostics.cs:12-30`) declares only `InputText`, `Words`, `Attempts`, `Frame`. `TiedRival` (`:58`) and `TiedRivalIsSibling` (`:66`) are on `VoxrMatchAttempt` (`:32`), a sibling top-level type — not nested, so no qualifier reaches them that way. A repo-wide search finds zero precedent for file-scoped qualification of this type.

The PR body writes `VoxrMatchAttempt.TiedRival` correctly twice elsewhere, and the shipped doc comment at `VoxrCommandParser.cs:2533` gets it right — so the body contradicts its own code on the single identifier the PR is about. The substance is sound: `git diff` on that file is empty and there are exactly two Editor readers, as claimed.

---

### Refuted, and why — so you can see what was considered and killed

- **"Diff is 10 renamed call sites" undercounts the 12 invocations.** REFUTED. The 12/10/11 arithmetic is right, but `git diff --stat` shows exactly **10 deletions**, and the sentence ("Diff is 10 … plus that rewrapping") is line-composition grammar. 10 is derivable and exact. The load-bearing half of that box is precise to the line: exactly four base lines cross 100 chars post-rename (4752→106, 4793→103, 4820→101, 5017→103) and exactly those four were rewrapped; 5014 lands at 91 and stays.

- `Planning~/features/disambiguation-pending/architecture.md:176,366,584,624` still spell `_rivalSlotBuf`, `BuildRivalCommand`, `RivalIntent`, `_tiedRivalBuf`. Locked process docs are historical records, so leaving them is probably right — but someone owns that call.
- `Tests~/Runtime`'s asmdef sets `"excludePlatforms": ["Android"]`, so the renamed test call sites never compile for the shipping target.
- The rewrap at `VoxrCommandRecogniser.cs:1106-1108` splits an `if` condition across three lines purely for width.

### Not verifiable from here

- **The Unity suite was not re-run by this review.** Re-running requires renaming `Tests~`→`Tests` in the tree under review, which would mutate it. The 131 EditMode / 522 PlayMode claims were audited for coherence only — and they reconcile exactly (EditMode: 118 `[Test]` + 13 `[TestCase]` in `Tests~/Editor` = 131; PlayMode: 476 + 10 + 21 + the 16-fixture `[ValueSource]` expansion at `VoxrWavReplayTests.cs:137` = 522).
- **On-device (Quest) behavior.** The PR argues no device check applies because nothing changes runtime behavior; the token-identity result above supports that reasoning, but no device was exercised.

**Verdict: the diff is sound and does what #102 asked.** Nothing in the merged artifact is defective. The three confirmed items are two wording corrections to the PR body and one local-tooling repair already applied outside the diff. Not approving or merging — yours to rule on.

<sub>Slim profile + `XFILE`. 4 finder angles, 4 findings surfaced, 3 confirmed / 1 refuted by independent verifiers.</sub>

jinwoo1601:
## All review findings addressed

No code change — the diff at `64b6359` is untouched, because all three confirmed findings were about the PR's description or about tooling outside it.

**Finding 1 — stale call site breaking the player-side compile check.** Fixed: `ab-rig/PlayerCheck.cs:77` now reads `on.TiedSiblingRivalAt(0, 0).Value`. Both rig entry points verified green against the renamed parser, each in a fresh staging (`StartupObject` does not survive an incremental rebuild):

- `PlayerCheck` — **7/7, exit 0**
- `DocCheck` — **50/50 doc claims verified, exit 0**

`Planning~/` is gitignored, so this repair is genuinely absent from the diff. It is worth having anyway: `PlayerCheck` compiles *without* `UNITY_EDITOR`, so its green run is the only evidence available that the rename is behaviour-preserving in the **shipped player configuration** — the configuration all 653 Unity tests structurally cannot reach, since Unity always defines that symbol.

**Finding 2 — "All seven are `internal`."** Corrected in the body to "four `internal` and three `private`, all inside an `internal class VoxrCommandParser` — none is `public`." The no-public-API / no-`CHANGELOG` conclusion is unchanged.

**Finding 3 — `VoxrMatchDiagnostics.TiedRival`.** Corrected in the body to `VoxrMatchAttempt.TiedRival`, matching the shipped doc comment at `VoxrCommandParser.cs:2533` and issue #102's own wording.

**Validation box 4** (falsified by finding 1) is rewritten rather than quietly re-ticked: it now states that the original "anywhere in the repo" wording was false, names the cause, and records that the fix lands outside the diff. A new box records the rig runs.

### Notes — dispositions

- **`Planning~/features/disambiguation-pending/architecture.md:176,366,584,624` keep the old names — deliberately, not by oversight.** All four are historical narrative: an as-built rationale, a bug post-mortem, a review-findings list, and a code sketch, each describing what was true when that feature was built. Under this project's workflow, process docs are immutable once locked; editing them to match a later rename would falsify the record. Nothing user-facing is stale — `Documentation~/`, `Samples~/`, and root markdown are confirmed clean of every name in this family. Say so if you would rather they were amended.
- **`Tests~/Runtime`'s `"excludePlatforms"`** is pre-existing and untouched by this PR. Worth its own issue if it should change; not something to settle inside a rename.
- **The `if`-condition rewrap at `VoxrCommandRecogniser.cs:1106-1108`** is CSharpier's own output for a line that crossed 100 chars. Left as the formatter produced it.

Still not approving or merging — yours to rule on.

PR #104, opened 2026-08-17
Pin the sibling-set edge branches and lookup invariants (#93)

### LINKED ISSUE #93: Sibling-set edge branches and lookup invariants have no executable coverage

jinwoo1601:
Closes #93

> **Revised after review.** All four findings from the review below were confirmed and fixed in `dff809b` (force-push; the previous head `5c4005c` carried an overclaim in its commit message, so it was amended rather than followed up). This body is corrected too — see [Review fixes](#review-fixes) at the end for exactly what changed.

## What

Five tests, no production change. **Four** were verified by deleting the code they guard and watching them go red, with the failure message quoted below so you can check each failed for the right reason. The fifth guards something no test can witness; it is marked as such in the table rather than counted as a witness it is not.

| # | Gap | Test | Suite | Mutation-witnessed |
|---|---|---|---|---|
| 1 | `BuildSiblingWarning` value-list dedup | `SiblingWarning_TwoPatternsSharingAValue_NameThatValueOnce` | PlayMode | yes |
| 1 | cancel-collision `alreadyReported` guard | `SiblingWarning_TwoPatternsCarryingOneCancelWord_ReportItOnce` | PlayMode | yes |
| 2 | per-round tie reset | `LastParseDiagnostics_TieInRoundOne_DoesNotLeakIntoRoundTwo` | EditMode | **no — not observable, see Finding 2** |
| 3 | invariant 8, allocation-free selection | `EagerSelection_OverASiblingTie_AllocatesNothingPerCall` | PlayMode | yes |
| 3 | invariant 9, lookup built once (F13) | `SiblingLookup_IsBuiltOncePerParser_NotOncePerParse` | PlayMode | yes |

Fixture notes, since the issue asked for these two to match the pattern-text dedup test:

- The **value dedup** fixture gives its two duplicate-valued members *different* pattern text — `set_a` reaches the frame by omitting its optional — so the pattern-text dedup does nothing and the value list is the only thing under test.
- The **cancel dedup** fixture makes *both* `"negative"` members answerable, which is exactly what `SiblingWarning_CancelCollisionReachableOnlyThroughOnePair_StillReports` is not: there the dedup must let the later member through, here it must suppress it.

### One correction to the issue's premise

Issue #93 says all these branches are "nothing fails if this is deleted". That is right for three of the four, and was right for the fourth **when the issue was filed**. It is no longer: `SiblingWarning_SymmetricSetWithABarePattern_NumbersEachPatternToo` asserts a full message on a set carrying `"mode"` twice, so the **value-list dedup was already going red incidentally**. That test landed in `d6fa33c`, roughly five hours after #93 was opened.

The new test is still worth having — the incidental guard rides on a fixture built to pin per-pattern element numbering and would stop covering the dedup the moment that fixture changed for its own reasons — but it is a *deliberate, isolated* guard replacing an accidental one, not first coverage. The test comment says so.

## Why

The branches were untested and the invariants were argued in prose. Mutation-testing each one is what turned that from a claim into evidence — and it caught two of my own tests being as blind as the code they were meant to guard. Both are below.

## Two findings beyond the issue

**1. `GC.GetAllocatedBytesForCurrentThread()` is inert on this runtime.** A probe measured **0 B moved after a deliberate 1 MB allocation**. I wrote the allocation test against that counter first and it was blind: with `EnsureSiblingLookup`'s memo guard deleted, the eager scan rebuilds the entire lookup — a `Dictionary`, per-bucket lists and four jagged arrays — on all 100 calls, and the counter still read zero and the test still passed. It now uses Unity's `Is.Not.AllocatingGCMemory()` constraint, which fails on that mutation.

This means **`ZeroAllocPollPathTests.ParsingErrorCode_AllocatesNothing` has been vacuous since it was written** (`Assert.AreEqual(0, delta)` where `delta` is always 0), and `ParsingPartialJson_AllocatesOnlyTheReturnedString` likewise (`0 < 20_000`). I left that file alone: different hot path, and the partial-JSON one asserts a *budget*, which `AllocatingGCMemory` cannot express — converting it is a design call, not a mechanical swap. Worth its own issue.

**2. Gap 2's premise does not hold — the hoist is unobservable.** The issue asks for a test that fails if `ParseInternal`'s tie locals are hoisted out of the extraction loop. There cannot be one. The clear-on-adopt block in the `Better` branch resets all eight of them whenever a new incumbent is taken, and a round that appends a diagnostic entry has taken one by definition — `bestScore` starts at `float.MinValue`, so the first admissible candidate always adopts, and a round that adopts nothing `break`s before appending. Confirmed by hoisting the three exemplar locals and running both suites green.

The per-round declaration is belt-and-braces over clear-on-adopt, not the rule enforcing it. I kept the test, because it is still the first multi-round diagnostic test in which round 1 actually *records* a rival — coverage that was genuinely absent — but its comment states what it does and does not pin instead of claiming a guard it does not provide.

## Validation

**Mutation pass** — four guards deleted simultaneously, then re-run:

| Mutation | Witness | Verdict |
|---|---|---|
| value-list dedup removed | `SiblingWarning_TwoPatternsSharingAValue_...` | FAILED — `... ("mode", "mode" or "level")` |
| `alreadyReported` removed | `SiblingWarning_TwoPatternsCarryingOneCancelWord_...` | FAILED — second warning for `ident_unknown` |
| memo guard removed | `SiblingLookup_IsBuiltOncePerParser_...` | FAILED — `Expected: same as ...` reference identity |
| memo guard removed | `EagerSelection_OverASiblingTie_...` | FAILED — `Expected: not allocates GC memory` |
| tie locals hoisted | — | **no witness exists**, see Finding 2 |

`SiblingWarning_SymmetricSetWithABarePattern_NumbersEachPatternToo` also failed under mutation 1 — not collateral, but the pre-existing incidental guard described above.

**Clean run** on the exact commit under review (`dff809b`), `Runtime/Commands/VoxrCommandParser.cs` pristine, probe test deleted:

- [x] Unity EditMode — **132/132**, 0 failed
- [x] Unity PlayMode — **526/526**, 0 failed
- [x] Both suites compiled clean (no `CS####` diagnostics)
- [x] No regressions; working tree restored (`Tests~` renamed back, `.meta` orphans cleaned)
- [x] No production code changed — `git diff origin/main HEAD` touches only the two test files, 288 insertions, 0 deletions
- [x] No `CHANGELOG.md` entry: not user-facing, consistent with prior test-only issues (#68, #76)

Host project `VoXR TestGround`, Unity 6000.4.7f1. No `NativeBridge~` change, so no arm64 rebuild and no on-device check needed.

## Review fixes

The review below found no code defect. All four findings were false or unsupported *statements*; all four are fixed in `dff809b`:

1. **Overclaimed mutation coverage** — the "Each was verified…" sentence and the unmarked table row, in both this body and the commit message. Commit amended so the claim never reaches `main`; body scoped to four, table marks the fifth.
2. **Value dedup described as unguarded** — corrected above and in the test comment, with the timing that explains why the issue said otherwise.
3. **Wrong hoist rationale** in `EagerSelection_...` — the real reason is that `new[]` inside the delegate is 100 allocations inside the measured region, not a warm-up effect. `BuildCoverageTables` regrows on length, not identity.
4. **"Round 1 wins on score"** in the Editor test — it wins on the start-index key, which `CompareCandidate` tests two lines before score (`VoxrCommandParser.cs:2899` vs `:2901`), as the source itself states at `:2873`.

### DIFF
--- a/Tests~/Editor/VoxrCommandParserDiagnosticTests.cs
+++ b/Tests~/Editor/VoxrCommandParserDiagnosticTests.cs
@@ -493,5 +493,78 @@ namespace VoXR.Tests.Editor
                 "a tie we cannot phrase a question about is not one we ask about"
             );
         }
+
+        [Test]
+        public void LastParseDiagnostics_TieInRoundOne_DoesNotLeakIntoRoundTwo()
+        {
+            // The first multi-round test in which round 1 actually RECORDS a rival. The only
+            // multi-round diagnostic test before it parsed a grammar with no sibling sets, so
+            // nothing was ever recorded in either round and the hand-off between rounds went
+            // unexercised (issue #93).
+            //
+            // What this does NOT pin, said plainly because issue #93 asked for it and the
+            // premise did not survive contact: ParseInternal declares its tie locals inside the
+            // extraction loop, and hoisting them OUT is unobservable. The clear-on-adopt block
+            // in the Better branch resets all eight of them whenever a new incumbent is taken,
+            // and a round that appends a diagnostic entry has taken one by definition —
+            // bestScore starts at float.MinValue, so the first admissible candidate always
+            // adopts, and a round that adopts nothing breaks before appending. The per-round
+            // declaration is belt-and-braces over clear-on-adopt, not the rule enforcing it.
+            // Confirmed by hoisting the three exemplar locals out of the loop and re-running
+            // both suites green. So this test pins the BEHAVIOUR — round 2 reports the tie it
+            // had, which is none — and no test can pin the declaration site.
+            //
+            // Two rounds, and only the first is a coin flip. "set alpha on" drops the
+            // discriminator and ties set_mode against set_level; "cease fire" that follows it
+            // is unambiguous. The tied round comes FIRST, which is the order a leak would show
+            // in, and what puts it there is the start-index key: CompareCandidate returns on
+            // `startIdx != bestStartIdx` before it ever reads a score, so the sibling pair
+            // adopted at index 0 refuses "cease fire" at index 3 without the two scores being
+            // compared at all. Score enters only as the > 0 admission floor.
+            //
+            // Said precisely because the arithmetic invites the wrong reading: the pair does
+            // also outscore "cease fire" here (0.75 against 0.4, which pays coverage for the
+            // three tokens it skips to reach itself), and that is a coincidence this fixture
+            // does not rest on. A change that sank the pair below 0.4 would NOT reorder the
+            // rounds; one that pushed it to zero would, by dropping it from the round entirely.
+            var parser = new VoxrCommandParser(
+                ShipSlots(),
+                new[]
+                {
+                    new VoxrCommandDefinition(
+                        "set_mode",
+                        new[] { new[] { "set", "{ship}", "mode", "on" } }
+                    ),
+                    new VoxrCommandDefinition(
+                        "set_level",
+                        new[] { new[] { "set", "{ship}", "level", "on" } }
+                    ),
+                    new VoxrCommandDefinition("cease_fire", new[] { new[] { "cease", "fire" } }),
+                }
+            );
+
+            var results = parser.Parse("set alpha on cease fire", null);
+
+            Assert.AreEqual(2, results.Length, "two commands extracted, in two rounds");
+            Assert.AreEqual("set_mode", results[0].Command.Intent);
+            Assert.AreEqual("cease_fire", results[1].Command.Intent);
+
+            var diag = parser.LastParseDiagnostics;
+            Assert.AreEqual(2, diag.Length);
+            Assert.AreEqual(
+                "set_level",
+                diag[0].TiedRivalIntent,
+                "round 1 is the coin flip, and is still reported as one"
+            );
+
+            Assert.IsNull(
+                diag[1].TiedRivalIntent,
+                "round 2 had one candidate — a rival held over from round 1 would name a "
+                    + "pattern that never competed for this command"
+            );
+            Assert.AreEqual(-1, diag[1].TiedRivalPatternIndex);
+            Assert.IsFalse(diag[1].TiedRivalIsSibling);
+            Assert.IsNull(diag[1].DescribeTiedRival());
+        }
     }
 }
--- a/Tests~/Runtime/VoxrCommandParserTests.cs
+++ b/Tests~/Runtime/VoxrCommandParserTests.cs
@@ -3,8 +3,13 @@ using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using NUnit.Framework;
 using UnityEngine.TestTools;
+using UnityEngine.TestTools.Constraints;
 using VoXR;
 using VoXR.Commands;
+// UnityEngine.TestTools.Constraints.Is derives from NUnit's and adds AllocatingGCMemory().
+// Aliased rather than imported bare because both namespaces export the name; this file used
+// no `Is.` before, so the alias costs nothing and keeps the ambiguity from ever arising.
+using Is = UnityEngine.TestTools.Constraints.Is;
 
 namespace VoXR.Tests.Runtime
 {
@@ -3910,6 +3915,52 @@ namespace VoXR.Tests.Runtime
         }
 
         [Test]
+        public void SiblingWarning_TwoPatternsSharingAValue_NameThatValueOnce()
+        {
+            // The VALUE list's dedup, the sibling of the pattern-text one above and added
+            // blind alongside it (issue #93). Both became reachable when issue #90 stopped
+            // dropping duplicate-valued members, and both collapse for the same reason the
+            // intent list does — except that here the reason is about the QUESTION: the choice
+            // the speaker faces is between distinct words, so ("mode", "mode" or "level")
+            // would be wrong about what is being asked, not merely repetitive.
+            //
+            // The duplicate-valued members render DIFFERENT pattern text here — set_a reaches
+            // the frame by omitting its optional — so the pattern-text dedup does nothing and
+            // the value list is the only thing under test.
+            //
+            // Not the only guard on this branch, and issue #93 said otherwise only because it
+            // was filed hours before the other one landed: SiblingWarning_SymmetricSetWith
+            // ABarePattern_NumbersEachPatternToo asserts a full message on a set that carries
+            // "mode" twice, so it too goes red without the dedup. It gets there incidentally,
+            // through a fixture built to pin per-pattern element numbering, and it would stop
+            // covering this the moment that fixture changed for its own reasons. This one is
+            // deliberate and isolated, which is what makes it the guard rather than a second
+            // accident.
+            LogAssert.Expect(
+                UnityEngine.LogType.Warning,
+                new Regex(
+                    "Intents 'set_a', 'set_b' and 'set_c' have patterns "
+                        + "\"set mode on \\?please\" \\(with its optional elements omitted\\), "
+                        + "\"set mode on\" and \"set level on\" that differ only at element 2 "
+                        + "\\(\"mode\" or \"level\"\\)"
+                )
+            );
+
+            var parser = new VoxrCommandParser(
+                Array.Empty<VoxrSlotDefinition>(),
+                new[]
+                {
+                    Sib("set_a", SibP("set", "mode", "on", "?please")),
+                    Sib("set_b", SibP("set", "mode", "on")),
+                    Sib("set_c", SibP("set", "level", "on")),
+                }
+            );
+
+            Assert.IsNotNull(parser);
+            LogAssert.NoUnexpectedReceived();
+        }
+
+        [Test]
         public void SiblingWarning_NamesARemedyThatActuallyRemovesTheTie()
         {
             // The message shipped saying "Diverge earlier", which does not work — and this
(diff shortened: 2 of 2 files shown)


jinwoo1601 (commented):
## Independent review — PR #104

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`, plus **`XFILE` added** because the new tests bind to a private production field by reflection string, a contract crossing the assembly boundary.
**Head reviewed:** `5c4005cd1dd4a7623f0bdf8302ae56122974e627` · **Base:** merge-base `47a00905b8a228c3a39f48aa8804f42bf45ddc21`

`LINE` re-derived all five added tests against `VoxrCommandParser.cs` and found every assertion holds and every test bites its stated mutation. `GONE` and `XFILE` returned zero findings. **No code defect was found.** All four findings below are false or unsupported *statements* — in committed comments, the PR body, and the commit message.

---

### 1 · The commit message and PR body claim five mutation witnesses; four exist — minor, CONFIRMED

PR body: "**Each** was verified by deleting the code it guards and watching it go red — the failure message is quoted below so you can check it failed for the right reason." The mutation table in the same body reads `| tie locals hoisted | — | not observable, see above |`, and Finding 2 states the mutation was run and both suites stayed **green**. Four failure messages are quoted for five tests.

The "What" table compounds it: `| 2 | per-round tie reset | LastParseDiagnostics_TieInRoundOne_DoesNotLeakIntoRoundTwo |` lists that test under the gap it does not close, unmarked. No scoping precedes the table.

Same sentence is in commit `5c4005c` — "**every one** verified by deleting the code it guards and watching it go red" — four paragraphs above "cannot be pinned". **That text lands on `main` permanently at merge.**

*Fix:* scope the claim to the four with witnesses and name the fifth up front as behaviour-only coverage; amend the commit message before merge.

---

### 2 · The value-dedup branch already had a guard; the comment says it had none — minor, CONFIRMED

`Tests~/Runtime/VoxrCommandParserTests.cs:3929` tells a future reader that "the one [fixture] that does carry a duplicate asserts only the `differ only at element N` clause".

`SiblingWarning_SymmetricSetWithABarePattern_NumbersEachPatternToo` (same file, `:4248`, on `main` before this branch) registers `set_mode`/`set_level`/`set_bare` such that its three-member set carries `"mode"` twice, and asserts `\("mode" or "level"\)` at `:4264`. Delete `if (!values.Contains(quoted))` (`VoxrCommandParser.cs:1962`) and `JoinWith` renders `"mode", "level" or "mode"` — that expectation can never be met. It goes red.

Timing matters for blame, not for truth: issue #93 was filed `2026-08-16T02:25:09Z`; the symmetric test landed in `d6fa33c` ~5 h later. **The issue was right when written; the PR's restatement is stale.** The PR's own mutation table records this test failing "as collateral" without reconciling it.

*Fix:* say the branch was already incidentally guarded and that this test guards it deliberately and in isolation.

---

### 3 · The allocation test's hoist rationale describes a mechanism that cannot occur — minor, CONFIRMED

`Tests~/Runtime/VoxrCommandParserTests.cs:4686`: "a fresh array per call would measure the warm-up instead of the steady state."

It would not. `BuildCoverageTables` regrows only under `if (_recognisedPrefix == null || _recognisedPrefix.Length < n + 1)` (`VoxrCommandParser.cs:3342`) — keyed on length, not identity — and the warm-up already sized the tables to 3, so a second two-element array changes nothing. `_coverageTokens = tokens` (`:3358`) is rebound each call, so the `ReferenceEquals` guard stays green too. No cache anywhere keys on the array reference.

The real reason the hoist is load-bearing: `new[] { "set", "on" }` inside the delegate is 100 array allocations *inside the measured region*.

*Why it matters:* a maintainer who checks the stated reason, finds it false, and inlines the array gets a red from the test's own garbage — and the natural next move, relaxing to a budget, silently re-blinds the invariant.

---

### 4 · "Round 1 wins on score" — it wins on the start-index key — minor, CONFIRMED

`Tests~/Editor/VoxrCommandParserDiagnosticTests.cs:519`.

`CompareCandidate` tests start index two lines *above* score: `if (startIdx != bestStartIdx) return startIdx < bestStartIdx ? Better : Worse;` (`VoxrCommandParser.cs:2899`) then `if (candidate.Score != bestScore)` (`:2901`). `set_mode`@0 is the first candidate examined and adopts; every `cease fire` candidate sits at `startIdx 3` and is refused on start index without its score being read. Registration order does not rescue the reading — the start index decides either way.

The source already says so at `:2873`: *"the start-index key below outranks score."*

The arithmetic quoted (0.75 vs 0.4) is itself correct, and the fixture and all assertions are right. But the fixture is stable against exactly the scoring changes the comment implies would break it, and fragile against a floor change it does not mention.

---

### Notes

*Unverified, not numbered — offered for the author's judgement, not as defects.*

- The allocation guard covers `TryEagerCommit` only; `ParseInternal`'s identical `ci × pi × startIdx` loop — the other half of the allocation-free claim — stays unguarded. `Parse` allocates its results, so a zero-alloc pin there needs a different shape.
- The new Editor test constructs a grammar that logs a sibling warning it does not `LogAssert.Expect`. Consistent with every other test in that file, which asserts nothing about logs.
- The PR reports `ZeroAllocPollPathTests` as blind on this runtime and deliberately leaves it alone. `CLAIM` found the characterisation accurate (`Assert.AreEqual(0, delta)` at `:51`, `Assert.Less(delta, 20_000)` at `:34`). Worth its own issue.

### Refuted

None — all four findings survived adversarial verification. `GONE` and `XFILE` produced no findings to verify.

- **The Unity runs.** `Tests~/` is tilde-suffixed and cannot compile or execute from a bare package checkout, so no finder agent could confirm or refute the 132/132 and 526/526 figures or any mutation-pass result. Those runs were performed separately against the host project (`VoXR TestGround`, Unity 6000.4.7f1); this review checked the claims for internal consistency and static plausibility only, which is how findings 1 and 2 surfaced.
- Nothing device-facing changed, so no Quest check applies.

jinwoo1601:
Pushed `dff809b` (force-push over `5c4005c`) addressing **all four** findings from the review above. Force-push rather than a follow-up commit because finding 1 lives in the commit message, and a knowingly-false claim heading to `main` is worth rewriting history to avoid.

| Finding | Addressed by |
|---|---|
| 1 · five witnesses claimed, four exist | Commit message rewritten: "Four of them were verified by deleting the code they guard… the fifth guards something no test can witness". PR body scoped to four; the "What" table now carries a **Mutation-witnessed** column marking the fifth `no`. |
| 2 · value dedup described as unguarded | Test comment now records that `SiblingWarning_SymmetricSetWithABarePattern_NumbersEachPatternToo` already goes red without the dedup, that #93 was filed ~5 h before it landed in `d6fa33c`, and why a deliberate isolated guard still earns its place. New PR section "One correction to the issue's premise". |
| 3 · wrong hoist rationale | Replaced with the real reason — `new[]` inside the delegate is 100 allocations *inside* the measured region — plus an explicit "not for warm-up" clause noting `BuildCoverageTables` regrows on length, not identity. |
| 4 · "Round 1 wins on score" | Now attributes round order to the start-index key with the precedence spelled out, and flags the 0.75-vs-0.4 gap as a coincidence the fixture does not rest on, naming which direction of change *would* reorder the rounds. |

The mutation table's `SiblingWarning_SymmetricSetWithABarePattern_NumbersEachPatternToo` row no longer calls that failure "collateral" — per finding 2 it is the pre-existing incidental guard, which is a different thing.

Re-verified on the exact amended commit rather than assuming comment-only edits are safe: **EditMode 132/132, PlayMode 526/526**, both suites compiled clean, no `CS####` diagnostics, working tree restored. Still no production code changed — `git diff origin/main HEAD` is 288 insertions, 0 deletions, across the two test files.

Nothing from the review is left unaddressed. The one item deliberately **not** actioned is the `ZeroAllocPollPathTests` note, which is out of this issue's scope and wants its own issue.

PR #106, opened 2026-08-17
Measure the poll-path allocation tests with a live instrument (#105)

### LINKED ISSUE #105: ZeroAllocPollPathTests is built on a dead counter — both assertions are vacuous
Found while pinning issue #93's invariant 8 (PR #104), and out of scope there. Both tests in `Tests~/Runtime/ZeroAllocPollPathTests.cs` pass unconditionally and have since they were written.

## The instrument does not work

`GC.GetAllocatedBytesForCurrentThread()` **does not move** in this project's Unity runtime. Measured 2026-08-17 with a deliberate probe — allocate a 1 MB `byte[]` between two reads and assert the delta is greater than zero:

```
counter moved 0 B after a 1 MB allocation
  Expected: greater than 0
  But was:  0
```

PlayMode, Unity 6000.4.7f1, via the `VoXR TestGround` host. `ZeroAllocPollPathTests` is a Runtime suite, so it runs in exactly that context — the finding applies to it directly, not by analogy.

## Consequence

Both assertions compare a number that is always `0`:

- `ZeroAllocPollPathTests.cs:51` — `Assert.AreEqual(0, delta, ...)` in `ParsingErrorCode_AllocatesNothing`. Always true.
- `ZeroAllocPollPathTests.cs:34` — `Assert.Less(delta, PartialBudget)` where `PartialBudget = 20_000` (`:17`), in `ParsingPartialJson_AllocatesOnlyTheReturnedString`. `0 < 20000`. Always true.

Neither test can fail. The file's header comment — "anything else (Substring, boxing, freshly allocated key arrays) is a regression" — describes a guard that is not in place.

How this surfaced: the first draft of #93's allocation test used the same counter and **passed with the invariant it was pinning deleted**. Deleting `EnsureSiblingLookup`'s memo guard makes the eager scan rebuild a `Dictionary`, per-bucket lists and four jagged arrays on all 100 calls, and the counter still read `0 B`.

## The fix is not uniform — one half needs a decision

**`ParsingErrorCode_AllocatesNothing` is a mechanical swap.** It returns an enum and should allocate nothing at all, so Unity's constraint expresses it exactly:

```csharp
using UnityEngine.TestTools.Constraints;
using Is = UnityEngine.TestTools.Constraints.Is;  // NUnit also exports `Is`

Assert.That(() =>
{
    for (int i = 0; i < 100; i++)
        VoxrJsonParser.ParseErrorCode(json);
}, Is.Not.AllocatingGCMemory());
```

`Jinwoo1601.VoXR.Tests.Runtime.asmdef` already references `UnityEngine.TestRunner`, so no wiring is needed. PR #104's `EagerSelection_OverASiblingTie_AllocatesNothingPerCall` is a working example, and it was mutation-verified to actually go red.

**`ParsingPartialJson_AllocatesOnlyTheReturnedString` cannot be converted.** It asserts a *budget*, because the method legitimately allocates one string per call — that is the thing it is meant to permit. `AllocatingGCMemory` is zero-or-fail and expresses no budget, so this needs a call between roughly:

1. Restructure so the returned string is outside the measured region, then pin zero on the rest.
2. Find an instrument that does report byte counts on this runtime (`Recorder` on `GC.Alloc`, or `Profiler.GetMonoUsedSizeLong`) and keep a budget.
3. Drop the budget and pin the intent a different way — e.g. assert the returned string is the *same* instance across calls where it should be, or assert on allocation *count* rather than bytes.
4. Delete it, and rely on the error-code test plus a comment, if the budget was never really enforceable.

I have no recommendation between these — it depends on how much the partial-JSON path is worth guarding.

## Also worth checking

Whether any other test in the repo reads `GC.GetAllocatedBytesForCurrentThread()`. At the time of writing, `grep` finds it only in this file and in a comment in `VoxrCommandParserTests.cs` explaining why it is avoided.

## Where

`Tests~/Runtime/ZeroAllocPollPathTests.cs` — `PartialBudget` (`:17`), `ParsingPartialJson_AllocatesOnlyTheReturnedString` (`:20`), `ParsingErrorCode_AllocatesNothing` (`:39`).

jinwoo1601:
Closes #105

## What

`Tests~/Runtime/ZeroAllocPollPathTests.cs` moves off `GC.GetAllocatedBytesForCurrentThread` and onto Unity's GC.Alloc profiler recorder.

- **`ParsingErrorCode_AllocatesNothing`** — the mechanical swap the issue specified. It returns an enum and should allocate nothing at all, so `Is.Not.AllocatingGCMemory()` expresses it exactly.
- **`ParsingPartialJson_AllocatesOnlyTheReturnedString`** — the half the issue left open. Took **option 3, allocation count rather than bytes**; rationale below.
- `PartialBudget` (the 20 KB byte budget) is deleted.
- A comment in `VoxrCommandParserTests.cs` (~:4676) said `ZeroAllocPollPathTests` "is left alone here… reported on the PR rather than fixed in passing." This PR makes that false, so it now points at #105 instead. That is the only reason that file is in the diff — no test in it changed.

No `CHANGELOG.md` entry: `Tests~/` is not shipped, so this is not user-facing.

## Why

The issue's premise is confirmed. Both assertions compared a number that was always `0`, and had since `44d364f` (2026-04-29).

**Why option 3 for the partial-JSON test.** The deciding fact came from reading Unity's own constraint source — `AllocatingGCMemoryConstraint.cs` is *itself* built on `Recorder.Get("GC.Alloc")` + `sampleBlockCount`, which is an allocation **count**. So the count-based assertion is not a new, unvalidated instrument; it is the exact mechanism already mutation-proven in this repo by #104's `EagerSelection_OverASiblingTie_AllocatesNothingPerCall`. That ruled out option 2 (hunting for a byte-reporting instrument) as strictly more risk for less.

A count also states the intent — "one string per call, nothing else" — more precisely than the budget did, and is the **stronger** guard: an inadvertent `Substring` per call costs ~78 B x 100 = ~7.8 KB, comfortably under the old 20 KB budget. That budget would have passed the very regression it was written to catch *even on a runtime where the byte counter worked*.

Option 1 was rejected because the only input that skips the string allocation is an empty partial value, which returns `string.Empty` before the slice — it would pin the key search but not the value extraction, where the regression the test names would live. Option 4 (delete) discards a guard that turns out to be cheap to make real.

## Validation

- [x] **Unity PlayMode 526/526, EditMode 132/132**, host `VoXR TestGround` (6000.4.7f1). Both suites green, `failed="0"` with nonzero totals. No compile errors.
- [x] **Mutation-verified — this is the part that matters.** A first-try green proves nothing here; it is the exact shape of the bug being fixed. Adding one throwaway allocation per call to `ParseErrorCode` and `ParseStringValue` (stored to a static so the JIT cannot elide it, behaviour-neutral) makes both tests fail:
  - `ParsingPartialJson_AllocatesOnlyTheReturnedString` → `Expected: 100 / But was: 200`
  - `ParsingErrorCode_AllocatesNothing` → `Expected: not allocates GC memory`
  - Blast radius was exactly those two of 526 — no other test guards this path.
  - Sensitivity is to a **single one-byte array per call**, so any realistic regression is caught. The mutation is reverted; `Runtime/VoxrJsonParser.cs` is unchanged in this diff.
- [x] Confirmed no other **test** reads `GC.GetAllocatedBytesForCurrentThread` (the issue's "also worth checking" item) — no call site remains in tracked test code. Scoped deliberately: the default `grep` here honours `.gitignore`, so it cannot see `Planning~/`, where the A/B rig still reads the counter at `ab-rig/Program.cs:196,202`. That is out of scope and harmless — the rig is a `net10.0` console app running on CoreCLR, where the counter works.
- [ ] Not covered: this proves the instrument fires under **PlayMode/Mono in batch mode**. It says nothing about IL2CPP or device builds, where allocation accounting can differ.

## For the reviewer

The open decision was mine to make and is the thing worth ruling on: **`Assert.AreEqual(100, allocations)` pins an exact count, so it fails if the parse path ever allocates fewer objects too**, not just more. That is deliberate — a drop would mean `Encoding.UTF8.GetString` stopped returning a fresh string, which is worth knowing — but if you would rather it were one-directional, `Assert.LessOrEqual` is a one-line change.

### DIFF
--- a/Tests~/Runtime/VoxrCommandParserTests.cs
+++ b/Tests~/Runtime/VoxrCommandParserTests.cs
@@ -4664,18 +4664,15 @@ namespace VoXR.Tests.Runtime
             // the first null), and nothing checked.
             //
             // Measured with Unity's own AllocatingGCMemory constraint and NOT with
-            // GC.GetAllocatedBytesForCurrentThread, which is what ZeroAllocPollPathTests uses.
-            // That counter is INERT here — measured at 0 B moved after a deliberate 1 MB
-            // allocation, this Unity version's collector not being one that maintains it — so
-            // an assertion written against it reads zero whatever the code does. This test was
-            // written that way first and was blind: deleting the memo guard in
-            // EnsureSiblingLookup makes this very scan rebuild the whole lookup (a Dictionary,
-            // per-bucket lists and four jagged arrays) on all 100 calls, and the counter-based
-            // version still reported 0 B and passed. The constraint below fails on it.
-            //
-            // ZeroAllocPollPathTests is blind for the same reason and is left alone here — it
-            // is a different hot path, and its partial-JSON test asserts a BUDGET, which no
-            // constraint expresses. Reported on the PR rather than fixed in passing.
+            // GC.GetAllocatedBytesForCurrentThread. That counter is INERT here — measured at
+            // 0 B moved after a deliberate 1 MB allocation, this Unity version's collector not
+            // being one that maintains it — so an assertion written against it reads zero
+            // whatever the code does. This test was written that way first and was blind:
+            // deleting the memo guard in EnsureSiblingLookup makes this very scan rebuild the
+            // whole lookup (a Dictionary, per-bucket lists and four jagged arrays) on all 100
+            // calls, and the counter-based version still reported 0 B and passed. The
+            // constraint below fails on it. ZeroAllocPollPathTests was blind for the same
+            // reason and was moved off that counter in issue #105.
             //
             // Measured over TryEagerCommit rather than Parse, because Parse necessarily
             // allocates the results it returns and so can never be pinned at zero. The eager
--- a/Tests~/Runtime/ZeroAllocPollPathTests.cs
+++ b/Tests~/Runtime/ZeroAllocPollPathTests.cs
@@ -19,9 +19,16 @@ namespace VoXR.Tests.Runtime
     // measured at 0 B moved after a deliberate 1 MB allocation — so the byte-delta
     // assertions this file was written with read zero whatever the code does, and both
     // passed unconditionally from the day they were written (issue #105).
+    //
+    // TheAllocationInstrumentsSeeADeliberateAllocation is the positive control against a
+    // repeat of that. Read it before trusting a green run of the other two.
     public class ZeroAllocPollPathTests
     {
         const int Iterations = 100;
+
+        // Not private: a private field only ever assigned is a CS0414 warning. Written by
+        // the positive control so its allocations cannot be optimised away as dead.
+        internal static byte[] Sink;
 
         [Test]
         public void ParsingPartialJson_AllocatesOnlyTheReturnedString()
@@ -41,6 +48,10 @@ namespace VoXR.Tests.Runtime
             // exactly that one allocation. Anything beyond it — an inadvertent Substring,
             // a boxed value type, a freshly built key array — shows up as a second
             // allocation per call and fails here.
+            //
+            // Deliberately an equality, not a ceiling: a count BELOW one-per-call would
+            // mean the returned string had stopped being freshly allocated, which is also
+            // a change worth failing on.
             Assert.AreEqual(
                 Iterations,
                 allocations,
@@ -57,7 +68,9 @@ namespace VoXR.Tests.Runtime
             VoxrJsonParser.ParseErrorCode(json);  // warm up
 
             // Returns an enum (value type), so nothing at all is permitted and Unity's
-            // zero-or-fail constraint expresses it exactly.
+            // zero-or-fail constraint expresses it exactly. Looped rather than called once
+            // so that an allocation appearing only on a repeat call — a lazy init that
+            // re-fires, a cache that rebuilds — is caught too.
             Assert.That(
                 () =>
                 {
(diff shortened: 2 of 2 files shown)


jinwoo1601 (commented):
## Independent review — PR #106

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`, **plus `PIT`**, added because the diff drives Unity's process-wide profiler `Recorder` and its thread-filter state. Nothing dropped.
**Head reviewed:** `aeca8edb15116ca7e0e2a773892660c01980aecd` · **Merge-base:** `a230f0ae`

**One confirmed finding, minor, in the PR body rather than the code.** The test change itself came back clean from all four angles: the helper is line-for-line Unity's own `AllocatingGCMemoryConstraint` plumbing, the measured region correctly excludes the closure and warm-up, NUnit argument order and `Iterations` are consistent, the `Is` alias shadows nothing, and `VoxrCommandParserTests.cs` is genuinely comment-only.

---

### 1. A ticked validation box overstates what was checked — minor, CONFIRMED

The box reads *"Confirmed no other test reads `GC.GetAllocatedBytesForCurrentThread` … **it now appears nowhere in the repo**."* The first half is true and is all invariant 5 needs. The restatement after the em-dash is false, and the sweep behind it could not have established it:

```
$ grep -rn "GetAllocatedBytesForCurrentThread" .        # default grep — honours .gitignore
Tests~/Runtime/VoxrCommandParserTests.cs:4667   (comment)
Tests~/Runtime/ZeroAllocPollPathTests.cs:18     (comment)

$ command grep -rn --exclude-dir=.git "GetAllocatedBytesForCurrentThread" .
./Planning~/features/coverage-in-selection/ab-rig/Program.cs:196   long bytesBefore = …
./Planning~/features/coverage-in-selection/ab-rig/Program.cs:202   long bytesAfter  = …
```

`git check-ignore -v` → `.gitignore:44:Planning~/`. Two **live call sites** survive in the A/B rig, feeding `bytes_per_utterance` at `:210`; the default sweep cannot see them. No correctness consequence — the rig is `net10.0`/`<OutputType>Exe</OutputType>` with no NUnit reference, so it runs on CoreCLR where the counter works, and the underlying invariant holds. Fix: scope the tick to what was searched.

**Fixed on this branch** — the box has been narrowed and the rig's desktop-.NET use noted, since leaving a knowingly false claim on the PR is worse than the one-line correction. The finding stands as reported.

---

### Refuted, so you can see what was considered and killed

- **"The exact-count reading is a new, unvalidated use of the API."** Refuted. `AllocatingGCMemoryConstraint.cs:49` prints *"The provided delegate made {0} GC allocation(s)."* — Unity's own source documents `sampleBlockCount` as an allocation count, so the exact-count semantics is the vendor's, not a novel reading. The PR also attributes its exact-count evidence to its own 100→200 mutation run, not to #104.

### Notes — unverified, not defects

- All four angles independently flagged that the helper drops the `#if !UNITY_WEBGL` guards Unity wraps around `FilterToCurrentThread()`/`CollectFromAllThreads()` (`ZeroAllocPollPathTests.cs:86,95`). Kept as a Note, not a finding, because it cannot misbehave on any platform this package runs on — but it is the most-reported item in the review, so it is your call rather than mine.
- `Assert.AreEqual` pins the aggregate, not the per-call distribution: 50 calls at 0 plus 50 at 2 would pass.
- `ParsingErrorCode_AllocatesNothing` would still pass vacuously if the recorder itself ever went inert, since Unity's constraint is zero-or-fail. Worth knowing, as that is the exact failure shape #105 is about. `PIT` established the count test is self-detecting here: with the profiler compiled out, `sampleBlockCount` reads 0 and `0 != 100` **fails loudly**. The two tests differ in this durability.
- `Iterations` in the constraint-based loop is decorative — zero-or-fail does not depend on call count.
- The old byte budget covered one class of regression the count does not: the permitted string growing in *size* without the allocation count changing.

### Not verifiable from here

- The Unity suite numbers (526/526 PlayMode, 132/132 EditMode) and the mutation results were audited for internal consistency against the code, **not re-run**. The 100→200 arithmetic and the quoted NUnit message shapes check out against the source.
- **On-device (Quest) behavior** — human-only per project CLAUDE.md; nothing in this diff speaks to it.
- IL2CPP allocation accounting, which the author already flags as uncovered.

*Automated review. Comment-only: GitHub blocks self-approval in a solo repo, so this is evidence for your ruling, never the ruling.*

jinwoo1601:
## Findings addressed — pushed as `bb36914`

### Fixed

**The `#if !UNITY_WEBGL` guards** (the Note all four angles reported). Mirrored on both `FilterToCurrentThread()` and `CollectFromAllThreads()`, so the helper is now a faithful transcription of the `AllocatingGCMemoryConstraint` it cites.

**The vacuous-if-inert hole on `ParsingErrorCode_AllocatesNothing`** — the Note worth real work, because it is the *exact* failure shape of #105. A zero-or-fail constraint passes when the recorder is dead, precisely as an expected-zero byte delta passed when the counter was dead. Neither existing test can detect that about itself.

New test `TheAllocationInstrumentsSeeADeliberateAllocation` asserts the converse: **both** instruments must observe a deliberate 64 B allocation — `GCAllocationsDuring(...) > 0`, and `Is.AllocatingGCMemory()` in its positive form. It is non-vacuous by construction, asserting a positive rather than a zero, and it would have caught the original defect the day this file was written in April.

That also removes the need to mutation-test it: for a positive assertion, passing *is* the evidence the recorder is live.

### Declined, with reasons

- **Aggregate vs per-call distribution** (50 calls at 0 + 50 at 2 would pass). The input is a fixed `byte[]` and the path is branch-free with respect to call index, so per-call variance cannot arise. Measuring 100 calls individually is machinery for an unreachable case.
- **`Iterations` decorative in the constraint loop.** Kept, and the reason is now in the code: the loop catches an allocation that appears only on a *repeat* call — a lazy init that re-fires, a cache that rebuilds — which a single call would miss.
- **The old byte budget caught size growth, the count does not.** The test's subject is *what* is allocated, not how big; the string's size is fixed by the input. A size guard would be a different test.
- **Exact count vs `LessOrEqual`.** Keeping the equality, now documented inline: a count *below* one-per-call would mean the returned string had stopped being freshly allocated, which is also worth failing on.

### Validation

- **PlayMode 527/527** (was 526 — the new test), **EditMode 132/132**. All three `ZeroAllocPollPathTests` pass, including the positive control.
- No `error CS`; no `CS0414` from the new `internal static byte[] Sink` field. The only warnings are two pre-existing `CS0618`s in `Editor/VoxrDebugWindow.cs`, untouched here.
- Unchanged from before: on-device Quest verification and IL2CPP allocation accounting remain uncovered.

The confirmed finding from the review (the overstated validation tick) was corrected in the PR body earlier and is unaffected by this push.

PR #114, opened 2026-08-17
Fire OnTalkStarted when a scene starts in Continuous mode (#109)

### LINKED ISSUE #109: OnTalkStarted does not fire when a scene starts in Continuous mode
**Symptom**: A scene authored with `Listening Mode = Continuous` starts recognition on enable, but `VoxrPushToTalkController.OnTalkStarted` never fires — a recording indicator wired to it stays dark even though the mic is live.

**Root cause**: `Awake` sets `_wantRecognising = true` for the Continuous case and `OnEnable` calls `StartRecognition()` without invoking `_onTalkStarted` (`Runtime/VoxrPushToTalkController.cs:118-134`). The only invocation sites are the `ListeningMode` setter (on a runtime *change* to Continuous, `:69`) and `PressTalk` (`:96`) — so the event's documented meaning ("recognition begins") is delivered on every path except the authored-Continuous one.

**Expected**: the event fires whenever recognition begins under the controller's control, including the authored-Continuous startup.

**Notes**: this is a small behaviour change — existing Continuous scenes would start receiving one extra `OnTalkStarted` at startup — so it wants its own test. `Documentation~/push-to-talk.md` currently documents the actual (asymmetric) behaviour; update it when this lands. Surfaced by the PR #108 review (finding F2 context); ruled as a bug 2026-08-17.

jinwoo1601:
Closes #109

## What

`VoxrPushToTalkController` now fires `OnTalkStarted` when a scene authored with `Listening Mode = Continuous` starts listening.

The intent to recognise moved from `Awake` into `OnEnable`, beside the `StartRecognition()` it authorises:

- `Awake()` is gone; its `_wantRecognising = true` for the Continuous case is now the first branch of `OnEnable`, which also invokes `_onTalkStarted`.
- That branch is guarded by `!_wantRecognising`, so it runs on the *first* enable only. Every later enable falls through to the pre-existing silent-resume path.
- New internal test seam `InitialMode`, which writes the serialized field the way the Inspector does — the public `ListeningMode` setter can't author the mode, because it *is* a change and fires the event itself.

## Why

The event was invoked in exactly two places — `PressTalk()` and the `ListeningMode` setter on a runtime change to `Continuous`. The one path with neither is a scene already set to `Continuous` when it loads, so it started the mic in silence and left a recording indicator wired to the event dark for the whole session.

The existing contract is that `OnTalkStarted` fires on each false→true transition of `_wantRecognising`, and that lifecycle enable/disable does not touch that flag and is therefore silent. That rule is deliberate and pinned by three tests plus the Lifecycle section of `push-to-talk.md` — it is what keeps Continuous listening alive across the Quest home overlay. `Awake`'s transition was simply the one that never announced itself. Moving it to the site that actually starts recognition closes the gap without weakening the rule.

Deliberately **not** done: firing `OnTalkEnded` from `OnDisable`. That would contradict `OnDisable_WhilePressed_DoesNotFireOnTalkEnded` and the documented silent-resume design.

## Behaviour change

Existing `Continuous` scenes now receive one extra `OnTalkStarted` at startup. That is the fix, and the issue calls it out. Anyone who worked around this by driving their indicator off `VoxrSpeechRecogniser.IsRecognising` is unaffected — that still works, and the event is now an alternative rather than a trap.

The want-flag semantics are unchanged for every controller that actually reaches `OnEnable`, including when `Speech Recogniser` is unassigned: the flag is still set for an authored-Continuous component in that case, so a later switch to `PushToTalk` still fires `OnTalkEnded` exactly as before.

**One further state does change** — the state where `Awake` and `OnEnable` genuinely differ, and the review caught this claim overreaching before it was corrected. A controller authored on `Continuous` whose *component checkbox* is unticked used to get the flag anyway, because Unity runs `Awake` on a disabled component of an active GameObject but never `OnEnable`. Switching such a controller to `PushToTalk` before it had ever been enabled therefore fired an `OnTalkEnded` that no `OnTalkStarted` had ever paired with, and called `StopRecognition()` on a recogniser that had never been started. That is now silent, and the announcement waits for the tick that actually starts recognition. Two tests pin both halves.

One caveat, now documented: the event fires from `OnEnable`, during scene load. Inspector-wired listeners always receive it (they are serialized with the component). A listener added from code must subscribe before the controller is enabled — `AddListener` from another component's `OnEnable` or `Start` is too late.

## Validation

- [x] Unity EditMode suite, host project `VoXR TestGround` (Unity 6000.4.7f1): **132/132 passed, 0 failed**
- [x] Unity PlayMode suite: **530/530 passed, 0 failed** — this is the platform that covers `Tests~/Runtime`, where the controller tests live
- [x] Three new PlayMode tests, all passing: `AuthoredContinuous_FirstEnable_FiresOnTalkStarted`, `AuthoredContinuous_DisableEnableCycle_DoesNotRefireOnTalkStarted`, `AuthoredContinuous_ThenSwitchToPushToTalk_FiresOnTalkEndedOnce`. The third is the guard on the `Awake` → `OnEnable` move: it fails if the enable announced the event but forgot to record the flag.
- [x] All three pre-existing lifecycle tests still pass unchanged (`OnDisable_WhilePressed_DoesNotFireOnTalkEnded`, `OnDisable_InContinuousMode_PreservesWantRecognisingFlag`, `OnEnable_WhenWantRecognisingTrue_DoesNotFireOnTalkStarted`), as does the whole `VoxrPushToTalkPauseTests` fixture
- [x] Suites re-run after a late rename of the test seam, so the green above is against this exact commit, not an earlier draft
- [x] No `NativeBridge~/` changes, so the bridge build and desktop harness do not apply
- [ ] **On-device (Quest) check needed:** this reworks `OnEnable`, and `OnApplicationPause` delivery on doff/system-menu cannot be verified from WSL. The PlayMode suite covers the state machine, not whether Android actually delivers the callback.

## Note for review

`Runtime/VoxrPushToTalkController.cs` shows three pre-existing `internal` setter properties reflowed from one-liners to block bodies. That is the repo's CSharpier hook reformatting the file on edit, not an intentional change — manual de-indents are silently reverted by it. No semantic content in those three lines differs.

## Docs

`Documentation~/push-to-talk.md` previously documented the asymmetry as intended behaviour; that note is replaced by the subscribe-before-enable timing caveat, the `On Talk Started` row is corrected, and the Lifecycle section gains the first-enable exception. `CHANGELOG.md` gets a `[Unreleased] → Fixed` entry that leads with the extra startup event.


jinwoo1601 (commented):
## Review — independent pass on `0512a63`

**Depth profile:** issue lane → **slim** (`CLAIM` + `LINE`, `GONE`), **plus `PIT`** — added because the diff deletes `Awake()` and reworks `OnEnable()` on a MonoBehaviour, which is the lifecycle trigger the profile names. No angle was dropped. `LINE` died mid-run on a session limit and was re-run to completion; its result below is from the completed re-run, not the aborted one.

Four finder angles ran against the diff with no context from the session that wrote it, then every numbered finding was adversarially verified by a separate agent. **One finding confirmed, two refuted.** The confirmed one is about the PR body's own wording, not the code — the implementation itself came back clean from all four angles.

---

### 1. The PR body's "want-flag semantics are otherwise unchanged" is inaccurate — minor · CONFIRMED

The *Behaviour change* section claims the flag "is still set for an authored-Continuous component … so a later switch to `PushToTalk` still fires `OnTalkEnded` exactly as before. Only the event is new." That sentence is unqualified, and one reachable state contradicts it.

Base `Awake()` ran for a component whose **own enabled checkbox** is off — Unity calls `Awake` on a disabled component of an active GameObject, and skips `OnEnable`. The write now lives in `OnEnable`:

```csharp
if (_listeningMode == VoxrListeningMode.Continuous && !_wantRecognising)   // :131
    _wantRecognising = true;                                               // :133
```

So a Continuous controller shipped with its checkbox unchecked now has `_wantRecognising == false` where it was `true`, and the untouched reader at `Runtime/VoxrPushToTalkController.cs:74` (`if (previous == VoxrListeningMode.Continuous && _wantRecognising)`) stays silent where base fired `OnTalkEnded` and called `StopRecognition()`.

The verifier rated the **code** direction an improvement: base fired an `OnTalkEnded` unpaired with any `OnTalkStarted`, and drove `vosk_bridge_stop()` on a session that had never started. The defect is the claim, not the behaviour. No shipped doc repeats the overclaim. No test covers this state in either revision.

**Fix:** narrow the body sentence to enabled controllers, or state the pre-enable delta alongside the "one extra `OnTalkStarted`" note. Optionally pin it with a test that toggles `controller.enabled` rather than the GameObject.

---

- **`push-to-talk.md:60`'s `On Talk Started` row is false when `Speech Recogniser` is unassigned.** Refuted: the row's head clause — "fired **whenever the controller starts recognition**" — governs the list, so with no recogniser nothing starts and the row correctly predicts silence. `Speech Recogniser` is also declared *Required* at `:55`.
- **Firing the event from `OnEnable` creates a startup-ordering hazard the doc recommends users into.** Refuted: serialized fields are populated by deserialization *before* `Awake`, so an Inspector-wired listener of the documented shape is safe — the package's own `PushToTalkDemo.ShowRecording` (`Samples~/PushToTalk/PushToTalkDemo.cs:112-116`) just sets `Image.color` on a `[SerializeField]` reference. The suggested remedy (defer the announcement to `Start`) would reopen a one-frame form of #109.

---

### Notes

Unverified, one line each — below the numbered-finding floor, listed for your judgement, not fixed.

- The replaced doc paragraph dropped a still-valid workaround: author `PushToTalk` and assign `ListeningMode = Continuous` from your own `Start()`. It is simpler than the "keep the GameObject inactive until you have subscribed" remedy the new caveat offers, and answers the same problem.
- `Samples~/PushToTalk/PushToTalkDemo.cs:39` subscribes from its own `OnEnable` — the exact pattern the new doc calls "too late". Latent only: the sample ships `_listeningMode: 1` (PushToTalk), so it never takes the new path.
- The doc's "`AddListener(...)` from another component's `OnEnable` … is too late" is absolute, but cross-GameObject `OnEnable` order is merely unspecified and is steerable via Script Execution Order. "Cannot be relied on" would be more accurate; `Start` is genuinely always too late.
- The new Lifecycle bullet ("The one enable that is *not* silent…") is likewise unqualified — that enable is silent when no recogniser is assigned.
- The event also fires from `Object.Instantiate` of an authored-Continuous prefab, before the caller can touch the returned instance; the caveat frames the timing as "during scene load" only.
- `AuthorContinuousInactive()` builds a **real** `VoxrSpeechRecogniser`, not the `FakeSpeechRecogniser` double that the sibling `VoxrPushToTalkPauseTests` fixture exists to keep native calls out of. Consistent with this fixture's existing pattern, but worth knowing.
- Pre-existing, untouched here: the `ListeningMode` setter uses `_speechRecogniser?.StartRecognition()` (`:68`, `:77`), and `?.` bypasses Unity's overloaded `==`, so a destroyed recogniser throws `MissingReferenceException` there.
- `Tests~/Runtime/VoxrPushToTalkPauseTests.cs` carries `VoxrPushToTalkController.cs:<line>` citations in comments that were already ~9 lines stale at the merge-base; this diff adds ~30 lines to that file and pushes them further out.

---

### What the angles verified clean

`LINE` re-derived all eight reachable `OnEnable` entry states against base `Awake()`+`OnEnable()` and found exactly one intended divergence (the fix). The `return` at `:141` is necessary — without it the non-null case would call `StartRecognition()` twice on one enable. No sequence double-fires `OnTalkStarted`; branch 1 cannot re-arm without an intervening `OnTalkEnded`. `OnApplicationPause`/`Update` see the same flag value at the same moment as base for any enabled component. All three new tests fail against their respective mutations — none is vacuous — and `_authoredGo` can neither leak nor be double-destroyed. The CSharpier reflow of the three `internal` setters was checked, not assumed: semantically identical.

### Not verifiable from here

- **On-device (Quest) behaviour** — whether Android delivers `OnApplicationPause` on doff/system-menu, real mic capture. Human-only; there is no device rig. The PR correctly leaves this box unticked, and it is the one deferral that matters for a diff reworking `OnEnable`.
- **Re-running the Unity suites** — this is a bare UPM package; the suites run only through the host project. The ticked EditMode 132/132 and PlayMode 530/530 claims were audited for scope and plausibility (PlayMode is correctly identified as the platform covering `Tests~/Runtime`), not reproduced.
- **Real Editor script-execution order** in a built scene — reasoned from Unity's documented rules only.

*Comment review — this lane's reviewer cannot approve or request changes. Nothing here has been fixed or ruled on.*

jinwoo1601:
## Review response — `ff36742`

All findings from the review above are addressed. The confirmed finding was about a claim, not the code, so it took three fixes rather than one.

**Finding 1 — "want-flag semantics are otherwise unchanged" (CONFIRMED, minor)**

- **PR body corrected.** The sentence is now scoped to controllers that actually reach `OnEnable`, and the divergent state is spelled out rather than swept under "only the event is new".
- **Behaviour pinned by two tests** — `AuthoredContinuous_ComponentDisabled_AnnouncesOnlyWhenEnabled` (announces nothing while unticked; announces exactly once when ticked) and `AuthoredContinuous_ComponentDisabled_SwitchToPushToTalkIsSilent` (the delta itself). The helper gained a `componentEnabled` parameter so the component checkbox is testable as an axis distinct from `SetActive`.
- **Documented** in the Lifecycle section of `push-to-talk.md` and in the CHANGELOG, since it is user-visible: a controller authored on `Continuous` but shipped unticked used to fire an `OnTalkEnded` unpaired with any `OnTalkStarted`, and stop a recogniser that had never started. That is now silent.

**Notes — six addressed**

- Restored the dropped `Start()`-assignment workaround, and named `IsRecognising` as the timing-free alternative. The caveat now lists three reliable ways to catch the startup event instead of two.
- `Samples~/PushToTalk/README.md` now explains that `PushToTalkDemo` subscribes in its own `OnEnable`, and why that is safe only because the scene ships in push-to-talk mode.
- "…from another component's `OnEnable` or `Start` is too late" → `Start` is always too late; `OnEnable` **cannot be relied on**, because cross-GameObject order is undefined unless pinned in Script Execution Order.
- The Lifecycle bullet is qualified with "provided a `Speech Recogniser` is assigned".
- The caveat now covers `Instantiate` of an authored-Continuous prefab, not only scene load.
- Five stale `VoxrPushToTalkController.cs:<line>` citations in `VoxrPushToTalkPauseTests.cs` renumbered to their current targets. They were already ~9 lines off at the merge-base; this diff had widened the drift.

**Two notes deliberately not acted on**

- `AuthorContinuousInactive()` builds a real `VoxrSpeechRecogniser` rather than the `FakeSpeechRecogniser` double. That double is a private nested class of the *pause* fixture; this fixture has used a real recogniser in every test predating this issue, and its existing tests already drive `StartRecognition()` the same way. Copying the double across would duplicate it and break the fixture's convention for no behavioural gain.
- The `?.` on `_speechRecogniser` in the `ListeningMode` setter (`:68`, `:77`) bypasses Unity's overloaded `==`. Real, but genuinely pre-existing and untouched by this diff — it wants its own issue, not a smuggled fix in a review-response push.

**Re-validation** — Unity EditMode **132/132**, PlayMode **532/532**, zero failures, no `error CS`. The +2 is exactly the two new tests; `VoxrPushToTalkPauseTests` still 8/8, confirming the comment-only edit there is inert. Result XMLs verified fresh (`22:10:39Z` / `22:11:11Z`), with the prior run's artifacts moved aside first so a stale file could not be misread as a pass.

On-device (Quest) verification remains deferred and human-only.

PR #116, opened 2026-08-17
Convert Inspector assets when Slot Assets is empty (#110)

### LINKED ISSUE #110: Inspector asset conversion is silently skipped when Slot Assets is empty
**Symptom**: Wire `VoxrCommandRecogniser` entirely in the Inspector with command-set assets but an empty **Slot Assets** array (a legitimate all-literal grammar — `cease fire`, `weapons mode`) and nothing is ever recognised. No log, no error.

**Root cause**: `Awake`'s asset-conversion path returns early when `slotAssets == null || slotAssets.Length == 0` (`Runtime/Commands/VoxrCommandRecogniser.cs:445-449`), and again on empty `commandSetAssets`, with no diagnostic on either path. The slot-array emptiness check looks like it should be a null check — a slot-free grammar needs no slot assets.

**Expected**: an empty (non-null) Slot Assets array converts normally with zero slots; and whenever conversion is skipped while command-set assets are assigned, a warning names why.

**Notes**: `Documentation~/inspector-authoring.md` currently documents the both-arrays-required behaviour; update it when this lands. Surfaced by the pre-release docs audit (#107); ruled as a fix 2026-08-17.

jinwoo1601:
Closes #110

## What

`VoxrCommandRecogniser.Awake()` no longer treats an empty **Slot Assets** array as "nothing wired". Slot assets are simply optional now — absent or empty, the array converts with zero slots — so an all-literal grammar (`cease fire`, `weapons mode`) can be authored entirely in the Inspector.

That leaves exactly one asset-list skip, and it no longer returns in silence: an empty **Command Set Assets** logs a warning naming itself — but only when **Slot Assets** carries assets. With neither list assigned nothing is logged: that is the code-driven case, where `Configure()` is expected to follow from `Start()`, and warning there would second-guess every code-configured component. A `Configure()` call landing before this component's `Awake()` also skips conversion, silently and by design; that skip is now documented rather than changed.

- `Runtime/Commands/VoxrCommandRecogniser.cs` — the two guards, reordered so the command-set check runs first (it is the list that must be populated).
- `Tests~/Runtime/VoxrAssetConversionTests.cs` — five PlayMode tests: the all-literal conversion end-to-end (`InjectText` fires the command) for both an empty and an absent slot array, the warning path, and the stays-silent path in both its empty-array and null-array shapes. They build the component on an inactive GameObject and write the private serialized fields before activating, since `Awake` runs on activation — the only order that reproduces what the Inspector produces.
- `Documentation~/inspector-authoring.md` — documented the both-lists-required behaviour; corrected. `Documentation~/api/command-recogniser.md`'s Inspector-field table said nothing either way; it now states the rule.
- `CHANGELOG.md` — entry under `[Unreleased]` → Fixed.

## Why

A slot-free grammar needs no slot assets, but `Awake` returned before converting anything when `slotAssets` was empty — and then also skipped the command sets, so neither `Configure()` nor `SetActiveSets()` ran. Nothing was logged on either path, so the symptom was a recogniser that heard speech and recognised no command at all, with no diagnostic pointing at the wiring.

Nothing about conversion, parsing, or scoring changes for a grammar that does declare slots.

## Validation

- [x] Unity **EditMode** suite: 132/132 passed, 0 failed (host project `VoXR TestGround`, Unity 6000.4.7f1, against this working tree)
- [x] Unity **PlayMode** suite: 537/537 passed, 0 failed — includes all five new tests, each confirmed present in the result XML
- [x] No compile errors or warnings in either run
- [x] `NativeBridge~/` untouched, so the bridge build and desktop harness are not implicated

## Not in scope

`Editor/VoxrBatchTestWindow.cs:106-139` applies the same both-arrays-non-empty rule, so an all-literal grammar still cannot be batch-tested in the Editor window. That path is **not** silent — it already names what is missing — so it is a different defect from the one this issue reports, and is left for a follow-up rather than widened into here.

## Needs a human eye

- The warning wording and the decision to stay silent when *neither* list is assigned.
- No on-device check is required: this touches `Awake` only, and the PlayMode suite covers that state machine per the project verification bindings.


jinwoo1601 (commented):
## Independent review — PR #116

**Depth profile:** issue lane → **slim** (`CLAIM` + `LINE`, `GONE`), widened with **`PIT`** (the diff edits a MonoBehaviour lifecycle method and the new tests depend on activation ordering) and **`XFILE`** (the tests bind to `[SerializeField]` names as reflection strings, and behavioural prose in two docs must agree with runtime code in a third). No angle was dropped.
**Head reviewed:** `1f24c025e6542b3a05a72e7e597a7cfff827d052` · **base (merge-base):** `51080bf`
Five finder agents ran with no context from the session that wrote the code; every numbered finding below was then passed to a separate verifier told to refute it. **The core guard rewrite verified clean** — re-derived across all eight input combinations of the two arrays, the only cell whose skip/convert decision changes is the one the PR declares, and three of the four new tests genuinely fail against the base revision. Both surviving findings are inaccuracies in prose this PR introduces; neither is a code defect.

---

### 1. `Documentation~/inspector-authoring.md:77` — "whenever … a warning names which list is at fault" is false for one skip path
**minor · CONFIRMED** · surfaced independently by `CLAIM`, `LINE`, `GONE`, `XFILE`

The new paragraph asserts "Whenever one list is populated and the conversion is skipped anyway, a `Debug.LogWarning` names which list is at fault," then enumerates the silent case exhaustively as "With *neither* list assigned nothing is logged." A third skip is neither — `Runtime/Commands/VoxrCommandRecogniser.cs:441-443`:

```csharp
// If user code already called Configure(), _slots is non-null and inspector assets are ignored.
if (_slots != null)
    return;
```

Both lists can be fully populated here, conversion is skipped, nothing is logged. Reachable: `Configure()` sets `_slots` (`:238`, `:270`) and depends only on inline-initialised collaborators, so it completes before this component's `Awake` — called from a lower Script-Execution-Order component, or on an inactive GameObject before `SetActive(true)`. The author sees no warning and concludes the wiring took. §"Code vs Inspector Priority" (`:114`) describes the *overwrite* ordering, not this silent skip.

**Fix:** scope the sentence to the two asset-list guards, and name the prior-`Configure()` skip as silent by design. The sibling row at `Documentation~/api/command-recogniser.md:21` already scopes it correctly ("it warns when `slotAssets` was populated").

### 2. `CHANGELOG.md:51` misstates the package's own documentation history
**minor · CONFIRMED** · surfaced by `CLAIM`

The shipped entry says: "`Documentation~/inspector-authoring.md` and `api/command-recogniser.md` documented the both-lists-required behaviour and have been corrected." The API file never documented that rule. Its base rows were bare —

```
| `slotAssets` | `VoxrSlotAsset[]` | -- | Slot definitions for Inspector authoring |
| `commandSetAssets` | `VoxrCommandSetAsset[]` | -- | Command set definitions for Inspector authoring |
```

(`51080bf:Documentation~/api/command-recogniser.md:20-21`) — and no line anywhere in that base file asserts either list must be populated. Only `inspector-authoring.md:77` carried the rule, which is also what issue #110 itself says. The API-table edit is an **addition** of previously-undocumented behaviour, not a correction of a false statement. The same sentence appears in the PR body, so it is wrong in two artifacts, one of which ships to users.

**Fix:** "…`inspector-authoring.md` documented the both-lists-required behaviour and has been corrected; `api/command-recogniser.md`'s Inspector-field table, which said nothing either way, now states the rule."

---

### Considered and refuted

- **"Empty Slot Assets now throws `ArgumentException` out of `Awake` on upgrade, and the CHANGELOG denies it."** Refuted on four counts: that exception is already documented as an expected `Awake()` error for exactly this wiring at `inspector-authoring.md:104-106`, on both revisions and untouched by the diff; the throw needs `initialActiveSetNames` non-empty, which the finding omitted; the end state is *identical to base* (`_parser` null, recogniser silent), so no working configuration regresses — the only new console error is for a configuration that was already 100% non-functional, which is what #110 asks for; and the codebase distinguishes *declare* from *reference* (the parser's own message is "references undefined slot"), under which the CHANGELOG's "a grammar that does declare slots" is exactly true.

### Notes

Unverified, below the numbered-finding bar — nothing here asserts the merged result is wrong.

- The `slotAssets == null` branch (`VoxrCommandRecogniser.cs:461-469`) looks unreachable in production: Unity materialises a serialized array as empty rather than null, both fields are private with no setter, and a runtime `AddComponent` leaves *both* null so the earlier guard returns first. Its user-facing warning string may therefore be unseeable, and `Awake_NullSlotAssets_WarnsWhenCommandSetAssetsAssigned` certifies a state only reflection can build. Flagged by four angles.
- `Awake_NoAssetsAssigned_StaysSilent` passes identically against the base revision — a regression guard, not fix evidence. Its comment names the code-configured (`null`/`null`) shape, but the fixture builds the Inspector-deserialized (empty/empty) shape; the `null`/`null` route was re-derived as silent too, but no test pins it.
- The two new warnings pass no context `Object` to `Debug.LogWarning`, so in a scene with several recognisers the console line cannot ping the offending GameObject. Consistent with the five pre-existing warnings in the same method, so this would be a package-wide change, not a fix to this diff.
- Command Set Assets populated with **Initial Active Set Names** left empty still leaves `_parser` null and recognises nothing, with no log — the same silent symptom #110 reports. Pre-existing and unchanged by this PR, now also reachable for all-literal grammars.
- `Editor/VoxrBatchTestWindow.cs:106-139` still gates on both arrays being non-empty, so the all-literal grammar the docs now bless cannot be batch-tested in the Editor window. The PR declares this out of scope; that justification checks out — the window renders a "Missing: …" `HelpBox`, so unlike `Awake` it was never silent — and no document is made false by leaving it.

### Verified clean

Reflection strings resolve to real fields under the stated `BindingFlags`; `InternalsVisibleTo("Jinwoo1601.VoXR.Tests.Runtime")` matches the asmdef name, so the `internal` setters are reachable; every new `using` is covered by an existing asmdef reference; both test regexes are substrings of the emitted warnings; the new Markdown anchor resolves to a real heading. No other file in the repo still asserts the old both-lists rule, and the shipped sample scene (`Samples~/CommandRecognition/CommandRecognition_Tactical.unity`) has both arrays empty, so it lands on the stays-silent branch and gains no new warning noise. Unity's `==` overload is used correctly throughout — plain reference checks on the arrays, the `UnityEngine.Object` overload on the per-entry checks. `LogAssert.NoUnexpectedReceived()` does catch warnings, not only errors, so the silence assertions are real.

- **The reviewers could not re-run the Unity suites.** This is a bare UPM package with no host project reachable from the review working directory, and the tests live under the hidden `Tests~`. The PR's two ticked suite boxes (132/132 EditMode, 536/536 PlayMode) were assessed by derivation only — the agents checked whether each test *could* discriminate the fix from the bug, not that the numbers are real.
- No compilation was performed here.
- On-device (Quest) behaviour — not implicated by this diff, which touches `Awake` only.

jinwoo1601:
### Review findings addressed — `0e267d7`

All findings from the review above are now closed, at the user's direction (the issue lane would otherwise have left the two minor ones for you).

**Finding 1 — `inspector-authoring.md:77` "whenever … a warning names which list is at fault".** Fixed. The sentence is scoped to the one asset-list skip that remains, and the prior-`Configure()` skip is now named explicitly as silent by design, with the two orders that reach it (a script earlier in Script Execution Order; `Configure()` on an inactive GameObject before `SetActive(true)`) and a link to *Code vs Inspector Priority*. The code was not changed — that skip is the documented precedence, so the doc was the thing that was wrong.

**Finding 2 — `CHANGELOG.md` misstating the doc history.** Fixed, and in both artifacts: the entry now reads "`inspector-authoring.md` documented the both-lists-required behaviour and has been corrected; `api/command-recogniser.md`'s Inspector-field table, which said nothing either way, now states the rule." The PR body carried the same sentence and has been edited too.

**Note — the unreachable `slotAssets == null` branch** (raised by four angles). Acted on rather than annotated: the branch and its warning are deleted, and the conversion loop now runs over `slotAssets?.Length ?? 0`, so an absent slot array converts to zero slots exactly like an empty one. This removes a user-facing warning string no user could trigger, and it makes the rule uniform — slot assets are simply optional. `Awake_NullSlotAssets_WarnsWhenCommandSetAssetsAssigned` is gone with it; `Awake_NullSlotAssets_ConvertsAllLiteralCommandSets` replaces it and asserts the conversion instead.

**Note — `Awake_NoAssetsAssigned_StaysSilent` naming the code-configured case but building the Inspector shape.** Fixed both ways: its comment now says which shape it builds, and a second test `Awake_NullAssetArrays_StaySilent` covers the `null`/`null` route an `AddComponent`-built component actually takes, which nothing had pinned.

**Note — no context `Object` on the warnings.** Not done, deliberately. The five pre-existing `Debug.LogWarning` calls in the same method also omit it, so passing `this` on only the new one would be locally inconsistent; making the console line pingable is a package-wide change worth its own issue, not a rider on this fix.

**Note — `initialActiveSetNames` empty still leaves `_parser` null and recognises nothing, silently.** Not done. It is pre-existing, unchanged by this PR, and reachable identically before it — a separate silent-failure defect, and a follow-up issue candidate rather than scope creep here.

**Note — `Editor/VoxrBatchTestWindow.cs` still gates on both arrays.** Not done, as declared. It is not silent (it renders a `Missing: …` HelpBox), so it is a different defect; a follow-up issue is the right home for it.

**Re-validated after the changes** — same host project and Unity version, fresh result XMLs, assemblies confirmed recompiled rather than cached:

- Unity **EditMode**: 132/132 passed, 0 failed
- Unity **PlayMode**: **537/537** passed, 0 failed — all five tests present by name; the deleted test confirmed absent from both the run and the source
- No compile errors and **no compiler warnings** — in particular no unused-using or unused-field warning from the test edits
- Test-harness revert clean; `NativeBridge~/` untouched

Still not verifiable from here, and not claimed: on-device Quest behaviour. Nothing in this change set touches that surface.

PR #108, opened 2026-08-17
Sweep the user docs against the code before release (#107)

jinwoo1601:
## Summary

Pre-release documentation sweep. Seven read-only audit agents verified every user-facing doc against source — each claim traced to file:line — and the ~70 confirmed findings were then applied. All relative links and anchors were re-verified after the edits (scripted check, 0 errors).

**Critical fixes**
- `command-sets.md`'s central claim was false: set switching removes **pattern literals only** — slot values, alias keys, digit vocabulary, and follow-up vocabulary stay in the grammar whichever sets are active, so the page's own `fire`/`five` example was a collision the feature cannot fix. The page now states what leaves, what stays, and what to expect (no in-set confidence gain, per the package's own measurement).
- Every tuning snippet in `command-recognition.md` assigned private `[SerializeField]` fields and did not compile; they are Inspector instructions now. A documented runtime `freeSpeechMode` toggle that does not exist is removed (the mode is authoring-time).
- `KNOWN_LIMITATIONS.md` was missing the `[FormerlySerializedAs]` prefab-instance-override upgrade hazard — whose symptom is indistinguishable from the coverage-demotion entry it sits beside.

**High-severity themes**
- The #73 completeness refusal (`required slot unfilled`) is now documented at every point it acts: pipeline diagram, `scoring.md` gates (+ diagnosis-table row), `OnUnrecognisedSpeech` firing lists, batch runner.
- #95 tie diagnostics: the batch window's tie line is documented; `scoring.md` no longer claims losing candidates are recorded nowhere, and no longer prescribes `disambiguateSiblingTies` for non-sibling ties (it provably cannot answer them); `KNOWN_LIMITATIONS.md` gains the duplicate-pattern entry (second intent permanently dead, no warning).
- Authoring workflow: `?word` optional-literal syntax added to every syntax enumeration; the construction-time warnings get an "Authoring warnings" section; the silent no-op when either asset array is empty, the three authoring-typo exceptions, minWords/maxWords bounds, and the missing Inspector-assignment steps in both quick starts are all documented.
- Mechanism corrections: `Configure`'s grammar-apply condition; `ReleaseNativeResources` non-owner behaviour; native-bridge build guide now installs the built `.so` (and gains an ABI exports table); `OnTalkStarted`'s actual firing behaviour; KNOWN_LIMITATIONS mechanism fixes (phantom `[CommandDemo]` log marker, slot-initial blast radius, >6-optional fallback reading, warning re-emission count, dead `v2.6` version labels).
[...]

PR #117, opened 2026-08-18
Warn on uppercase and punctuated alias keys (#111)

### LINKED ISSUE #111: Validation warnings skip uppercase and punctuated alias keys
**Symptom**: Declare an alias with an uppercase or punctuated key — `{ "Alpha", "one" }` — and it can never match: VOSK outputs lowercase and the alias lookup is ordinal. No validation warning fires, unlike the identical mistake in a slot *value*.

**Root cause**: `RunValidationWarnings` (`Runtime/Commands/VoxrCommandParser.cs:683-727`) applies the uppercase and punctuation checks inside the `slot.Values` loop only; the `slot.Aliases.Keys` loop emits just the single-character warning (`:715-726`).

**Expected**: alias keys get the same uppercase and punctuation warnings values get.

**Notes**: the docs (`api/command-definitions.md`, `api/scriptable-objects.md`, `command-recognition.md`) currently state the trap explicitly — soften them to match once the warning exists. Surfaced by the PR #108 review verification of finding F3; ruled as a fix 2026-08-17.

jinwoo1601:
Closes #111

## What

`RunValidationWarnings` (`Runtime/Commands/VoxrCommandParser.cs`) now runs the uppercase and punctuation checks over `slot.Aliases.Keys`, not just `slot.Values`. Alias keys get the same three checks values get; the existing single-character alias warning is unchanged and still fires last in the loop.

Three tests in `Tests~/Runtime/VoxrCommandParserTests.cs`: an uppercase key warns, a punctuated key warns, and a well-formed key logs nothing at all (`LogAssert.NoUnexpectedReceived()` with nothing queued) — so the new checks can't quietly become noise on every well-authored grammar.

Docs that stated the gap as a hand-checked trap are softened: `command-recognition.md`, `api/command-definitions.md`, `api/scriptable-objects.md`, and — beyond the three the issue named — `inspector-authoring.md`, whose *Authoring warnings* list enumerated "Four checks" with value-only uppercase/punctuation bullets and would otherwise have gone stale. `CHANGELOG.md` entry under `[Unreleased]` → `Fixed`.

`KNOWN_LIMITATIONS.md` needed no edit: it only ever claimed the single-character warning exists, which is still true.

## Why

Declare `{ "Alpha", "one" }` and the alias can never fire — VOSK emits lowercase and strips punctuation, and the lookup the key feeds is ordinal — yet the identical mistake in a slot *value* has always warned. The asymmetry was an accident of the checks living inside the `slot.Values` loop; the alias loop below it emitted only the single-character warning.

Surfaced by the PR #108 review verification of finding F3, ruled as a fix 2026-08-17.

## Validation

- [x] Unity PlayMode: **540/540 passed, 0 failed** (537 → 540; the three new tests are in the Runtime/PlayMode assembly and all passed)
- [x] Unity EditMode: **132/132 passed, 0 failed** (unchanged count — no EditMode tests added)
- [x] Both runs on Unity 6000.4.7f1 via host project `VoXR TestGround`, against this working tree; result `start-time` verified to postdate the `Tests~` → `Tests` rename, so neither XML is a stale leftover. Working tree reverted cleanly afterwards (no `Tests/`, no orphan `.meta`).
- [x] No existing fixture regresses: every alias key across the demo grammar, samples, and test fixtures is already lowercase and punctuation-free, so no previously-silent construction starts warning.
- n/a On-device (Quest) check: this is parser-side validation logging only — no mic capture, lifecycle, or native-bridge surface. `NativeBridge~/` untouched.


jinwoo1601 (commented):
## Independent review — PR #117

**Depth profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`. No angle added or dropped: the diff crosses no contract boundary (`XFILE`) and touches no MonoBehaviour lifecycle, coroutine, or main-thread-only API (`PIT`).
**Head reviewed:** `83132f417d33346916b7709b8c5434b3fd790078` · **merge-base:** `3fb68c3`

Finder agents inherited nothing from the session that wrote the fix. `CLAIM` and `LINE` returned no findings; `GONE` returned two, one of which cleared the materiality floor and was adversarially verified.

---

### Finding 1 — CONFIRMED · minor · docs

**The merged single-character bullet prescribes a remedy that negates itself.**

`Documentation~/inspector-authoring.md:91`

```
- **Single-character value or alias variant** -- one character is too little for reliable recognition; for a value, declare an alias instead (`a` -> `one`), and for a variant, prefer a longer, phonetically distinct word.
```

The base had two bullets and only the *value* one prescribed a fix; the variant one gave a reason and stopped. Merging them put the value remedy and its own negation in one sentence: `a` → `one` **is** a one-character variant, so the second clause condemns exactly what the first clause tells the author to write. That shape is sanctioned and shipped — `Slot_Quantity.asset:23`, `CommandDemo.cs:61`, `KNOWN_LIMITATIONS.md:116` — and `VoxrCommandParser.cs:709` prescribes it while `:742` warns on the result, which `VoxrCommandParserTests.cs:4130` records as expected, not a defect.

Suggested replacement: keep the value remedy, then say the resulting one-character variant warns in turn and that this is expected and harmless.

---

### Notes

Unverified; below the materiality floor — nothing here asserts the merged result is wrong.

- The two API alias rows (`api/command-definitions.md:53`, `api/scriptable-objects.md:18`) dropped the still-true "such a key never matches" alongside the now-false "only the single-character case is warned about". The replacement says only that a warning fires, and neither alias row carries the `#authoring-warnings` link its value row has.
- `char.IsPunctuation` fires on an apostrophe, so a plausible alias key like `o'clock` now warns. Consistent with the value check, but no doc names that case.
- `char.IsPunctuation` misses `+ < = > | ~ $ ^` and backtick (Math/Currency/Modifier symbols). Shared verbatim with the pre-existing value check, so not introduced here.
- No test pins the combined case (a key uppercase *and* punctuated *and* single-character), nor the new warnings' order relative to the pre-existing single-character one.
- The API rows say "the same three counts" although the single-character *message* still differs between values and aliases (pre-existing, deliberate).
- `Documentation~/troubleshooting.md:44` ("Verify patterns and slot values are **lowercase**") still omits alias keys. Pre-existing; not made false by this diff.

---

### What cleared

- **Runtime block re-derives correctly.** `key != key.ToLowerInvariant()` is an ordinal `string` compare, silent on already-lowercase, empty, digits and caseless scripts, and warns on `İ` (U+0130) where invariant lowering changes length. The `break` exits only the char loop, so one punctuation warning per key. `key` can never be null — `VoxrSlotDefinition.cs:48` and `VoxrSlotAsset.cs:48` throw at the `Dictionary` indexer first.
- **The pre-existing single-character alias warning is byte-identical** to base, and the entire value loop is untouched. `RunValidationWarnings` is still unguarded by `UNITY_EDITOR`, so the docs' "player builds as well as the Editor" claim holds.
- **No fixture starts warning.** Every alias key in the repo — `a`, `a one`, `h one`, `h two`, `jackals`, `red`, `red dragon`, `orange`, plus the sample `.asset` variants — is lowercase and punctuation-free. The ordered `LogAssert` sequences in `VoxrCommandParserTests.cs:4135` and `VoxrCommandRecogniserInjectionTests.cs:731` use `a`/`jackals`, on which neither new check fires, so their order and count are unchanged.
- **The three new tests would compile and their regexes match the emitted strings** (`-` in `hi-lo` is a literal outside a character class). Their fixtures cannot trip another construction-time scan: `WarnOnDroppableRequiredLiteral` returns early at `patternCount < 2`, the single pattern has no optional elements, and `_siblingSets` is empty — so `LogAssert.NoUnexpectedReceived()` is satisfiable in all three.
- **No stale doc statement survives** anywhere in `Documentation~/`, `README.md`, `KNOWN_LIMITATIONS.md`, `Samples~/`, or the `Runtime/` XML comments. The `#authoring-warnings` anchor still resolves and the `Aliases` `null` note survived.

### Verification status of the ticked boxes

Better than "unverified": the reviewer could not re-run Unity, but the run artifacts survive in the host project and were read directly. `TestResults/PlayMode-i111.xml` reads `passed="540" failed="0"`, `EditMode-i111.xml` reads `passed="132" failed="0"`, and the preceding `PlayMode-i110b.xml` reads `537` — matching the claimed 537 → 540 delta, with all three new tests present at `result="Passed"`. `start-time` postdates the `Tests~` → `Tests` rename, so neither XML is a stale leftover, and no `Tests/` directory or orphan `.meta` remains.

**Not verifiable from here:** on-device (Quest) behavior — correctly marked `n/a` by the author, since this is log-only parser validation with `NativeBridge~/` untouched. The repo has no test or lint CI (`release.yml` only), so no CI evidence exists for any claim.

### Disposition

No blocker, so nothing was fixed on the branch — per the issue lane, everything below blocker stays here unruled for you. Finding 1 is a one-line doc edit if you want it before merge.

jinwoo1601:
Pushed `161df47`, addressing every item from the review above.

**Finding 1 (CONFIRMED, minor) — fixed.** `inspector-authoring.md:91` no longer negates itself: the bullet keeps the value remedy (`a` → `one`) and then says the one-character *variant* it produces warns in turn, that this is expected, and why. `command-recognition.md:175` carried the same tension four lines under its own `{ "a", "one" }` sample and got the same treatment — it was a line this PR already touched, so leaving it inconsistent with the bullet would have been a new mismatch.

**Notes — five of six addressed:**

- API alias rows regain the still-true "never matches" alongside the corrected warning claim, and now carry the `#authoring-warnings` link their value rows always had.
- The punctuation guidance names the concrete fix — write `oclock`, not `o'clock` — which is what an author hitting the apostrophe case needs.
- `troubleshooting.md:44` now reads "patterns, slot values, and alias keys are **lowercase** and punctuation-free" instead of omitting aliases.
- The "same three counts" phrasing that implied identical wording is gone; the rows now enumerate the three checks instead.
- Coverage: `ValidationWarnings_AliasKeyTrippingTwoChecks_WarnsInValueLoopOrder` pins warning **order** — the two new checks fire in front of the pre-existing single-character one, which is exactly what the ordered `LogAssert` queues in `VoxrCommandRecogniserInjectionTests` depend on. Worth recording that the review's framing of this note was not reachable as written: **no key can trip all three checks**, since a one-character key is either an uppercase letter or a punctuation mark and never both. The two reachable pairs (`X` → uppercase + single-character, `.` → punctuation + single-character) are pinned instead, one per slot, because slot order is array order while `Dictionary` order within a slot is unspecified.

**Note deliberately not actioned:** `char.IsPunctuation` missing `+ < = > | ~ $ ^` and backtick. Fixing it means widening the check to `char.IsSymbol`, and that predicate is shared verbatim with the **value** check — so the change would alter pre-existing value-side behaviour and could start warning on user grammars that have been silent for releases. That is a behaviour change beyond #111's scope and wants its own issue and its own ruling. Say the word and I'll file it.

**Re-verified after the fixes:** Unity EditMode 132/132, PlayMode **541/541**, zero failures, on 6000.4.7f1 via `VoXR TestGround`. +1 is the new order test. No regression in the ordered-`LogAssert` fixtures — `VoxrCommandRecogniserInjectionTests` 62/62 and `SiblingWarning_DemoGrammar_WarnsOnlyOnTheReachableTie` both pass, and the test assembly was confirmed recompiled (`CacheMiss`) rather than served from cache, so the new regexes really did compile. Result XMLs `start-time`-checked against the run launch; tree reverted clean.

`Runtime/` is unchanged since the reviewed commit — all of the above is docs and tests.

PR #118, opened 2026-08-18
Record tie diagnostics in the session log (#112)

### LINKED ISSUE #112: Session log: record tie diagnostics and complete the readme's path list
Two session-log gaps, same file (`Editor/VoxrDebugSessionLog.cs`), one PR:

1. **Tie diagnostics never reach the export.** #95's `TiedRival` / `TiedRivalIsSibling` (`Runtime/Commands/VoxrMatchDiagnostics.cs:58,66`) are rendered in the debug window and the batch runner's per-row diagnostics but are absent from `AttemptDto` (`Editor/VoxrDebugSessionLog.cs:286-297`) — so in the artifact pitched for whole-playtest analysis, a registration-order coin flip reads as an ordinary win. Add both fields and bump `schemaVersion` (currently `1`, `:26`); additive, so existing consumers keep working.

2. **The embedded `readme` lists four of the five synthetic-attempt paths** (`:36-40`): it omits the disambiguation-answer path (`chosen via vocabulary, now awaiting confirmation`). `Documentation~/scoring.md`'s "Reading a session log" table has all five — bring the string level with it.

When this lands, update `Documentation~/editor-testing.md`, which currently states tie detection is window-only. Surfaced by the pre-release docs audit (#107) and the PR #108 review; ruled 2026-08-17.

jinwoo1601:
Closes #112

## What

**1. Tie diagnostics reach the export.** `AttemptDto` gains two fields, copied in `BuildEntry` from the `VoxrMatchAttempt` fields #95 already added:

- `tiedRival` — the equally-good rival the attempt beat on registration order alone, as `intent (pattern N)`. Empty (`""`, matching the file's null-becomes-empty convention) when nothing tied it.
- `tiedRivalIsSibling` — whether that rival was one dropped word apart (speech ambiguity) rather than a duplicate or overlapping pattern (a grammar defect).

`schemaVersion` goes `1` → `2`. Both fields are additive, so a consumer written against `1` still reads the file.

**2. The embedded `readme` lists all five synthetic-attempt paths.** It described four, folding confirm/cancel and follow-up slot-fill into one clause and omitting the disambiguation answer (`chosen via vocabulary, now awaiting confirmation`). It now matches `Documentation~/scoring.md`'s "Reading a session log" table. The same string also documents the two new fields, since the log is meant to be self-interpreting for external tooling.

**3. Docs.** Three places asserted tie detection was window-only and are now false:

- `Documentation~/editor-testing.md:68` — "Tie detection is not in the log."
- `KNOWN_LIMITATIONS.md:709` — "The exported session log carries no tie fields, so the diagnostic is window-only."
- `Documentation~/scoring.md:18` — the Vocabulary table's **Winner** row, "though not in the exported session log, which records winners only." Missed by the first commit and caught by the review below; fixed in `1f95570`.

`Documentation~/scoring.md` gains the two rows in its session-log field map (§3, selection) and a clause in each of its two tie symptom rows, which previously pointed only at the Editor panel.

Nothing about parsing, selection, or firing changes.

## Why

The session log is the artifact pitched for whole-playtest analysis, and it was the one surface the #95 tie line did not reach. A registration-order coin flip is identical to a clean win in every other exported field — same `score`, same `accepted`, same empty `rejectReason` — so a permanently dead intent behind a duplicate pattern, or the wrong sibling firing, read there as a healthy match.

## Validation

- [x] **Unity EditMode** (host project `VoXR TestGround`, Unity 6000.4.7f1, this working tree): 133/133 passed, 0 failed. The new `BuildEntry_RecordsTiedRivalAndWhetherItWasASibling` is present in the result XML and passed; +1 test vs. the previous run on this host, which is exactly this test.
- [x] **Unity PlayMode**: 541/541 passed, 0 failed — unchanged, as expected for an Editor-only change.
- [x] New test covers both tie kinds, the untied case (`tiedRival` is `""`, not null), and asserts both fields appear in `JsonUtility.ToJson` output — `JsonUtility` serialises public fields only, so that assertion is what proves the fields actually reach the file. The three cases — sibling tie, non-sibling tie, no tie — are asserted at both the DTO and the JSON level. (Review finding 3: this box originally said "round-trip", which overstates a one-way serialise-and-inspect, and claimed an untied case the new test did not then have; both corrected in `b00d2b3`.)
- [x] CSharpier clean on both edited `.cs` files.
- [x] `NativeBridge~/` untouched, so the native build and desktop harness were not run.
- No on-device (Quest) check applies: `Editor/` is compiled out of builds and nothing here touches runtime, mic capture, or MonoBehaviour lifecycle.


jinwoo1601 (commented):
## Independent review — PR #118

**Profile:** issue lane → **slim** (`CLAIM`, `LINE`, `GONE`), **plus `XFILE`** — added because the diff changes serialized field names and a schema version that form a contract described across four files. `PIT` not run (no MonoBehaviour lifecycle, coroutine, or main-thread API in the diff).
**Head reviewed:** `09e91cd6094d68134dce56b89a07384fd328db7d` · **base:** merge-base `403e665`.

Four finder agents ran against the diff with no context from the session that wrote it, then each numbered finding was adversarially verified. Three findings CONFIRMED, one REFUTED and demoted. The code change itself — `AttemptDto`, the `BuildEntry` copy, the `SchemaVersion` bump, and the new test — traced clean on every angle; all three findings are about prose.

---

### 1. `Documentation~/scoring.md:18` still says the session log carries no tie information — **blocker**, CONFIRMED

Found independently by `CLAIM`, `GONE`, and `XFILE`, and again by an unrelated verifier. The Vocabulary table's **Winner** row is untouched by this PR, in a file the PR did edit:

> One exception: a rival that *tied* the winner exactly is named on the attempt (`Tied with:` in the Editor's last-match panel and in batch-runner diagnostics), though not in the exported session log, which records winners only.

That is now false — `Editor/VoxrDebugSessionLog.cs:236` copies `tiedRival = a.TiedRival ?? ""` into a public, `JsonUtility`-serialized field — and it contradicts this PR's own new rows 465 lines later (`:483`, `:499`). It is the **third** instance of exactly the claim the PR set out to retire, so the sweep found two of three; the PR body's "Two places" undercounts by one. It is also the first thing a reader hits in that document, so a tooling author is told the field does not exist before reaching the field map. A full `.md` sweep confirms three sites total, no fourth.

**Fix:** name all three surfaces in that clause, keeping the still-true winners-only rule — the rival is named *on* the winner's attempt, never *as* one.

### 2. The `readme`'s `tiedRivalIsSibling: false` branch calls a benign same-intent tie a grammar defect — **minor**, CONFIRMED

New prose in `Editor/VoxrDebugSessionLog.cs:53-56` says `false` means "a duplicate or overlapping pattern, which is a grammar defect". But `VoxrCommandParser.cs:1845-1853` rejects any **same-intent** pair before sibling analysis, so a command's own second phrasing records `false` while being neither duplicate nor defective. A live test pins exactly that shape: `Tests~/Editor/VoxrCommandParserDiagnosticTests.cs:259-283` — one intent `set_mode`, patterns `set {ship} mode on` / `set {ship} level on`, `TiedRivalIntent == "set_mode"`, `Assert.IsFalse(...TiedRivalIsSibling)`. The field's real meaning is "one dropped word apart **and cross-intent**".

The conflation is pre-existing canon, but hedged there ("such as"); this PR states it categorically, and repeats it at `Documentation~/editor-testing.md:68`. The PR's own new `scoring.md` field-map row stops short of "grammar defect" — internally inconsistent. **Fix:** make the `false` branch non-diagnostic and point at the discriminator already in the JSON — compare `tiedRival`'s intent against the attempt's `intent`.

### 3. A ticked validation box claims a `JsonUtility` round-trip that does not exist — **minor**, CONFIRMED

The box reads "New test covers both tie kinds, the untied case … and a `JsonUtility` round-trip". There is no round-trip: `grep FromJson` over `Tests~/Editor/VoxrDebugSessionLogTests.cs` returns nothing; the test calls `ToJson` and asserts with `StringAssert.Contains` (`:144`, `:222`). And the untied case is not in the new test — both its attempts pass non-null rivals (`:116`, `:129`); the untied asserts were appended to the pre-existing `BuildEntry_NullStringsBecomeEmpty` (`:88-89`). The merged doc comment repeats it: `:99` — "hence the round-trip assertions rather than DTO reads alone."

The box's underlying purpose is met — one-way `ToJson` + `Contains` genuinely proves the fields reach the JSON, which is what matters for a public-fields-only serialiser. Only the technique label and the attribution are wrong. **Fix:** say "assertions on the serialised JSON", and attribute the untied case to the test that carries it.

---

### Refuted

- **`scoring.md`'s duplicate-pattern symptom row names `tiedRivalIsSibling: false`, a value every untied attempt also carries.** The mechanism is real, but the sentence is true in its context — the row's premise is that a tie occurred, it opens as a back-reference to the sibling row above it, and the field map 17 lines up already states `tiedRival` is empty when nothing tied. A tightening opportunity, not an error. Demoted to Notes.

### Notes

Unverified, one line each — none of these assert anything merged is wrong.

- `Documentation~/editor-testing.md`'s enumeration of an attempt's exported fields was not extended with the two new ones; the paragraph below it covers them, so this may be deliberate.
- Nothing anywhere pins `schemaVersion`, so the 1→2 bump is unguarded against an accidental revert (pre-existing gap, not introduced here).
- The JSON assertions cover only the `true` case; `tiedRivalIsSibling: false` and an empty `tiedRival` are never asserted at the JSON level.
- The `readme`'s "speech ambiguity the runtime can offer as a choice" overstates what `true` guarantees — the parser records a sibling rival whether or not it can be offered as a choice. Phrasing inherited near-verbatim from #95's own XML doc.

- **The EditMode/PlayMode pass claims.** This is a bare UPM package; `Tests~/` cannot run standalone and no host Unity project is reachable from a review agent. Those boxes were audited for internal plausibility only — the named test exists, sits in the EditMode folder, and compiles against the 11-arg `VoxrMatchAttempt` constructor — not re-run. The CSharpier box **was** independently re-run and is clean.
- Actual `JsonUtility` output, which was reasoned about from the documented public-fields-only rule rather than executed.
- On-device (Quest) behavior — not applicable, `Editor/` is compiled out of builds.

jinwoo1601:
Pushed `1f95570` addressing **finding 1** (the blocker): `Documentation~/scoring.md:18`'s Winner row now names all three surfaces — the Editor panel, the batch-runner diagnostics, and `tiedRival`/`tiedRivalIsSibling` in the exported log — while keeping the winners-only rule the sentence was carrying (the rival is named *on* the winner's attempt, never as one of its own). The CHANGELOG's count of corrected sites goes two → three, and the PR body's "Two places" above should be read as three.

Findings 2 and 3 are left unfixed and unruled, per the issue lane's blocker-only rule. Docs-only push — no C# changed, so the EditMode/PlayMode runs recorded above still cover the code as-is.

jinwoo1601:
Pushed `b00d2b3`, addressing the remaining findings.

**Finding 2** (readme's `false` branch called a benign same-intent tie a grammar defect) — verified the mechanism directly before rewriting: `VoxrCommandParser.cs:1844-1853` rejects same-intent pairs before any sibling analysis, and `LastParseDiagnostics_SameIntentTie_RecordsANonSiblingRival` pins the resulting shape. The `false` branch now names both cases and points at the discriminator already present in the JSON — `tiedRival`'s intent against the attempt's own `intent`. Corrected in all **four** places carrying the conflation, one more than the finding cited: the `readme` string, `editor-testing.md`, `scoring.md`'s field map, and the CHANGELOG entry. The same rewrite drops Note 4 — the claim that a sibling tie is one the runtime *can offer as a choice*, when the parser records it whether or not it can be phrased as a question (`VoxrCommandParser.cs:2261-2264`).

**Finding 3** ("round-trip") — the doc comment at `VoxrDebugSessionLogTests.cs:99` now says assertions on the serialised JSON. The test also gains the untied attempt it was described as covering, so sibling / non-sibling / no-tie are all asserted at both the DTO and JSON level; the box's original claim is now true rather than merely corrected.

**Notes also addressed:** `editor-testing.md`'s field enumeration mentions the tie fields; `scoring.md`'s non-sibling symptom row names the pair (non-empty `tiedRival` + `false`) instead of the bare boolean every untied attempt carries — the refuted finding, taken as the tightening the verifier said it was.

**Not done, deliberately:** nothing pins `SchemaVersion`. It is a private const with no other reader; a test asserting it equals its own literal catches only a deliberate edit and could not have caught the doc drift this branch existed to fix. Left as found — a real but pre-existing gap.

**Re-verified:** EditMode 133/133, PlayMode 541/541, zero failures, `BuildEntry_RecordsTiedRivalAndWhetherItWasASibling` passing with the widened assertions. CSharpier clean. Working tree reverted clean.

PR #119, opened 2026-08-18
Floor the follow-up re-score at zero (#113)

### LINKED ISSUE #113: Pin whether a negative follow-up score can reach a handler
Both normal firing paths floor candidates at `Score <= 0` before anything reaches a subscriber (`Runtime/Commands/VoxrCommandParser.cs:2314`, `:2856`). `ScoreFollowUp` (`:3658-3724`) — the re-score used when follow-up speech completes a pending command — has no such floor, and nothing pins whether a zero-or-negative follow-up score is reachable (a required-slot miss carries `-1.0`, so the arithmetic can go negative in principle).

**Task**: write the test that tries to construct a non-positive follow-up score reaching `OnCommandConfirmed`/`OnCommandRecognised`. If reachable, add the same floor the flush paths have; either way, keep the test as the pin.

Surfaced by the pre-release docs audit (#107, an A3 auditor question); ruled as worth pinning 2026-08-17.

jinwoo1601:
Closes #113

## Answer to the question the issue asked

**Yes — a non-positive follow-up score can reach a handler, and I measured it doing so: `-0.1111` delivered through `OnCommandConfirmed`.** So this PR does both halves of the task: the test that tries to construct it, and the floor.

## What

- `VoxrCommandRecogniser.ProcessParsedResultsCore` — the `Score <= 0f` floor both flush paths carry (`CompareCandidate`'s first test and `ParseInternal`'s `bestScore` check), restated in the follow-up arm. A non-positive merged command is refused instead of fired, the pending is left exactly as it stood, the utterance is reported unrecognised, and the Editor diagnostic carries `follow-up re-score <n> <= 0`.
- Three tests: the arithmetic pin (`ScoreFollowUp` returns `-1/9`), the reachability pin (a non-positive score reaches neither `OnCommandConfirmed` nor `OnCommandRecognised`), and a pin that the pending survives the refusal unmodified and un-cancelled.
- `CHANGELOG.md` under `[Unreleased] → Fixed`; new reject-reason row in `scoring.md`'s session-log table (and its count, five → six).

## Why

The issue noted the arithmetic can go negative in principle. It can, but the reason it becomes *reachable* is a definition-resolution split that the issue did not name:

- `IsIncomplete` — the gate deciding the merged command fires — resolves the intent through `CommandSetManager`'s dictionary, which `BuildLookup` fills **last-write-wins** (`CommandSetManager.cs:90`).
- `ScoreFollowUp` resolves the intent by scanning the parser's command array and **breaking on the first match** (`VoxrCommandParser.cs:3679`).

Register two definitions under one intent — nothing rejects that — and the two halves read different patterns for the same command. The short one calls it complete; the long one charges it for five required slots the matched pattern never had. With one definition per intent the two agree and a completed fill always re-scores exactly `1.00`, which is why nothing had caught this.

Floored rather than reconciled, deliberately: the floor holds whatever the two halves disagree about, and a non-positive score is not fireable for any reason. `scoring.md` §1 already states the rule without qualification.

Refusing rather than re-arming: the merged command is complete by slots, so `AdvanceSlotFill` would install a pending with nothing left to fill — one `TryFollowUpSlotFill` declines forever and `FireAsIs` would eventually fire carrying this same score.

## Validation

- [x] Unity EditMode (`Tests/Editor`), host `VoXR TestGround` @ 6000.4.7f1 — **133/133**, `failed="0"`.
- [x] Unity PlayMode (`Tests/Runtime`) — **544/544**, `failed="0"`; exactly +3 over the 541 on this host before the change, so the new tests were collected rather than silently dropped.
- [x] **Mutation-verified.** A green pin that never ran against the bug proves nothing, so I neutralised the floor and re-ran PlayMode: 542/544, with both new pending-command tests failing — `fired with Score -0.1111 / Expected: False / But was: True`, and `the pending survives the refusal / Expected: True / But was: False`. That number is the reachability evidence, and it matches the −1/9 the parser-level test predicts. The parser-level test passed under the mutation, correctly — the floor sits downstream of the arithmetic, not in it. Guard restored byte-exact afterwards.
- [x] `NativeBridge~/` untouched, so the bridge gates do not apply.
- [ ] On-device (Quest) check: **not needed** — the change is in `ProcessParsedResultsCore`, not mic capture or the `OnApplicationPause` delivery path. Stated for completeness, not deferred work.

## Note for the reviewer

The root cause — two definitions under one intent resolving to different patterns — is left standing. Making `BuildLookup` first-wins would align the two, but it changes intent resolution for every other consumer (`RequiresConfirmation`, `ComputeUnfilledSlots`, the flush-path completeness gate, the sibling-rival lookup), which is well beyond what #113 asked for. Worth its own issue if you want it closed at the root; the floor is correct independently of it.


jinwoo1601 (commented):
## Independent review — PR #119

**Depth profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE` — **plus `PIT`**, added because the diff inserts an early `return` into a MonoBehaviour method that raises public events and manages pending state. No angles dropped.
**Head reviewed:** `e5ab2de08bf0a6cbae90132ae77b81bf09b296bb` · **merge-base:** `df2a24f720ecd46f7f4cf2d7e6b7986ff372e8bd`

Five numbered findings, all CONFIRMED by adversarial verification. **F1 is a behavioral regression this PR introduces** and three of the four angles found it independently. The runtime change was otherwise clean: `PIT` cleared the `#if UNITY_EDITOR` bracketing, the nullable-struct reads, culture formatting, and event re-entrancy; `LINE` cleared the `<= 0f` boundary against both flush floors, the NaN path, what the early return skips, and all the new tests' arithmetic, and confirmed neither new recogniser test is a tautology.

---

### F1 — The floor sits above the completeness split, so it discards a partial fill's progress · **major** · confidence high · CONFIRMED

`Runtime/Commands/VoxrCommandRecogniser.cs:789` — the guard is the first statement of the Step 5 arm, but `bool followUpIncomplete = IsIncomplete(...)` is only computed at `:819`. So the guard also catches fills that are *not* complete, and its own comment at `:781-782` ("The merged command is complete by slots") is an unestablished premise — false on that path. Base routed those to `AdvanceSlotFill`, which keeps the filled slot, recomputes `UnfilledSlots`, restarts the `pendingTimeout` clock, and raises `OnCommandPending` — issue #77's explicit "a refusal to fire rather than a refusal to progress" (`:816-818`).

Verified by execution, not inspection: the runtime sources were compiled standalone and driven with the short definition given one more required slot (`["launch","{weapon}","target","{target}","{fuse}"]` last, the 9-element pattern first):

```
utterance1 "launch missiles target": score 0.2000, pending UnfilledSlots=[target, fuse]
utterance2 "hotel one":  merged score = -0.1111, IsIncomplete(merged) = True
  BASE -> AdvanceSlotFill: ReEnteredPending, newUnfilled=[fuse], CreatedTime 0->7, keptTarget=True
  HEAD -> guard fires, returns; unfilled still [target, fuse]
  then "impact" -> NULL (stuck); "hotel one" -> -0.1111 forever
```

An incremental multi-slot exchange can no longer progress at all and dies at the original timeout however promptly the speaker answers. Neither new test covers this — both use the pattern that makes the fill complete.

**The obvious fix is insufficient.** Moving the floor below the `:819` split is necessary but not sufficient: `AdvanceSlotFill` *stores* the merged −1/9 command, and that stored score then reaches `OnCommandConfirmed` by two further routes — `FireAsIs` on timeout (`PendingCommandHandler.cs:348-351`) and a confirm word on a `PartialMatch` pending (`PendingCommandHandler.cs:162-166`), which fires `pending.Command` without re-testing anything. The current placement is airtight against those only because it refuses everything. A complete fix has to keep the re-arm reachable *and* stop a non-positive score being stored — flooring the `Complete` arm while retaining the pending's prior positive score across the re-arm. That store hole is pre-existing at base, so closing it is a genuine scope decision, which is why this is left for your ruling rather than patched.

**Bounded, not spurious:** reachable only under the duplicate-intent divergence — with one definition per intent, `CompareCandidate`'s admission rule makes `ScoreFollowUp <= 0` arithmetically impossible after any fill. But that is the same shape this PR builds its own justification and fixture on.

### F2 — `scoring.md`'s new row claims only `pendingTimeout` ends the pending; five other things do · **minor** · confidence high · CONFIRMED

`Documentation~/scoring.md:472` states "The pending is left exactly as it stood, and only `pendingTimeout` ends it." The refusal returns with the pending live and `Reason == PartialMatch`, and Step 1 runs `TryHandleConfirmCancel` on *any* live pending with no reason filter (`VoxrCommandRecogniser.cs:663-670`). Verified reachable: a cancel word (`PendingCommandHandler.cs:116-117`, "cancel first, under every reason"); a confirm word (`:162-166`), which *fires* the original command; preemption by a complete new command (`VoxrCommandRecogniser.cs:859-860`); the public `CancelPendingCommand()` (`:325`); and replacement by the next partial-match pending (`PendingCommandHandler.cs:71-73`). The same false claim is repeated in the code comment at `VoxrCommandRecogniser.cs:784-785`.

### F3 — The "Five pipeline events" count this PR bumped in one place is now stale in two others · **minor** · confidence high · CONFIRMED

The PR correctly changed `scoring.md:464` five → six, but the identical count and enumeration live at two more sites, both untouched and now false:
- `Editor/VoxrDebugSessionLog.cs:35` — "Five pipeline events instead publish a single synthetic attempt", enumerating exactly five. This constant **ships inside every exported session log** (`:141`, `readme = Readme`).
- `Documentation~/editor-testing.md:64` — "Five pipeline events (`no match`, confirm/cancel, an answer to a disambiguation prompt, follow-up slot-fill, pending timeout)".

No test pins either string. `CHANGELOG.md:28`'s "four of the five pipeline events" is #112's historical record and is correct as written — do not touch it.

### F4 — The new `OnUnrecognisedSpeech` firing case is absent from every doc enumeration of when that event fires · **minor** · confidence high · CONFIRMED (established during F4/F5 verification; sites quoted, not separately re-verified)

`VoxrCommandRecogniser.cs:803` adds an unconditional new firing case, and it appears in none of the three places that enumerate them: `Documentation~/command-recognition.md:595` ("This happens in four situations"), the table at `Documentation~/scoring.md:271-279`, and `Documentation~/api/command-recogniser.md:44`. This holds for *every* grammar reaching that line. The PR updated only scoring.md's session-log reject-reason table.

### F5 — `OnUnrecognisedSpeech` fires unconditionally on the refusal path, against the file's own suppression rule · **nit** · confidence medium · CONFIRMED, theoretical

`:803` fires unconditionally, while Step 7's tail suppresses it for threshold-filtered results (`:1054-1055`, "the user said a valid command, just not confidently"). The in-code rationale at `:787-788` ("what Step 6 would say for it anyway") holds only when `resultCount == 0`; Step 6 is gated on exactly that (`:863`), so it is not the governing precedent when the utterance also parsed. Verification **narrowed the finding**: the debounce (`:958`) and sibling-tie (`:988`) sites are provably unreachable here — both require the predicate that sets `hasCompleteNewCommand`, which gates the whole arm at `:763`. Only the `minConfidence` site (`:946`) can co-occur, and only in a grammar that already has duplicate intents. Dropping the event instead was refuted — it would leave the integrator with no signal, against the pin at `Tests~/Runtime/VoxrCommandRecogniserInjectionTests.cs:958`.

---

- `Tests~/Runtime/VoxrPendingCommandTests.cs:1504` uses editor-only `EditorPendingCommand` outside the `#if UNITY_EDITOR` region (`:1104-1174`) that wraps four identical earlier uses; fine under the bound PlayMode-in-editor run, and line 1372 already does the same pre-existing — convention drift, not a new hazard.
- No test covers F1's partial-fill-with-non-positive-score path.
- The duplicate-intent fixture may trip a construction-time grammar warning; the new tests use neither `LogAssert.Expect` nor `NoUnexpectedReceived`. Harmless (warnings do not fail), but unpinned.
- Duplicate intents are accepted silently at `Configure` time with no warning at all — a separate issue candidate, declared out of scope for this review.

- **On-device (Quest) behavior** — never checkable from WSL. Not implicated: the change is in `ProcessParsedResultsCore`, not mic capture or the `OnApplicationPause` path.
- **The Unity test counts** (133/133 EditMode, 544/544 PlayMode, 542/544 under mutation). These came from a host project the review agents could not re-run, so they are author claims here. `CLAIM` did find them internally consistent — +3 matches three new `[Test]` methods, 544−2 = 542, and `-1/9` renders as `-0.1111` under `F4` and `-0.11` under `F2` — and confirmed by reading that neither new recogniser test is a tautology.

*No finding was softened for having been written by the session that authored the PR. Nothing here is a ruling: per the issue lane, findings below blocker stay unfixed for you to decide on. Not approved, not merged.*

jinwoo1601:
Pushed `fec7fa3`, which addresses **all five** findings from the review above — including F1, which I had left for your ruling.

**F1 (major).** The floor now sits below the completeness split and requires `!followUpIncomplete`, so it refuses only a command about to fire; an incomplete fill goes on to `AdvanceSlotFill` and the exchange advances as #77 intends.

Moving it was necessary but not sufficient, exactly as the verifier warned: `AdvanceSlotFill` *stores* the merged command, and `Complete`, the confirm-word arm and `FireAsIs` on timeout all deliver a pending's command without re-testing its score. So the floor is now enforced at the store too — `AdvanceSlotFill` retains the pending's prior admissible score rather than keeping a non-positive one. The fill is still kept: this floors the score, not the progress. When the command finally completes, `ScoreFollowUp` recomputes from the full slot set and the retained value is discarded unread.

`VoxrCommand.WithScore` is new and deliberately lives inside the struct — `_registeredSlotNames` is private, and rebuilding through the public constructor would silently drop it and change what `GetSlot` reports for a registered-but-unmatched slot.

**F2.** `scoring.md` now lists the six real endings; the same false claim in the code comment is rewritten.
**F3.** `Editor/VoxrDebugSessionLog.cs:35` and `Documentation~/editor-testing.md:64` both corrected — the first ships inside every exported session log.
**F4.** The new firing case added to `command-recognition.md`, `scoring.md`'s table, and the API reference.
**F5.** A new `anyConfidenceFilteredNewCommand` flag, computed in the Step 4 loop, suppresses the refusal path's `OnUnrecognisedSpeech` exactly where Step 7 already suppresses it. The debounce and sibling-tie sites need no counterpart — both sit behind the predicate that sets `hasCompleteNewCommand`, which skips the whole arm.

### Validation

- EditMode **133/133**, PlayMode **545/545** — +1 over the previous 545-expected count, matching the one new test.
- New test `FollowUpFill_PartialAndNonPositive_StillProgressesAndCanComplete` drives the whole three-utterance exchange: partial fill re-arms (kept slot, second `OnCommandPending`, only `{fuse}` outstanding), the stored score stays positive, and the final answer fires at `+1/9`. That last step was impossible before this push — the exchange stalled to timeout.
- **Mutation-verified again.** Disabling only the new clamp produced exactly one failure, on the retention assertion: `Expected: greater than 0.0f / But was: -0.111111112f`. Worth noting the other two follow-up tests are blind to that mutation — they assert observable behaviour and never read the stored score — so the new assertion is carrying its own weight rather than duplicating them.

The one thing I did **not** change is the root cause: two definitions under one intent still resolve to different patterns with no warning at `Configure` time. That remains a separate issue candidate, as the review noted.

Not approved, not merged — still yours to rule on.

jinwoo1601:
Filed the root cause as #120 — the first/last divergence itself (`CommandSetManager.BuildLookup` last-wins vs `ScoreFollowUp`'s first-match scan), with the other four `TryLookupCommand` consumers it reaches, `requiresConfirmation` among them.

The floor in this PR holds regardless of why the two disagree, so #120 is not a blocker for merging this — but they should be read together.

PR #121, opened 2026-08-18
Guard the ListeningMode setter with an explicit null check (#115)

### LINKED ISSUE #115: ListeningMode setter uses ?. on the recogniser, bypassing Unity's destroyed-object check
**Symptom**: switching `VoxrPushToTalkController.ListeningMode` after the assigned `VoxrSpeechRecogniser` has been destroyed calls into the destroyed component instead of skipping it, and throws `MissingReferenceException` rather than doing nothing.

**Root cause**: the `ListeningMode` setter is the only place in the file that reaches the recogniser through the null-conditional operator:

```csharp
// Runtime/VoxrPushToTalkController.cs:68  (PushToTalk -> Continuous)
_speechRecogniser?.StartRecognition();

// Runtime/VoxrPushToTalkController.cs:77  (Continuous -> PushToTalk)
_speechRecogniser?.StopRecognition();
```

`_speechRecogniser` is a `VoxrSpeechRecogniser : MonoBehaviour` (`Runtime/VoxrPushToTalkController.cs:18`), i.e. a `UnityEngine.Object`. Unity overloads `==`/`!=` on that type so a *destroyed* object compares equal to `null` while its managed wrapper is still alive. C#'s `?.` performs a plain reference-null check and does **not** call the overload, so for a destroyed recogniser `?.` treats it as non-null and dispatches the call.

Every other site in the same file uses an explicit `== null` / `!= null` check, which does honour the overload — `PressTalk` (`:91`), `ReleaseTalk` (`:102`), `Start` (`:126`), `OnEnable` (`:132`), `OnDisable` (`:138`), `OnApplicationPause` (`:144`), `Update` (`:162`). The two `?.` lines are the odd ones out, so the file is internally inconsistent about the one check that matters here.

*(All line numbers above are against `main` at `7cefd40`. PR #114 rewrites `OnEnable` and shifts everything below it, but leaves `:68`, `:77` and the field declaration at `:18` exactly where they are, so the two offending lines are stable across both.)*

**What actually throws**: the call proceeds into the destroyed component. The first Unity-side access then fails — on Android the permission path calls `StartCoroutine` (`Runtime/VoxrSpeechRecogniser.cs:305`), which throws `MissingReferenceException` on a destroyed MonoBehaviour. On paths that happen not to throw, the behaviour is still wrong: the controller is driving a component Unity considers dead.

**Repro sketch** (not yet written as a test):
1. Put `VoxrSpeechRecogniser` and `VoxrPushToTalkController` on separate GameObjects, assign the reference.
2. `Destroy(recogniserGameObject)` and let a frame pass.
3. `controller.ListeningMode = VoxrListeningMode.Continuous;`
4. Expected: no-op, like every other guarded path. Actual: dispatches into the destroyed recogniser.

**Fix**: replace both `?.` with the explicit `!= null` check the rest of the file already uses. Note the two lines are not symmetric in consequence — `:68` also has `_onTalkStarted?.Invoke()` immediately after it (`:69`), which fires regardless, so decide deliberately whether the event should still fire when there is no live recogniser to start. That question is the same one [#114](https://github.com/jinwoo1601/VoXR-Speech-Recognition/pull/114) settled for the authored-Continuous path, where the announcement is deliberately paired with a `StartRecognition()` that actually happened; the setter currently announces unconditionally, so the two paths disagree.

`?.` elsewhere in the package is fine and should be left alone — `EditorMicBackend` (`Runtime/EditorMicBackend.cs:20`) and `PendingCommandHandler`'s pending command are plain C# types with no overloaded equality.

**Provenance**: surfaced as an out-of-angle note by the `PIT` reviewer on PR #114 and confirmed pre-existing there; deliberately left out of that PR to keep an unrelated fix from riding along in a review-response push.

jinwoo1601:
Closes #115

## What

`VoxrPushToTalkController.ListeningMode`'s setter no longer reaches the recogniser through `?.`. Both transitions now use the explicit `!= null` check every other site in the file already uses:

- `PushToTalk` → `Continuous` — `StartRecognition()`
- `Continuous` → `PushToTalk` — `StopRecognition()`

(The two `?.` sites being replaced are at `:68` and `:77` **of the base commit** `57e9c95`, the coordinates issue #115 uses. Deliberately not restated against head, where they move with every push.)

The issue also asked for a deliberate decision on the events, which fired unconditionally. **Both `_onTalkStarted` and `_onTalkEnded` now fire inside the null guard**, beside the call that earns them. `_wantRecognising` is still updated unconditionally in both branches, so the mode change itself and `PressTalk`'s re-entry guard are unaffected.

Also updated: `CHANGELOG.md` (`[Unreleased]` → `Fixed`) and `Documentation~/push-to-talk.md`, whose *Setter semantics* table stated both events fire without qualification.

## Why

`_speechRecogniser` is a `VoxrSpeechRecogniser : MonoBehaviour`, i.e. a `UnityEngine.Object`. Unity overloads `==`/`!=` so a **destroyed** object compares equal to `null` while its managed wrapper is still alive. C#'s `?.` performs a plain reference-null check and does not call the overload, so for a destroyed recogniser it treated the reference as non-null and dispatched. The first Unity-side access then throws — on Android the permission path calls `StartCoroutine` (`Runtime/VoxrSpeechRecogniser.cs:305`), which throws `MissingReferenceException` on a destroyed MonoBehaviour.

On the event decision, the two precedents in this file agree and I followed them rather than splitting the difference:

- **Start** — #109/PR #114 settled this for the authored-`Continuous` path: `OnEnable` fires `OnTalkStarted` only inside its `_speechRecogniser != null` guard, deliberately pairing the announcement with a `StartRecognition()` that actually happened. The setter was the path that disagreed.
- **Stop** — `ReleaseTalk` already returns at `if (_speechRecogniser == null) return;`, before `_onTalkEnded?.Invoke()`. Gating the setter's stop the same way also keeps start/end **balanced** in the far more common failure case — a reference never assigned at all — where an unconditional `OnTalkEnded` is an end that no start ever paired with. That is exactly the defect `AuthoredContinuous_ComponentDisabled_SwitchToPushToTalkIsSilent` pins.

The remaining unbalanced case is narrow and pre-existing in kind: a recogniser destroyed *mid*-`Continuous` gets no `OnTalkEnded`. `ReleaseTalk` and `OnDisable` already behave that way, so this makes the setter consistent rather than introducing a new shape.

`?.` elsewhere in the package is untouched — `EditorMicBackend` and `PendingCommandHandler`'s pending command are plain C# types with no overloaded equality.

## Validation

- [x] **Unity EditMode — PASS, 133/133** (0 failed, 0 skipped, 0 inconclusive), host project `VoXR TestGround` on Unity 6000.4.7f1 against this working tree.
- [x] **Unity PlayMode — PASS, 548/548** (0 failed, 0 skipped, 0 inconclusive). Both runs verified fresh by their `start-time` stamps, not stale result XML.
- [x] Three regression tests added to `Tests~/Runtime/VoxrPushToTalkControllerTests.cs`, all passing in PlayMode: `ListeningMode_SetToContinuous_RecogniserDestroyed_IsNoOp`, `ListeningMode_SetToPushToTalk_RecogniserDestroyed_IsNoOp`, `ListeningMode_SetToContinuous_RecogniserNull_IsNoOp`. The destroyed-recogniser cases use `Object.DestroyImmediate(_speech)` while keeping the C# reference the controller holds — the state `?.` mishandles.
- [x] All three fail against the pre-fix code by their event-count assertions, since the old setter invoked both events unconditionally. That is reasoning from the diff, **not** a recorded red run — I did not re-run the ~40-minute suite against reverted source.
- [ ] **On-device (Quest) check not required for merge, but not performed.** No native-bridge or mic-capture change here, and the state machine is covered by the PlayMode tests above. The original `MissingReferenceException` symptom was reported on the Android permission path and has not been reproduced on hardware.


jinwoo1601 (commented):
## Independent review — PR #121

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE`, **plus `PIT` added** because the whole diff turns on Unity object-lifetime semantics and on when `UnityEvent`s fire relative to lifecycle hooks. **Head reviewed:** `08a9541`, merge-base `57e9c95`. Four finder agents ran with no context from the session that wrote the code; every numbered finding below was then adversarially verified by a separate agent.

**The runtime change itself verified clean.** Both guards are the genuine overloaded `UnityEngine.Object` form (no `is null`, no `ReferenceEquals`, no cast to `object` — any of which would have re-introduced the bug silently); `_onTalkStarted`/`_onTalkEnded` correctly keep `?.` because `UnityEvent` is a plain C# type; `_listeningMode` and `_wantRecognising` remain unconditional, so no downstream method desynchronises; and the live-recogniser path is identical in call order to the base. Four findings below, one of them a blocker I have already fixed.

---

### 1. CONFIRMED · blocker · the "throws `MissingReferenceException`" root cause was unsupported — **fixed in `88f4d03`**

Three of four angles independently doubted it, and the verifier settled it by compiling a Roslyn repro of the exact call shape. The setter reaches the permission path via `StartRecognition()` → `StartRecognitionCore()` → `_ = StartRecognitionAsync()` (`Runtime/VoxrSpeechRecogniser.cs:284-291`). `StartRecognitionAsync` is an `async Task`, so **every** exception in its body — including the `StartCoroutine` at `:305` — is captured onto the returned Task, which `_ =` discards. Nothing propagates to the setter. The stop direction is stronger still: `StopRecognitionCore` (`:372-414`) touches only managed fields, a static owner compare and P/Invoke — no Unity-side instance member on any platform, so it could never have thrown anywhere.

The shipped `CHANGELOG.md:53` headline therefore asserted a caller-visible outcome the code does not produce, across both transitions. **Note this contradicts issue #115's own root-cause text** — the prescribed fix is right, the stated symptom is not. Corrected in the CHANGELOG and the source comment: the pre-fix failure was *silent*, which is precisely why it survived.

### 2. CONFIRMED · minor · a shipped sample doc still states the events fire unconditionally

`Samples~/PushToTalk/README.md:97-98` — "Switching to `Continuous` fires `OnTalkStarted`; switching away while recognising fires `OnTalkEnded`." True at base, false at head. It is tracked and ships via `package.json`'s samples entry, and this is not inherited drift: PR #114 (`ff36742`) swept this same file when it made the analogous event-pairing change, adding the caveat that still sits directly beneath the now-stale sentence.

Severity is held down by the sample scene wiring a recogniser (`PushToTalk.unity:1193`), so a reader following the sample as shipped sees what it describes — it misleads only a reader who generalises. Secondary, weaker hit: `Documentation~/push-to-talk.md:69` ("That is a *change*, so the setter fires the event"), 18 lines below the new guard paragraph that arguably covers it. **Left unfixed for your ruling** — it is below blocker.

### 3. CONFIRMED · minor · two `Assert.DoesNotThrow` messages claimed a check they cannot perform — **messages fixed, assertion not tightened**

Per finding 1, neither dispatch can throw at this call site. So `Assert.DoesNotThrow(..., "must skip a destroyed recogniser, not dispatch into it")` was inert while advertising the invariant #115 is actually about. The suite's own canon flags exactly this trap: `VoxrPushToTalkPauseTests.cs:112-115` warns of an assertion "unable to see the very throw it exists to rule out".

Scope correction the verifier insisted on: the tests are **not** unpinned. The call and the event sit inside the same guard, so the event-count assertions are a faithful proxy today; escaping needs a contrived guard-splitting edit, not a plain revert. I corrected the messages to claim only what they check. **Not done, left for you:** pinning non-dispatch directly with a call-counting double over `internal virtual StartRecognitionCore` — the verifier confirmed `FakeSpeechRecogniser` (`VoxrPushToTalkPauseTests.cs:32-65`) makes this viable, in a separate fixture since `SetUp` wires `_speech` into `_command`.

### 4. CONFIRMED-AS-CLARITY · nit · PR body line citations are base coordinates in present tense

The body says the setter "**now** uses the explicit `!= null` check" and then cites `:68` / `:77`. Those are base coordinates. At head, `:68` is a comment and `:77` is `StartRecognition()` — the *start* call, under the bullet labelled `StopRecognition()` (actually `:90`). Issue #115 stamped these numbers with an explicit revision disclaimer; the PR body reuses them and drops it. PR-body only — no shipped artifact carries line coordinates.

---

### Refuted

- **The stop-side precedent citation is sound.** An angle claimed `AuthoredContinuous_ComponentDisabled_SwitchToPushToTalkIsSilent` does not pin the defect the body attributes to it (its recogniser is alive; `_wantRecognising` is what short-circuits). The verifier refuted it: the body's antecedent is the *defect* — "an end that no start ever paired with" — and the test's own message pins exactly that ("nothing had announced a start for this to pair with", `:406-407`). The angle also wrongly claimed `ReleaseTalk` is the only site pinning the recogniser-missing case; this PR adds one.

- The destroyed-recogniser preconditions assert only the "reads as null" half via `Assume.That(_speech == null)`; nothing pins the live-wrapper half that makes the state distinct from plain null.
- No test covers `Continuous → PushToTalk` with a **never-assigned** recogniser — the case the PR body calls "the far more common failure case". It is reachable via authored-`Continuous` + null reference.
- The new doc paragraph's "the same way `PressTalk()` and `ReleaseTalk()` do nothing without it" glosses that the setter still flips `_wantRecognising`, which those two do not.
- The tests destroy the component on the shared GameObject rather than issue #115's repro (destroy the recogniser's GameObject, let a frame pass).
- Pre-existing, adjacent: `VoxrCommandRecogniser.OnDisable`'s `if (speechRecogniser == null) return;` also skips `_pending.Cancel()` and the buffer flush — inert here, a trap for a future fixture that destroys the recogniser with a pending command live.

- **On-device (Quest/Android) behaviour.** Finding 1's conclusion is compiler semantics (identical under Mono and IL2CPP), but whether an exception is raised *inside* the discarded Task on real hardware was not observed. The honest pre-fix symptom on device is a silent no-effect plus an unobserved faulted Task.
- **Independent re-run of the Unity suites.** The PR's 133/548 pass counts are an author claim; agents were barred from re-running (it requires renaming `Tests~` and a host project, which would dirty the tree). The claim was audited for internal consistency instead and holds: PR #119 recorded 544, its review commit added one `Tests~/Runtime` test (545 at merge-base), +3 here = 548, with EditMode unchanged at 133.

Nothing here is approved or merged — findings 2 and 4, and the untightened half of 3, are yours to rule on.

jinwoo1601:
Pushed `88f4d03` addressing **finding 1** (CONFIRMED blocker — the unsupported `MissingReferenceException` root-cause claim in the `CHANGELOG.md` headline and the setter comment) and the message half of **finding 3**. Suites re-run after the edit: EditMode 133/133, PlayMode 548/548. Findings 2 and 4, and the untightened assertion in 3, are left for your ruling.

jinwoo1601:
Pushed `a938bf4` — **all four findings are now addressed.**

- **Finding 2** (stale docs) — qualified `Samples~/PushToTalk/README.md:97-98` and the weaker hit at `Documentation~/push-to-talk.md:69`.
- **Finding 3** (inert assertion) — the two destroyed-recogniser tests now pin non-dispatch by **counting calls** through a `CountingSpeechRecogniser` double over the `internal virtual StartRecognitionCore`/`StopRecognitionCore` seam, replacing the `Assert.DoesNotThrow` that could not fail. The preconditions also pin the live-wrapper half of the destroyed state now, so the tests cannot degrade into duplicates of the plain-null case.
- **Finding 4** (line coordinates) — PR body corrected to anchor them to the base commit rather than restating them against a head that moves with every push.
- **Finding 1** was fixed earlier in `88f4d03`.

**The finding-3 fix was mutation-verified, not assumed.** Restoring `_speechRecogniser?.StartRecognition()` outside the guard while leaving the events gated — precisely the edit the old assertions let through — turns both tests red on `StartCalls`/`StopCalls` (`Expected: 0 / But was: 1`), with `ListeningMode_SetToContinuous_RecogniserNull_IsNoOp` still green and no other test in the 548 affected. The mutation was then reverted; the guard is back.

Final state: EditMode **133/133**, PlayMode **548/548**.

Still not verified, unchanged from the review: on-device (Quest/Android) behaviour. The `Notes` section of the review remains open and unactioned — those were explicitly not findings.

PR #122, opened 2026-08-18
Warn when one intent is registered by two definitions (#120)

### LINKED ISSUE #120: Two definitions under one intent resolve differently in the parser and the recogniser, with no warning
**Symptom**: register two `VoxrCommandDefinition`s under the same `Intent` — in one `Configure(slots, commands)` call, or across two command sets made active together — and nothing warns. The second definition is unreachable through the intent lookup, and the parser and the recogniser silently disagree about which of the two a parsed command *means*.

**Root cause**: the two resolutions have opposite tie-breaks, and neither knows the other exists.

```csharp
// Runtime/Commands/CommandSetManager.cs:90-91  — LAST registration wins
for (int i = 0; i < commands.Length; i++)
    _commandLookup[commands[i].Intent] = commands[i];
```

```csharp
// Runtime/Commands/VoxrCommandParser.cs:3684-3693  — FIRST match wins
for (int ci = 0; ci < _commands.Length; ci++)
{
    if (string.Equals(_commands[ci].Intent, intent, StringComparison.Ordinal))
    {
        var patterns = _commands[ci].Patterns;
        if (patternIdx >= 0 && patternIdx < patterns.Length)
            pattern = patterns[patternIdx];
        break;
    }
}
```

`VoxrCommand` carries only `Intent` and `MatchedPatternIndex` — never the command index that actually won the parse — so once a result leaves `ParseInternal` there is no way to recover which definition produced it. Every consumer re-resolves from the intent string, and `CommandSetManager`'s dictionary and `ScoreFollowUp`'s scan then answer differently. `MatchedPatternIndex` is applied to whichever definition each one picked, so it can index a *different* pattern of a different length.

**What it actually breaks**: this is the root cause behind [#113](https://github.com/jinwoo1601/VoXR-Speech-Recognition/issues/113). There, `IsIncomplete` (`Runtime/Commands/VoxrCommandRecogniser.cs:1156-1159`, via `TryLookupCommand` → last-wins) read a short definition and called a follow-up fill complete, while `ScoreFollowUp` (first-wins) re-scored it against a longer one and charged it for required slots the matched pattern never had — delivering a command scoring `-0.1111` through `OnCommandConfirmed`. PR #119 floors the score, which holds regardless of *why* the two disagree, but it deliberately leaves the disagreement itself in place.

The same divergence reaches every other `TryLookupCommand` consumer, each of which may now be reading the other definition than the one that matched:

- `VoxrCommandRecogniser.cs:866` — `AllowPartialMatch` and the pending it enters
- `VoxrCommandRecogniser.cs:979` — `RequiresConfirmation`, i.e. whether a destructive command asks before firing
- `VoxrCommandRecogniser.cs:1069`, `:1120` — the disambiguation winner's and rivals' definitions
- `PendingCommandHandler.ComputeUnfilledSlots` — which slots the follow-up is told to go looking for

`RequiresConfirmation` is the one worth stating plainly: two definitions under one intent, one of them `requiresConfirmation`, and which behaviour you get depends on registration order rather than on the command that matched.

Per-intent debounce (`CommandDebouncer`) is also keyed on intent, so the two definitions share one cooldown — defensible, but it should be a decision rather than an accident.

**Repro sketch** (the #113 fixture is a working one — see `FollowUpFill_ReScoreNonPositive_DoesNotReachAHandler` in `Tests~/Runtime/VoxrPendingCommandTests.cs` on PR #119's branch):
1. `Configure` with two definitions both named `launch_weapon`, the first a 9-element pattern, the last `["launch", "{weapon}", "target", "{target}"]`, both `allowPartialMatch`.
2. `InjectText("launch missiles target")` → pends on `{target}` (resolved through the *last* definition).
3. `InjectText("hotel one")` → the fill re-scores against the *first*. Nothing warned at any point.

**Task**: decide which of these the package wants, then do it —

1. **Warn at construction.** Cheapest and consistent with how every other grammar-authoring hazard is reported here: `WarnOnDroppableRequiredLiteral` / `WarnOnExcessiveOptionalExpansion` / `WarnOnSiblingDiscriminator`, dispatched together at `VoxrCommandParser.cs:655-660` and each `[System.Diagnostics.Conditional("UNITY_EDITOR")]` (`:819`). A `WarnOnDuplicateIntent(commands)` alongside them fits the existing shape exactly. **Note the ordering constraint at `:657-659`** — `VoxrCommandRecogniserInjectionTests` queues *ordered* `LogAssert` expectations across those scans, so a new scan must be appended, not inserted between them.
2. **Or align the two resolutions** so first/last cannot diverge. Changing `BuildLookup` to first-wins is a one-line change but silently alters intent resolution for all five consumers listed above, so it needs its own test pass.
3. **Or make the ambiguity unrepresentable** — carry the winning command index on `VoxrCommand` so consumers resolve the definition that actually matched instead of re-deriving it from a string. Correct, and much the largest.

(1) and (3) are complementary; (2) is a cheaper substitute for (3) and does not remove the authoring hazard, only the disagreement. My suggestion is (1) now — it turns a silent trap into a visible one for the cost of one scan — and (3) only if a real grammar ever needs same-intent definitions to coexist.

Worth confirming while doing this: whether duplicate intents are meant to be *legal at all*. Nothing documents them either way, and if the answer is "no", (1) is the whole fix.

**Provenance**: identified while resolving #113 (PR #119), and confirmed by three independent review angles plus adversarial verification on that PR. Deliberately left out of #119 so a score floor and a grammar-validation change would not ride together.

jinwoo1601:
Closes #120

## What

A fourth construction-time authoring scan in `VoxrCommandParser`, `WarnOnDuplicateIntent(commands)`, appended **after** `WarnOnSiblingDiscriminator()` so the ordered `LogAssert` queues in `VoxrCommandRecogniserInjectionTests` that span the existing scans are unaffected. `[Conditional("UNITY_EDITOR")]` like the two scans beside it.

It reports, once per intent, any intent carried by two or more `VoxrCommandDefinition`s — naming the intent, how many definitions carry it, and the first pattern of each of the two definitions the two resolutions disagree over. Intents are compared ordinally, matching both lookups.

Also:

- `Documentation~/command-recognition.md` — new **Register each intent exactly once** subsection under Authoring hazards, which settles the question the issue left open: duplicate intents are an authoring mistake, and one definition holding both patterns is the remedy. The section intro is updated to cover three scans rather than two.
- `CHANGELOG.md` — entry under `[Unreleased] → Added`.
- Four new tests in `Tests~/Runtime/VoxrCommandParserTests.cs`; the two deliberate duplicate-intent fixtures in `VoxrPendingCommandTests` now `LogAssert.Expect` the warning (matching the `ConfigureAmbiguous` idiom already in that file).

This is option **(1)** from the issue — the author's own recommendation. Options (2) and (3) are deliberately not taken: aligning the two resolutions would settle *which* definition answers without making the second one reachable, and carrying the winning command index on `VoxrCommand` is only worth it if a real grammar ever needs same-intent definitions to coexist. Nothing about parsing, scoring, or firing changes; a player build is untouched.

## Why

`CommandSetManager.BuildLookup` keys a `Dictionary` on intent, so the **last** registration wins there; `VoxrCommandParser.ScoreFollowUp` scans `_commands` and stops on the **first**. `VoxrCommand` carries `Intent` and `MatchedPatternIndex` but never the command index that won the parse, so every consumer re-derives the definition from the intent string and the two can hand back different ones — `MatchedPatternIndex` applied to a pattern of a different length, a different unfilled-slot set for a follow-up to chase, a different `allowPartialMatch`, and a different `requiresConfirmation` (whether a destructive command asks before firing decided by registration order rather than by the command that matched). This is the root cause behind #113, which PR #119 floored without removing the disagreement.

Warning is the cheapest correct answer because the ambiguity buys nothing: two definitions under one intent say nothing that one definition holding both patterns cannot, and that form is unambiguous everywhere.

## Validation

- [x] **Unity 6000.4.7f1 via host project `VoXR TestGround`, both platforms, fresh runs (`start-time` 2026-08-18):** EditMode **133/133**, PlayMode **552/552**, `failed="0"`, `inconclusive="0"`, `skipped="0"`.
- [x] **Mutation-verified, not just green.** With the `WarnOnDuplicateIntent(commands)` call commented out and the four new tests re-run: `DuplicateIntent_TwoDefinitions_WarnsAtConstruction` and `DuplicateIntent_ThreeDefinitions_ReportedOncePerIntent` **fail**, the two `_SaysNothing` tests still pass. The assertions are not inert.
- [x] **Static sweep for grammars this would newly warn on.** A balanced-paren scan of every `new VoxrCommandParser(...)` / `.Configure(...)` / command-array literal in the repo found exactly three duplicate-intent fixtures: the two in `VoxrPendingCommandTests` (now expecting the warning) and `VoxrCommandParserDiagnosticTests.LastParseDiagnostics_SameIntentAcrossTwoCommands_IsNotTheRecordedRival`, which uses no `LogAssert` at all and is left alone. The `Samples~/CommandRecognition` asset grammar has no duplicate intents, so the shipped sample stays silent.
- [x] Debounce claim in the new doc section checked against source: `_debouncer.IsOnCooldown(cmd.Intent, …)` / `RecordFire(cmd.Intent, …)` — keyed on intent, so duplicate definitions do share one cooldown.

No native-bridge or MonoBehaviour-lifecycle code is touched, so no on-device Quest check is implied by this change.


jinwoo1601 (commented):
## Independent review — PR #122

**Depth profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`, plus **`XFILE` added**: this diff's entire deliverable is a message asserting facts implemented in other files, so both ends of those assertions needed tracing.
**Head reviewed:** `ea38c9ed73ef78226e5e57629bc4cb4caf614be4` · **Base:** merge-base `982ab5c882ad3fa4e865ba80b4ba0fea58d46814`

Four finder agents, no context inherited from the session that wrote the code. Five findings survived dedup and the materiality floor; four CONFIRMED, one REFUTED. The scan's *logic* verified clean line-by-line (loop bounds, once-per-intent suppression, `count`, LAST/FIRST index wiring, regex escaping, null/empty `Patterns` paths) — every confirmed finding is about a **statement being false**, not about the code misbehaving.

---

### 1. `Runtime/Commands/VoxrCommandParser.cs` — the warning says "only one of them is reachable", which is false — CONFIRMED, major

Also in `Documentation~/command-recognition.md:445` ("Only one of them is ever reachable.") and `CHANGELOG.md:12`.

Selection iterates every definition and every pattern — `VoxrCommandParser.cs:2257-2260`, with `_commands = commands` at `:432` holding the raw un-deduped array. **Both** duplicates' patterns compete and either can win. Only `BuildLookup`'s dictionary collapses on intent, and the parse never reads it. The runtime has a branch existing solely because both reach selection: `AreSiblingRivals` at `:1937-1943` rejects a rival sharing the winner's intent *across two commands*.

In the doc's own `fire_at` example, "fire at hotel one now" is won by the **first** definition — the one `BuildLookup` discards. An author told it is unreachable draws the wrong conclusion about their own grammar.

What is unreachable is the definition *back from the intent*, not its patterns.

### 2. `CHANGELOG.md:12` — the lead claims the divergence was fixed — CONFIRMED, minor

"**…instead of letting the two resolve differently.**" "X instead of Y" asserts Y stopped. It did not: `CommandSetManager.cs:91` still last-wins, `VoxrCommandParser.cs:3775-3783` still first-wins, and the PR's own source comment concedes it — `VoxrCommandParser.cs:1020` "// Reported rather than repaired."

The file's idiom is consistent the other way: every other "instead of / rather than" lead names a Y that genuinely stopped (`:55`, `:16`, `:83`), while report-only entries use plain "now warns" (`:22`, `:59`). This entry is the only one whose contrasted Y survives. The past-tense body ("was silently accepted", "the last registration **won** there") corroborates the wrong reading, against the present tense the same PR used in `command-recognition.md:459`.

### 3. `KNOWN_LIMITATIONS.md:321-325` — the "four passes" count is now five — CONFIRMED, minor

The entry *"Validation warnings re-emit on every active-set switch"* enumerates the constructor's passes and totals them: "Two further scans are Editor-only … So in the Editor all **four** passes re-run on every switch." Exact at base; the ctor block now runs five (`VoxrCommandParser.cs:654-662`).

Doubly stale, because the new scan is precisely a warning that repeats on every switch: `Activate` concatenates active sets without dedupe (`CommandSetManager.cs:53-79`), so cross-set duplicates re-warn on each `SetActiveSets`. That is the noise this very limitation documents. This file is a shipped product doc, linked from `Documentation~/index.md:35`.

### 4. Identical duplicate registrations produce a false advisory — CONFIRMED, minor

`Activate` validates set-name *existence* only (`CommandSetManager.cs:46-51`) and never de-duplicates, though `Configure` rejects duplicate set names at `:28-29`. So `SetActiveSets("weapons", "weapons")` — or, with no caller error at all, **the same `VoxrCommandAsset` placed in two active set assets** (`VoxrCommandSetAsset.cs:17`, nothing prevents it) — yields structurally identical definitions under one intent.

The message then names the same definition on both sides (`DescribeDefinition` renders both parentheticals byte-identically), asserts the resolutions "disagree about which" when both resolve to equal values, and offers a remedy — merge the patterns — that is inapplicable to a single authored definition. Once per distinct intent, so a 12-command set activated twice logs 12 of them.

- **`Documentation~/inspector-authoring.md` is stale** (its "Three more scan the patterns" list and "The two grammar-shape hazards" count). REFUTED: `:85` scopes its promise to "the shapes below" rather than claiming an inventory; "Three more **scan the patterns**" stays true because the new scan reads the registration list, not patterns; and "two grammar-shape hazards" maps to its own two bullets, which are exactly the two `###` subsections the new one is deliberately not classed with. The page was already a survey-with-deferral at base — the cancel-collision warning (`VoxrCommandParser.cs:1632`) appears in none of its bullets and did not before this PR either.

### Notes

Unverified, one line each — none is a defect claim.

- `command-recognition.md:106` ("Two pattern shapes … **Both** are covered in Authoring hazards") stays literally true but now routes a reader to three subsections.
- `VoxrCommandParser.cs:657`'s pre-existing "Appended last on purpose" comment now sits above a call that is no longer last; the added line below mitigates but does not correct it.
- The message names four affected consumers but omits the two disambiguation ones the issue listed (`VoxrCommandRecogniser.cs:1135`, `:1186`).
- No test covers the cross-command-set duplicate path that the doc and CHANGELOG both advertise; all four new tests construct the parser directly.
- `WarnOnDuplicateIntent` is O(n²) over the command list — construction is not a hot path, and the neighbouring scans are worse.

### Verified clean

Call-site placement after `WarnOnSiblingDiscriminator()` and the ordered-`LogAssert`-queue premise (a repo-wide sweep found no `LogAssert.Expect` sequence spanning a duplicate-intent construction); once-per-intent suppression and `count` hand-walked over `[A,A]`, `[A,A,A]`, `[A,B,A,A]`, `[A,B,C]`, `[A]`, `[]`; LAST/FIRST attribution against `BuildLookup` and `ScoreFollowUp` receiving the same array in the same order; ordinal comparison matching both lookups; the debounce claim; all four consumer claims traced to live line numbers; `default(VoxrCommandDefinition)` unable to reach `DescribeDefinition` (an earlier ctor pass throws first); `[Conditional]`/`DescribeDefinition` interaction in the player and A/B-rig configurations; the PR's static-sweep claim (re-run independently twice — exactly three duplicate-intent fixtures, no duplicates among the 11 sample intent assets); and the claimed test totals, which reconcile from source (EditMode 120 `[Test]` + 13 `[TestCase]` = 133; PlayMode 506 + 21 + 10 − 1 + 16 manifest cases = 552, with the diff adding exactly 4).

### Not verifiable from here

The claimed Unity run itself could not be re-executed by the review agents — this is a bare UPM package with no host project in the repo. The numbers were checked for internal consistency against the diff, not reproduced. On-device (Quest) behaviour is untouched by this diff and was not checked.

jinwoo1601:
Pushed `4df476b` addressing **finding 1** (the false "only one of them is reachable" claim) and **finding 2** (the CHANGELOG lead claiming the divergence was fixed) — both are text this PR itself authored. The warning message, the doc section, its code comment, and the CHANGELOG entry now say that all the definitions' patterns still compete in the parse and any can win, and that what only one definition is reachable for is resolution *back from* the intent. The CHANGELOG lead reads "instead of silently accepting it" and its body is in the present tense, since the divergence survives.

Findings **3** (`KNOWN_LIMITATIONS.md`'s "all four passes") and **4** (identical duplicate registrations producing a false advisory) are left unfixed for your ruling: 3 edits a file this PR does not touch, and 4 needs a code branch plus a decision about what the message should say when the two registrations are structurally identical.

Re-verified after the push: EditMode **133/133**, PlayMode **552/552**, `failed="0"`, `inconclusive="0"` (Unity 6000.4.7f1 via `VoXR TestGround`, both runs fresh). The four new tests and the three `VoxrPendingCommandTests` fixtures still pass — their regexes match the rewritten message unchanged.

jinwoo1601:
Pushed `efbc06e` — the remaining two findings and the notes are now addressed. All four confirmed findings are closed.

**Finding 3 — `KNOWN_LIMITATIONS.md` "all four passes".** Now reads five, with the duplicate-intent scan named among the Editor-only ones. It says explicitly that this scan *compounds* that limitation rather than merely joining it: the duplication it reports can be **created** by the switch, since two sets each carrying a command under one intent are only in conflict while both are active.

**Finding 4 — false advisory on interchangeable registrations.** The scan now picks its report from the pair the two resolutions actually reach:

- **Divergent** (`commands[i]` and `commands[last]` distinguishable) → the existing message.
- **Interchangeable** → a new message: nothing disagrees, but only one registration is reachable from the intent and each extra copy adds a parse candidate that ties the original exactly, leaving registration order to break a tie between a command and itself. It names the two real causes — a set named twice in one `SetActiveSets` call or in `initialActiveSetNames`, and one command asset placed in two sets active together — and drops the "merge the patterns" remedy, which is meaningless when there is only one distinct definition.

`AreInterchangeable` compares a definition's observable surface: patterns element-wise ordinal, plus `allowPartialMatch` and `requiresConfirmation`. Reference equality would not do, since the asset path builds a fresh `string[][]` per definition. Pattern **order** counts as identity, because `MatchedPatternIndex` indexes that array.

Keyed on the reached pair rather than on the whole group deliberately: with `[A, A', A]` both resolutions land on `A`, so nothing disagrees even though `A'` is genuinely unreachable — and the interchangeable report covers that correctly.

I did **not** make `Activate` reject repeated set names. That would be a behaviour change to a public API on a path that works today, and it would not close the shared-asset route anyway.

**Notes also addressed:** the stale "Appended last on purpose" comment now describes the block rather than one call; `command-recognition.md:106` routes to all three hazards; the divergent message now names the disambiguation consumers it had omitted; and the cross-command-set path has tests, which it did not before.

**Validation.** EditMode **133/133**, PlayMode **557/557** (five new tests), `failed="0"`, `inconclusive="0"`, `skipped="0"` — Unity 6000.4.7f1 via `VoXR TestGround`, fresh runs.

Mutation-verified again, since the new branch is the whole of finding 4's fix: forcing `AreInterchangeable` to `false` fails exactly `DuplicateIntent_InterchangeableDefinitions_ReportsCopiesNotADisagreement` and `DuplicateIntent_SameCommandInTwoActiveSets_IsReportedAtRebuild` (555/557), while the three divergence tests keep passing — so the branch is load-bearing and the divergence tests genuinely do not depend on it.

Note on tooling for anyone re-running this: `-testFilter "*DuplicateIntent*"` matches **zero** tests here and Unity still exits 0 on a zero-test run, so the mutation check was done against the full suite.

PR #125, opened 2026-08-27
Bar a candidate whose first required element matched nothing (#124)

### LINKED ISSUE #124: A command whose leading required literal was never spoken can win a later extraction round and fire
## Summary

A pattern that never matched its **leading required literal** can win a later sequential-extraction round and fire, at the default `minScore`. In a command-and-control grammar this means an utterance the speaker intended as one command emits a **second, unspoken command** — in my case a maneuver order that a spaceship then executed.

The two rules that combine are each working as documented; the problem is that clearing the orphan-run terminator test is taken as sufficient to also *fire*.

## Repro

Minimal grammar (two intents, `{track}` is a `NumberSequence`):

```csharp
new VoxrCommandDefinition("query_time_to_target", new[] { new[] { "time", "to", "target" } })
new VoxrCommandDefinition("intercept_target",     new[] { new[] { "intercept", "track", "{track}" } })
```

Defaults: `minScore 0.60`, `minConfidence 0.30`, `coverageWeight 1.0`.

Say: **`time to target track one two four four`**

| round | intent | pattern | score | accepted |
|---|---|---|---|---|
| 1 | `query_time_to_target` | `time to target` | 1.0000 | ✓ |
| 2 | `intercept_target` | `intercept track {track}` | 0.6667 | **✓** |

Round 2 fires `intercept_target` with `track = "one two four four"` — **the word "intercept" was never spoken**, and the per-word table confirms every token was decoded cleanly (confidences 1.0 except one 0.5).

Session-log attempt (schemaVersion 2):

```json
{
  "intent": "intercept_target",
  "pattern": "intercept track {track}",
  "score": 0.6666666865348816,
  "minScore": 0.6000000238418579,
  "aggregateConfidence": 0.5,
  "minConfidence": 0.30000001192092896,
  "accepted": true,
  "rejectReason": "",
  "tiedRival": "",
  "slots": [{ "name": "track", "value": "one two four four" }]
}
```

## Why it happens

Both halves are documented behaviour in `Documentation~/scoring.md`, and they interlock:

1. **Round 1 is charged no coverage.** `orphanedAfter` stops at the first trailing token where some pattern "can begin a match", and that test asks whether any pattern matched from there ends up having matched *strictly more* of its required elements than it missed. From `track`, `intercept track {track}` matches 2 (`track`, `{track}`) and misses 1 (`intercept`). 2 > 1, so `track` terminates the orphan run, `time to target` is charged nothing and scores a clean **1.0**.

2. **The same candidate then wins round 2 and fires.** `2 / 3` = **0.667**, which clears the default `minScore` of 0.60.

So the test that excuses the short pattern from paying for the tail is the test that licenses the tail's candidate to fire. `scoring.md` is explicit that the terminator margin is *"deliberately one notch stronger than admission … Terminating another command's orphan run is a larger claim, because it moves score off a command that is firing, so it takes strictly more evidence for than against."* The gap is that **firing** with the pattern's *first* required element absent is arguably a larger claim still, and currently needs no extra evidence at all.

`scoring.md` already names this residual for the analogous slot case:

> coverage does **not** reach the case where the stranded value's own first word begins some other pattern, because the orphan run terminates there and the bare form is charged nothing — register `["hard", "stop"]` here and the bug returns in full at the default weight.

This is that residual, except the stranded tail matches **another intent's order pattern** rather than a slot value, so the outcome is a phantom command rather than a dropped argument.

## Why it matters

For a dictation-style grammar this is a spurious extra command. For a command-and-control grammar it is a safety failure: the operator asked a **read-only query** and the system executed a **maneuver order** on a target they never named as one. Nothing in the transcript, the readback, or the per-command event indicates a second command was invented — `OnCommandRecognised` fires once per command, so a subscriber cannot even tell the two came from one utterance.

Note also that a 3-element pattern missing exactly one element lands on 0.667, only 0.067 above the default gate — the same narrow margin `scoring.md` already flags as uncomfortable for the `decelerate by {burn_level}` case.

## Possible directions

Not prescribing, but the ones I considered:

- **Treat the pattern's first required element specially.** A pattern that missed it could still terminate an orphan run (the conservative behaviour that protects sequential extraction) but be barred from *firing*. That splits the two uses of the margin, which `scoring.md` already argues are claims of different strength.
- **An authoring warning** for the shape: a bare pattern whose trailing tokens can begin a match on another intent's *non-initial* required literal. This sits naturally beside the two hazards already scanned for at construction.
- **An option to cap commands per utterance** (or disable sequential extraction). Grammars where one utterance is always one command — every legitimate compound being a single pattern — currently have no way to say so, and must re-implement the filter on `OnCommandsRecognised` downstream.

Happy to test a fix against the real grammar that hit this.

## Environment

- VoXR **1.5.0** (`#v1.5.0`), Unity **6000.4.7f1**, Windows, VOSK decoder
- Config as shipped: `minScore 0.60`, `minConfidence 0.30`, `coverageWeight 1.0`, `bufferWindow 2.0`, `prefixHoldSeconds 0.6`, `eagerFlushOnCompleteMatch 1`, `disambiguateSiblingTies 0`
- Reproduced twice in one session on consecutive utterances; a third utterance of the bare form alone (`time to target`) behaves correctly.

jinwoo1601:
Closes #124. Backlog item 1 of the G1-locked `leading-miss-bar` design — **the entire behaviour change**. Items 2–4 (product docs, the construction warning, the 2.0.0 release) are not built here.

## What it does

> **DR-1 — a candidate whose FIRST REQUIRED element matched nothing may compete and consume, but may not fire.**

`time to target track one two four four` fired `query_time_to_target` at 1.0000 **and** `intercept_target` at 0.6667, with "intercept" never spoken. A read-only query executed a manoeuvre order, and nothing marked the second command as invented — every token decoded cleanly, and `OnCommandRecognised` fires once per command, so a subscriber could not tell the two came from one utterance.

The rule is **positional, not statistical**. The first required element is the pattern's anchor — the verb. Later required elements are arguments. Losing an argument leaves the action identified and the score reports the damage; losing the verb leaves no evidence any action was requested. `minScore` sees `2/3` and cannot ask *which* third went missing, which is why this is not a threshold change. **No score moves.** What moves is whether an already-computed score may produce a result.

**Refuse-to-FIRE, not refuse-to-compete (DR-4).** The barred candidate still wins its round and still consumes its span. That is load-bearing rather than incidental: `SkippedBefore` is a bare prefix subtraction with no excusal, so the phantom is a **coverage shield**. Remove it from selection and `target hotel one cease fire` drops `cease_fire` from 1.0000 to 2/(2+3) = 0.4000 — under the gate, firing nothing. On the 699-row corpus that costs 11 clean commands; refuse-to-FIRE costs **0**.

## Three seams

| Seam | Where | What |
|---|---|---|
| **A — the latch** | `MatchResult`, `TryMatchScored` | one `bool`, one local, two latch sites |
| **B — flush suppression** | `ParseInternal` | one `best*` local, one adopt assignment, one guard above every allocation and the tie write (DR-3) |
| **C — eager refusal** | `TryEagerCommit` | explicit condition + the load-bearing comment amendment (DR-5) |

The latch derives "no required element reached yet" from `matchedRequired == 0 && missedRequired == 0` — the ledger that already exists — rather than tracking it separately. Verified: exactly four sites touch those counters and **no optional branch touches either**, so design §9's third trap ("the latch must skip optionals") becomes structurally impossible instead of a convention someone must remember.

## Verification

- **EditMode 133/133, PlayMode 567/567**, both `failed="0"`, freshness verified. Baseline at `b83b62f` was 133 / 557.
- **Three mutation checks**, run rather than asserted:
  - delete the eager refusal → fails the eager test and **nothing else** in either suite;
  - convert the bar to refuse-to-compete → fails the leading-debris test at exactly **0.4000**, and only its *score* assertion catches it (count and intent still pass);
  - allocate inside the latch → fails the alloc pin **on the `AllocatingGCMemory` constraint**, past its warm-up, so the pin is not blind to this environment's documented `GetAllocatedBytesForCurrentThread` false-green.
- **Rig A/B over 699 utterances.** The shipped build reproduces the pinned `corpus-bar-on.tsv` **byte-identically**, and `b83b62f` reproduces `corpus-bar-off.tsv` byte-identically. 96 rows differ between arms and **all 96 are pure result-count drops** — nothing re-ranked or re-scored. (The 48/17 in the CHANGELOG is the same diff gated at `minScore` 0.60.)
- **G-5:** allocation is *exactly* unchanged where the bar is inert (495.5 → 495.5 B/utterance across 603 rows) and drops 112.6 B/utterance where it fires.
- **Deferred, human-only:** on-device Quest verification. No native-bridge or lifecycle code is touched, so no ABI rebuild is owed and no `.so` is committed — reported as deferred, not claimed.

## Please look at these — four things I want ruled, not just merged

**1. A canon erratum, and it is in a load-bearing comment.** Design §2.4/§5.2 asserts *"an eager commit IS a fire"*, and my first draft repeated it in the exact comment DR-5 requires — the one standing against a future "this is redundant" deletion. **It is false.** `Commit` makes the recogniser call `FlushBuffer`, which routes through `ProcessParsedResults` into `ParseInternal`, the very method the bar lives in. A reader who follows `Commit` two hops disproves the note and is then licensed to distrust all of it. The condition is still load-bearing for a narrower reason, now recorded: deleting it costs **the buffer** — the transcript is consumed at partial-poll time, so a continuation that would have completed a genuine command arrives to an emptied buffer. **DR-5's ruling is unaffected; only its stated rationale was wrong.** Recorded as an erratum, no design branch — same disposal as item 2 below.

**2. Canon §6.2 carries a separate factual error; DR-2 stands.** §6.2 argues the literal-vs-slot fork *"only bites for a slot-leading pattern with … five or more"* required elements. That is false: the completeness gate at `VoxrCommandRecogniser.cs:929` gets there before length or score matter, so the fork does not "only bite at five or more". **Corrected after review:** my first statement of this said such a pattern "can never fire today at any length", which overstates it — `:929` is followed by an escape hatch, and a command with `AllowPartialMatch` is routed to `EnterPending` rather than rejected, then fires once the missing argument is supplied. The erratum holds unconditionally only where `AllowPartialMatch` is off. §6.2 is still wrong and DR-2 (ruled on principle, explicitly *"not on numbers"*) is unaffected — but since this annotates an immutable document, the supporting fact now carries its condition. T7 is reframed to pin what is observable: element-type uniformity in `ParseInternal`.

**3. A third fire path the design does not discuss.** The bar is applied to the round **winner**. A *tied rival* that is itself leading-missed is still recorded and, with `disambiguateSiblingTies` on, can still be offered as a disambiguation choice and fire if the speaker picks it. Reachable with a sibling set whose discriminator is at element 0. Harm is bounded — the speaker has to explicitly utter the discriminator to select it — which is why I did not expand scope to "fix" it here: extending DR-1 from the winner to rivals is a design question, not an implementation detail. **Flagging for your call.**

**4. Accepted consequences — four, not the two I first listed.** Three are reporting-layer; the fourth is not, and it is the one worth your attention.

- A barred round records **no** Editor parse-diagnostic entry (forced by DR-3's placement — keeping it demands the placement DR-3 rejects).
- A barred utterance's `rejectReason` reads `no match` rather than a score rejection.
- The **batch test runner** reports it as *"no pattern matched"* — but a pattern *did* match, won its round, consumed its span, and was refused by position, so a grammar author certifying that utterance is told the wrong thing.
- **Not reporting-layer, and not inert for your configuration:** the eager refusal sits above both `HoldExtendable` returns, so a barred winner on an extendable pattern converts `HoldExtendable → None` and waits the full `bufferWindow` instead of `prefixHoldSeconds`. Inert at the shipped default (`prefixHoldSeconds = 0`), but issue #124 reports `bufferWindow 2.0, prefixHoldSeconds 0.6` — so on the reporter's own setup a barred utterance is reported unrecognised ≈1.4 s later than before. Nothing fires either way; it is reported later. DR-5 rules the gate's position beside #66/#70, so moving it below the hold returns would be a design change, not a fix.

The first three are requirements §8.4's family — leaning accept, with the cheap remedy (an allocation-free barred-round counter) belonging to item 3. The fourth is a latency change and is yours to weigh.

**5. Two recogniser-path behaviour changes, now enumerated in the CHANGELOG** — found by review rather than predicted by the design. A barred utterance no longer **cancels a live pending** or preempts a follow-up slot-fill: the phantom used to count as a complete new command and clear whatever question was outstanding. And threshold-filtered barred candidates move from *silent* to raising `OnUnrecognisedSpeech`.

Also worth knowing, none of them blocking:

- **A measured +3.2% wall-clock cost** (+0.628 µs/utterance) — a uniform per-parse constant tax from the two latch comparisons, *larger* on utterances where the bar never fires. Predicted by architecture §2.1; note F15's "two stack locals" understates it slightly, since a per-candidate struct field also landed. The field was moved into `MatchResult`'s existing padding hole so the struct stays 32 bytes rather than growing to 36.
- **A barred round does not consume the result-buffer round cap**, so the worst-case round count per parse is now bounded by token count rather than by `commands.Length`. No correctness impact; the loop still terminates and buffer indexing is unaffected.
- **Open question §9.1** — whether T7's slot-leading fixture should join the corpus. Leaning **no**: it would invalidate the byte-identical baselines every prior arm rests on.
- **Flagged for item 2, deliberately not fixed here:** `Documentation~/scoring.md` §1 publishes the `cease fire` / "fire" → 0.50 row. The *number* survives (re-pinned on a trailing miss) but its *mechanism* does not, and the companion test pins a different input — so that published row ends up pinned by nothing. DR-9 scopes item 2 to §7 D only, so §1 is unowned.
- **The `CHANGELOG.md` entry lands on this branch**, over-riding canon §9's assignment to item 2. Ruled a scheduling call: §9 is a work partition, not a decision record, and no DR pins the changelog's branch. Item 2 extends the entry rather than originating it.

## Review

Two independent passes, both folded in: an 11-angle `review-cycle` on the branch (commit 2) and a 12-angle `review-pr` on this diff at full profile (commit 3, verdict posted as a review on this PR).

Neither found a blocker in the code — the rule, its placement, the byte-identical G-2 scope fence and the corpus reproduction all held under adversarial re-derivation, and `CompareCandidate` was verified byte-identical too. What they found was in the prose, and the two worst were mine: the load-bearing comment DR-5 requires justified itself first with a mechanism this codebase does not have, then with an invariant the cooldown already violates. Both are corrected, the second stated narrowly on purpose. Items 2, 4 and 5 above are corrections that changed what this PR asks you to rule on.

Known and carried, not fixed: the round-cap bypass, the tied-rival path in item 3, and four coverage gaps the review named — the eager condition is pinned only as a pure function (the recogniser-level half that #70 carries), the adopt-site reset is unexercised on the flush path, DR-2's slot half is unpinned on the pending path, and the `requiresConfirmation` / `disambiguateSiblingTies` CHANGELOG claims have no test. Each is a test to add; none changes behaviour.

https://claude.ai/code/session_01Umxc3U6642ftnjSjMLh9pE


jinwoo1601 (commented):
## Independent review — `review-pr`, **full** profile

**Head reviewed:** `92ae39abf9b8518dc975f7b039d3378209cde5e5` · **Base (merge-base):** `b83b62f` · **Angles run:** CLAIM + all 11 of the `review-cycle` roster (LINE, GONE, XFILE, PIT, WRAP, REUSE, SIMP, EFF, ALT, CONV, GAP). None dropped, none added.

**The mechanism is sound.** LINE re-derived every changed statement and hand-scored all nine new fixtures against the real scoring constants and found nothing; WRAP confirmed the two selection scans cannot disagree; XFILE confirmed the `ResultBuffer`/`TiedSiblingBuffer` lockstep survives the early `continue` across every consumer; SIMP found nothing removable. The G-2 scope fence was checked mechanically rather than trusted — `BuildCoverageTables`, `SkippedBefore`, `OrphanedAfter`, `IsAdmissibleStart` **and** `CompareCandidate` extracted from both revisions and compared **byte-identical**. Independent verification of the rig pins: `corpus-bar-off` vs `corpus-bar-on` differ on exactly **96 of 699** rows, every one a strict count drop with byte-identical surviving scores and slots; gated at 0.60 they reproduce the CHANGELOG's **48 / 17 / zero clean 1.00 lost** exactly; and refuse-to-compete loses **11** commands that score 1.0000, reproducing DR-4's figure.

**What the review found is concentrated in the prose, not the code** — and the most serious item is a claim in the PR body itself.

---

### 1. The §6.2 erratum's supporting fact is overstated, and the PR asks you to strike canon on it — **major**, CONFIRMED

The PR body says: *"§6.2 … is false: `VoxrCommandRecogniser.cs:929` rejects any winner carrying an unfilled required slot **regardless of score**, so such a pattern can never fire today at any length."*

`:929` does not do that. It reads `if (cmd.Score < minScore || incomplete)` and the very next block is an escape hatch (`:931-948`): a command with `AllowPartialMatch` and unfilled slots is **routed to `EnterPending`**, not rejected — and a completed pending fires through `InterpretResolution` → `OnCommandRecognised`. This PR's own new test says so in as many words (`VoxrPendingCommandTests.cs:208-211`).

So the erratum holds only where `AllowPartialMatch` is off. *"Can never fire today at any length"* is false as written. **DR-2's ruling is unaffected either way** — §6.2 records it as ruled "on principle, not on numbers" — but the erratum needs restating with its condition attached, or withdrawing, before it is used to annotate an immutable document.

### 2. The replacement rationale for the eager condition is falsifiable in one hop — **minor**, CONFIRMED

`VoxrCommandParser.cs:4162` now argues: *"the eager gate exists to speculate about what the flush will do, so it **must mirror every flush-side refusal** or it discards a buffer on a verdict the flush would not have reached."*

The recogniser refuses to fire a command inside its cooldown (`VoxrCommandRecogniser.cs:979-982`), and `ProbeEagerCommit` calls `TryEagerCommit(tokens, wordConfidence, minScore, minConfidence)` (`:639`) — **no cooldown input at all**. A debounced command therefore commits eagerly, the buffer is consumed, and nothing fires: precisely the harm the sentence claims the invariant prevents.

This matters more than its severity suggests. That comment is what F7 and architecture §8 designate as *the only thing standing against* a future "this is redundant" deletion — and this PR rewrote it specifically because the previous rationale was refutable. Replacing one overstatement with another reproduces the failure mode the rewrite was for. Its own words: *"the overstated version invites its own refutation."* The narrow true version ("the eager gate must not commit on a verdict the parser's flush would refuse") is enough.

### 3. A third accepted consequence is in the code but not in the PR body — and it is not inert for the reporter — **minor**, CONFIRMED

The body lists *"two accepted consequences, both reporting-layer."* The diff marks a third at `VoxrCommandParser.cs:4233`: the bar sits above both `HoldExtendable` returns, so a barred winner on an extendable or un-analysable pattern converts `HoldExtendable → None`, lengthening the wait from `prefixHoldSeconds` to the full `bufferWindow`. It is **not** reporting-layer.

The comment mitigates with *"inert at the shipped default, where `prefixHoldSeconds` is 0"* — true of the field default. But issue #124's reporter states their configuration as `bufferWindow 2.0, prefixHoldSeconds 0.6`, so it is live for the exact environment this fix targets: ≈1.4 s later "didn't catch that". Worth listing as a third item for the same ruling, with that caveat attached.

### 4. Erratum 1 was recorded in two places and the claim lives in four — **minor**

The eager erratum is scoped to "design §2.4/§5.2". The falsified claim also sits in **DR-5's own justification cell** (`design-docs/leading-miss-bar.md:419`, *"Under A the eager path would bypass the bar entirely and defeat G-1"*), in requirements §4.3 (`:142`), and in architecture §4.1/§8 (`:189`, `:360` — *"the one defect that would ship green"*). Compare erratum 2, which architecture §10 recorded with named actions and doc locations. The asymmetry matters because the erratum downgrades this seam from "the G-1 hole" to "a buffer-preservation gate", and the next reader of DR-5 is fork C, which canon says may collapse the two sites back into one.

### 5. Two behaviour changes on the recogniser path that nothing enumerates — **minor**, CONFIRMED

Both follow from a barred utterance producing zero results, and neither is in the CHANGELOG's otherwise careful enumeration:

- **A barred utterance no longer cancels a live pending, and no longer preempts a follow-up slot-fill.** `hasCompleteNewCommand` (`VoxrCommandRecogniser.cs:763`) gates both `:772` (follow-up preemption) and `:883` (`if (_pending.HasPending && hasCompleteNewCommand) InterpretResolution(_pending.Cancel())`). The reported phantom — complete, 0.667, slot filled — set it and cancelled a live pending. It no longer can. `LeadingRequiredMiss_OpensNoPending` covers "opens no pending", never "does not cancel one".
- **Threshold-filtered barred candidates go from silent to an event.** Step 7 drops confidence- and debounce-filtered results silently (`:1078`, `if (!anyThresholdFiltered)`); those utterances now take Step 6's unconditional `OnUnrecognisedSpeech`. The CHANGELOG's "arrive as `OnUnrecognisedSpeech`, exactly as any other unmatched speech does" is true of the new state but hides that a subset was previously silent.

### 6. Cost: the round cap is bypassed, and the amplification is worst on small grammars — **minor**

`_resultBuf = new VoxrCommandResult[Math.Max(commands.Length, 1)]` (`:602`) and the top-of-loop `if (_resultCount >= _resultBuf.Length) break;` capped rounds at `commands.Length`, because every surviving round emitted. A barred round is free, so the bound is now `tokens.Length` — worst case moves from `O(min(T,C) · C·P·T·L)` to `O(T² · C·P·L)`, each round a full `(command × pattern × startIdx)` scan. Largest amplification on the *smallest* grammar: a one-command `intercept track {track}` against "track one track two track three…" ran exactly one round before and now runs one full scan per barred segment. Claim 13 states the new bound but frames it as "No correctness impact" — true, and incomplete: it is a cost claim, and it plausibly contributes to the measured +3.2%.

### 7. Coverage gaps — **minor**

- **The eager condition is pinned only as a pure function** (`VoxrEagerCommitTests.cs:923`). Its entire stated purpose is a recogniser-level consequence (the buffer), and nothing drives `VoxrCommandRecogniser` with `EagerFlushOnCompleteMatch = true` and a non-zero `BufferWindow` on a leading-missed shape — the new recogniser test sets `BufferWindow = 0f`, bypassing the eager path. **The codebase already learned this exact lesson**: `VoxrCommandRecogniserInjectionTests.cs:1288` records that "every eager test called `TryEagerCommit` as a pure function and so structurally could not observe what happened after a refusal", and the sibling #70 gate carries both halves. DR-5's gate got only one.
- **The adopt-site reset is unexercised on the flush path.** `bestLeadingRequiredMissed = matchResult.LeadingRequiredMissed;` (`:2382`) must be an unconditional reset so a barred incumbent beaten by a clean rival un-bars the round. In every new parser test the barred candidate is the only candidate or is adopted last, so a sticky `|=` variant passes all of them. The one fixture with the right shape (`VoxrCommandParserTests.cs:5120`) is routed only through `TryEagerCommit`.
- **DR-2's slot half is pinned by nothing on the pending path.** Architecture §10 ruled that the one behaviour uniformity genuinely changes is on the pending path, "which is T9's territory" — but T9's fixture (`VoxrPendingCommandTests.cs:198`) uses a **literal** anchor. T7's comment still opens "this test is its ONLY evidence" without the firing-sense reframing §10 ruled.
- **Two CHANGELOG claims are unpinned:** `requiresConfirmation` and `disambiguateSiblingTies`. No existing disambiguation fixture can reach the bar — `ConfigureAsking` uses an interior discriminator.
- **DR-3's placement is review-enforced only.** Every position between the progress guard and the emit is behaviourally identical, and the allocation half is unmeasured on the `Parse` side. Already acknowledged in architecture §8; restating so it is visible at the gate.

### 8. A third reporting surface — **minor**

`Runtime/Testing/VoxrBatchTestRunner.cs:121` reports a barred utterance as *"expected intent '…' but no pattern matched"*. For #124's own shape a pattern **did** match — it matched, won its round, consumed its span, and was refused by position. A grammar author certifying `"track one two four four"` is told their pattern is unreachable rather than barred. Not a regression (the file is untouched and the count contract holds); a message written for one condition now covers two.

---

### Nits

- `VoxrCommandParser.cs:3062` — a stray blank line before `MatchResult`'s closing brace. **Flagged independently by six angles**, and CONV measured it as the *only* one of 753 added lines that CSharpier would rewrite.
- `VoxrCommandParserTests.cs:4859` — `feat-leading-miss-warning` is the only branch name anywhere in tracked source; house idiom is issue-plus-item. (The `Runtime/` instance was already fixed; this one was missed.)
- `VoxrCommandParserTests.cs:2590` — "maneuver": the sole US spelling in the repository, in a codebase whose public surface is `OnCommandRecognised`. Its two siblings in this same PR use "manoeuvre".
- `VoxrCommandParser.cs:1654` — "the single set that survives to warn, `mode_weapons@3`, has an **interior** discriminator". It is index 2 of a 3-element pattern — trailing, not interior. The operative half (the bar never touches it) is true either way.
- `VoxrCommandParser.cs:3430` — the object initializer appends the new member last while the field is declared third in the bool run; the struct's other nine members mirror declaration order exactly. Pointed, given the field comment argues at length that its position is load-bearing.
- `VoxrCommandParserTests.cs:5095` — a *new* comment repeats "two stack locals", the wording the PR body itself flags as understating what landed.
- `VoxrCommandParser.cs:3032` — the layout argument holds only because every field is a primitive (managed-sequential). Add one reference-type field and CoreCLR/Mono drop to auto layout and the placement stops meaning anything. There is no `sizeof` pin anywhere to catch it; worth stating the precondition.
- `VoxrEagerCommitTests.cs:1587` — `TryEagerCommit_LeadingTwoElementTie_RefusesForScoreAtDefaultAndForTheTieBelowIt` keeps a name its own added comment concedes is now false ("that shape can never reach the sibling condition again"), while its two peers were renamed for exactly this reason.
- `VoxrCommandParserTests.cs:2548` duplicates the preceding test's grammar block byte-for-byte; the file's house pattern is a region-local static builder (`MissedLiteralParser()`, `SetOnSiblings()`), which still satisfies architecture §10's "explicit local grammar" rule.
- `LeadingMissLatch_AllocatesNothingPerCall` duplicates the exemplar's measurement block verbatim; a shared helper would keep each test's fixture and warm-up while removing the copy.
- A barred round still pays `Array.Copy` into `_bestSlotBuf` (and the two Editor slot-index buffers) plus the full `RecordTiedSiblingRival` path, all of it read only inside the block the `continue` skips. Safe to gate on the flag — but the saving lands only where the bar fires, which is already net cheaper, so it does not recover the +3.2%.
- Two new parser tests omit the `LogAssert.NoUnexpectedReceived()` their five siblings carry.
- Two new tests land under section banners that no longer describe them (`#65 §5.1` and `#74 F13`, the latter's prose saying "two claims" above three tests).
- `Documentation~/scoring.md` §7 D still publishes `approach_target` at 0.67 / `"accepted": true` — a second stale worked example beyond the §1 row the PR discloses. Out of scope (item 2), flagged so it is not lost.
- `requirements.md:262` still states "The parse diagnostic keeps recording the barred round", contradicted by its own §3 and §8.4 and by the code; `requirements.md:132-134` still calls the tie-buffer argument "the only one that produces wrong output", uncorrected after architecture §8's `CORRECTED 2026-08-27` retracted it.

### Considered and refuted

- *"The eager refusal is dead code / redundant with the flush suppression."* Refuted two independent ways: the `bestMissedRequiredSlot` gate above narrows it to required-**literal** anchors rather than subsuming it, and the whole-buffer gate below passes on the shape in question (a leading literal miss consumes no token, so `bestEndIdx == tokens.Length` still holds).
- *"The `missedRequired == 0` conjunct should be dropped as redundant."* It is redundant, and the prior round's deliberate keep is right: removing it makes the latch's correctness depend on idempotence rather than on the guard reading as "no required element reached yet".
- *"`MatchResult` grew."* Measured: base 32 bytes, shipped 32 bytes, appended-after-`MissedRequired` 36. Claim 12 holds.
- *"The tie-buffer skip could leave a stale record readable."* Refuted: index `i ≥ _resultCount` is never read, and any later emitting round rewrites the slot before incrementing.
- *"`CHANGELOG`'s 48 rows contradicts the PR's 96."* Both correct — 96 raw, 48 gated at `minScore` 0.60. Reproduced from the pinned TSVs.

### Not verifiable from here

Unity EditMode/PlayMode runs (they need the host project and a Unity install), the three mutation runs, and the A/B rig re-execution — all were checked for *consistency* with the diff rather than re-run. On-device Quest verification is human-only and is correctly reported as deferred, not claimed.

---

**No blocker.** Nothing found alters what the parser does; the rule, its placement, the scope fence and the corpus reproduction all hold under adversarial re-derivation. Items 1–3 are corrections to *statements* — one in the PR body you are being asked to rule on, one in the comment F7 designates as load-bearing, one a disclosure gap — and items 5–7 are behaviour and coverage the gate should see before it closes. Ruling is yours.

🤖 `review-pr` (full profile, 12 angles) · [Claude Code](https://claude.com/claude-code)

jinwoo1601:
Pushed `a9bbb37` addressing the review's statement-level findings (items 1–4 and the nits). Behaviour is unchanged — the only non-comment edits are one removed blank line, an object-initializer line moved to mirror declaration order, and one test renamed.

- **Finding 1** — the §6.2 erratum now carries its `AllowPartialMatch` condition, in the PR body and in architecture §10's erratum record. §6.2 is still false; "can never fire at any length" was not.
- **Finding 2** — the eager comment's "must mirror every flush-side refusal" is replaced with the narrow true claim, and now says explicitly why the general form is wrong (the cooldown is a flush-side refusal this scan is not handed).
- **Finding 3** — the `HoldExtendable → None` latency is listed as a fourth accepted consequence in the PR body and in the CHANGELOG, with the note that it is *not* inert at the reporter's `prefixHoldSeconds 0.6`.
- **Finding 5** — the pending-cancel/follow-up-preemption change and the batch-runner reporting surface are now enumerated in the CHANGELOG.
- **Nits** — stray blank line, last branch name in tracked source, last US spelling, "interior" → trailing, initializer order, the struct-layout precondition, and `…AndForTheTieBelowIt` → `…AndForTheBarBelowIt` (its own comment had conceded the old name was false).

**Deliberately not fixed, carried for the ruling:** finding 4 (erratum 1's scope — the claim also sits in DR-5's justification cell in the locked design, which I will not edit), finding 6 (the round-cap bypass), and finding 7's four coverage gaps. Each is a test to add or a canon edit to authorise, not a code change.

EditMode 133/133, PlayMode 567/567 after the push.

jinwoo1601:
All four G2 rulings applied. Pushed `eab7eb6`; branch is now three feature commits plus two review commits.

**1 — Erratum E-1 widened, no re-lock.** The false "an eager commit IS a fire" premise is now annotated at every one of its nine locations, including **DR-5's own justification cell**, which the first erratum missed. Ruling text untouched; the design doc gained only pointer annotations, and architecture §10 carries the full record with a location table. Fork C is named as the reader it exists for. `Planning~/` is gitignored, so none of this touches the diff.

**2 — Tied-rival fire path → #126.** Not fixed here: extending DR-1 from the round winner to tied rivals is a scope question canon did not rule, so it goes where it can be ruled on its own evidence.

**3 — Hold latency → #127.** Accepted for this PR with DR-5's position kept; filed so it is ruled alongside item 3, which owns the author-facing half and could pair the shortened wait with a diagnostic. The issue records that it is *not* inert at the reporter's own `prefixHoldSeconds 0.6`.

**4 — Two of four coverage gaps closed here; two filed as #128.**
- **Closed:** the eager condition now has its recogniser-level half (`EagerFlush_LeadingRequiredMiss_KeepsTheBufferForTheRestOfTheUtterance`) — the half this suite's own section note records issue #74 item 2 losing for exactly this reason, and the half #70's gate carries. And the adopt-site reset is pinned (`LeadingMiss_ABarredIncumbentBeatenByACleanRival_UnbarsTheRound`), where a sticky `|=` tidy-up would previously have passed every test while barring rounds a clean command won.
- **Filed:** DR-2's slot half on the pending path, and the two unpinned CHANGELOG claims.

Worth recording: the first of those two tests **failed on its first run, and the failure was mine, not the code's** — the control ran without draining the buffer, so all four injections accumulated into one utterance. Both behavioural assertions had passed. The drain is now part of what the test proves, and the flushed text is asserted to be the *merged* string, which is what demonstrates the buffer survived rather than being silently dropped.

**EditMode 133/133, PlayMode 569/569.** No runtime behaviour changed after the review — the only non-comment edits across both review commits were one removed blank line, an object-initializer line reordered, and one test renamed.

Issue #130, opened 2026-08-27
Tracking: leading-miss-bar backlog items 2-4 (#124 closed with item 1)


jinwoo1601:
#124 auto-closed when #125 merged (the PR carried `Closes #124`). That was correct for item 1 — the behaviour change is done and the reported bug is fixed — but the design behind #124 has a **four-item backlog**, and items 2–4 are not built.

That backlog is canonical and locked at G1, but it lives in `Planning~/design-docs/leading-miss-bar.md` §9 — which is **gitignored and local to one machine**. With #124 closed there is now nothing on GitHub recording that three items remain. This issue is that record.

## Backlog state

| # | item | DRs | state |
|---|---|---|---|
| 1 | **`feat-leading-miss-bar`** — the whole behaviour change | DR-1/2/3/5/6/8 | ✅ **merged** (#125, `41ea036`) |
| 2 | **`feat-leading-miss-docs`** — product docs | DR-9 | ✅ **merged** (#131, `747c8df`); closed #129, whose audit found ~40 stale sites across 8 files |
| 3 | **`feat-leading-miss-warning`** — construction-time author-facing warning | — | ✅ **merged** (#134, `1c892eb`); closed #128. Deliverable A only — B was **cut at measurement**, evidence refiled as #135 |
| 4 | **`chore/release-2.0.0`** — the version bump | DR-10 | ⬜ **the only item left**; via the `release` skill. #129 no longer gates it |

Dependency-ordered. Each is one feature branch under the implementation track, picked up only when actually going to be built.

## Ordering constraint worth stating explicitly

**Item 2 must land before item 4.** Right now `main` carries the new behaviour with unmodified docs — but no tag captures that mismatch, so nobody installing a released version is affected yet. Cutting 2.0.0 before the docs are fixed would stamp a release in which `troubleshooting.md` gives advice that cannot work (see #129, priority 1).

Requirements §1 says that if item 1 ships alone "the package is correct — it is merely under-documented." Do not lean on that sentence when sequencing: #129 establishes the docs are *mis*-documented, including reversed guidance in three places.

## Item 3 has grown a dependency it did not have at G1

Canon rules item 3 independently droppable. Two things filed since argue it is no longer free to drop:

[...]

Issue #129, opened 2026-08-27
feat-leading-miss-docs: DR-9's scope is ~10x too small — full audit of what #125 made false


jinwoo1601:
The leading-required-miss bar shipped in #125 (merged, `41ea036`). **No file under `Documentation~/` was touched** — `git diff --name-only b83b62f..41ea036 -- 'Documentation~/' KNOWN_LIMITATIONS.md README.md` is empty. That was the ruled plan: DR-9 assigns the product docs to backlog item 2 (`feat-leading-miss-docs`).

This issue records what a full audit of the shipped docs against the merged behaviour actually found, because **DR-9's scope is roughly an order of magnitude too small.** DR-9 names `scoring.md` §7 D. The real surface is ~40 statements across 8 files, and it includes advice that is now actively wrong rather than merely stale.

> Requirements §1 says that if item 1 ships alone, "the package is correct — it is merely under-documented." That is the sentence to distrust when planning this. It is mis-documented, not under-documented.

## Blocking relationship

**Item 2 must land before item 4 (the 2.0.0 release).** Today `main` carries new code with old docs, but no tag captures the mismatch. Cutting 2.0.0 first would stamp a release in which `troubleshooting.md` tells users to lengthen a pattern to fix something lengthening cannot fix.

---

## Priority 1 — advice that is now wrong, not just stale

A reader who follows these will do the wrong thing.

| site | quoted | why it is wrong now |
|---|---|---|
| `troubleshooting.md:81` | *"For the two-element case, lengthen the pattern or accept the ambiguity."* | Lengthening does not help a **leading** miss. `cease fire now` heard as "fire now" scores 0.67 and is still barred. For a leading miss neither knob nor length recovers it — the speaker re-utters. |
| `command-recognition.md:439` | *"**Making the difference come earlier in the pattern does not help.** … `weapons mode active` and `navigation mode active` tie at 0.67 and fire the first-registered again."* | **Reversed.** "mode active" now misses element 0 of both, so both are barred and nothing fires. Moving a discriminator to the first required element is now an effective mitigation at any pattern length. Belongs out of "what does not work". |
| `KNOWN_LIMITATIONS.md:603-607` | *"**What does not work**: moving the difference *earlier* … Grow the pattern back (`weapons mode active`) and the wrong-command behaviour returns at 0.67."* | Same reversal. It does not return; both candidates are barred at any length. |
[...]

Issue #128, opened 2026-08-27
Two leading-miss-bar guarantees are unpinned: DR-2's slot half on the pending path, and two CHANGELOG claims


jinwoo1601:
Two claims made by #125 (the leading-required-miss bar) that no test pins. Neither is a defect — the behaviour is correct in both cases — but both are documented guarantees with nothing guarding them, and both were ruled at G2 as ride-along work rather than merge blockers.

The other two gaps that review found (the eager condition's recogniser-level half, and the adopt-site reset) **were** closed in #125 at `eab7eb6`.

## 1. DR-2's slot half is unpinned on the pending path

Architecture §10 reframed T7 after the §6.2 erratum, and ruled:

> It cannot be "the only evidence for DR-2" in the *firing* sense … The one behaviour uniformity genuinely changes is on the **pending path**, which is T9's territory.

T9 shipped as `LeadingRequiredMiss_OpensNoPending` (`Tests~/Runtime/VoxrPendingCommandTests.cs`), built on:

```
launch_weapon: ["launch", "{weapon}", "target", "{target}"]   // allowPartialMatch
```

That is a **literal** anchor. So the behaviour §10 assigned to T9 — a *slot*-anchored, leading-missed `allowPartialMatch` candidate no longer opening a slot-fill prompt — is pinned by nothing, and DR-2 (the one decision record ruled with **zero** corpus measurement behind it, explicitly "on principle, not on numbers") is evidenced only at `Parse` level, where the erratum says the difference is not about firing.

**Fix:** a slot-anchored variant of T9 — a pattern whose first required element is a slot, on an `allowPartialMatch` command, asserting `HasPendingCommand == false` when that slot matches nothing.

## 2. Two CHANGELOG claims have no test

The `[Unreleased]` entry for #124 enumerates the downstream subtractions. Three are named; only the first is pinned.

| claim | pinned by |
|---|---|
| `allowPartialMatch` no longer opens a slot-fill prompt | `LeadingRequiredMiss_OpensNoPending` |
| `requiresConfirmation` no longer asks you to confirm an action nobody requested | **nothing** |
| `disambiguateSiblingTies` no longer raises "which did you mean?" about two invented alternatives | **nothing** |

Both are true — they follow from a barred round producing no result, which is the same mechanism T9 pins — but they are published statements in a changelog with no guard.

[...]

Issue #127, opened 2026-08-27
Leading-miss bar suppresses the shortened eager hold, lengthening the wait before OnUnrecognisedSpeech


jinwoo1601:
Accepted for #125 and filed here so it is ruled alongside `feat-leading-miss-warning` (item 3 of #124), which owns the author-facing half of that design and is the natural place to pair a shortened wait with a diagnostic.

## What changed

The leading-required-miss bar (#125) refuses a barred candidate in `TryEagerCommit`:

```csharp
if (bestLeadingRequiredMissed)
    return EagerCommitVerdict.None;
```

DR-5 places that gate **beside the #66 and #70 completeness refusals** — which puts it *above* both `HoldExtendable` returns:

```csharp
if (_canCommitEarly == null)
    return EagerCommitVerdict.HoldExtendable;

if (!CanCommitEarly(bestCommandIdx, bestPatternIdx))
    return EagerCommitVerdict.HoldExtendable;
```

`HoldExtendable` is what arms the shortened wait — `_eagerHoldArmed` feeds `EffectiveBufferWindow`, which returns `prefixHoldSeconds` instead of `bufferWindow`. So for a barred winner on an extendable or un-analysable pattern, the verdict flips `HoldExtendable → None` and the shortened wait is never armed.

## Impact

**Latency only. Nothing fires differently** — a barred utterance reports `OnUnrecognisedSpeech` either way; it just reports it later.

- **Inert at the shipped default**, where `prefixHoldSeconds = 0f`.
- **Not inert where it is set.** Issue #124's own reporter runs `bufferWindow 2.0, prefixHoldSeconds 0.6`, so on that configuration a barred utterance takes ~2.0 s instead of ~0.6 s to say "didn't catch that" — about 1.4 s of extra silence, on exactly the shape #124 is about.

## Why this is worth ruling rather than just accepting

The sibling-tie refusal ~60 lines below was **deliberately** placed *after* both hold returns, and its comment states the principle:

> a refusal must not lengthen a wait it is not responsible for. Here the only transition it introduces is Commit -> None

The bar breaks that principle. The argument for accepting it: a tied candidate may still fire after the wait, so lengthening it buys nothing, whereas a barred candidate produces no result at **any** wait length — so the only cost is how soon the speaker is told. That distinction is real but was reasoned at implementation time, not ruled at G1.

## Options

[...]

PR #131, opened 2026-08-27
Document the leading-required-miss bar (#124 item 2)

jinwoo1601:
Backlog **item 2** of the G1-locked #124 design. Item 1 (#125, `41ea036`) changed parser behaviour and touched **no** documentation; this closes that gap. **Blocks item 4** — cutting 2.0.0 first would stamp a release whose troubleshooting page cannot work.

Scope was ruled at G0 as the **full #129 surface** rather than DR-9's `scoring.md` §7 D alone, plus two ride-alongs (`scoring.md` §1's row, and the `Readme` string shipped inside every exported session log).

## Three pieces of advice were reversed, not merely stale

| site | said | now |
|---|---|---|
| `troubleshooting.md` | lengthen the pattern | works for a middle/trailing miss; **no length** recovers a dropped *first* required element |
| `command-recognition.md`, `KNOWN_LIMITATIONS.md` | "make the difference come earlier" is under **what does not help** | it is now among the more effective mitigations, **at any length** |
| `KNOWN_LIMITATIONS.md` repro | `["cease","fire"]` vs `["resume","fire"]`, first word dropped | **could no longer reproduce anything** — both candidates barred, tie never live. Replaced with a non-leading discriminator |

`scoring.md` gains the rule's canonical section and a **barred winner** glossary term; ~37 sites link to it rather than restate it. A new known-limitation entry records the bar's own cost.

## Measured, not asserted

Both rig arms validated **before** any delta was read — `--grammar` byte-identical to the committed 133-word pin, replays byte-identical to the committed corpus TSVs.

- **The published sweep figure does not move.** 28 blocked / 27 recovered / 1 silent on *both* arms. #129 expected this to need re-deriving. It does not: the bar advances `searchStart` exactly as an emitting round does, so it can only remove results, never change which rounds occur. The one silent case is still `"switch navigation mode"`.
- **The bar's cost is 9 rows of 699** — not #129's `17`, nor design §7.3's `9–12`. Those counted different things: 17 is every row now firing nothing (fixes *and* costs), 9 is the genuine loss. All nine are `delete`-category rows, against 39 invented commands suppressed and **0** clean `1.00` commands destroyed.

## Verification

- **EditMode 133/133, PlayMode 569/569** — unchanged from `41ea036`.
- **`DocCheck` 50 → 66 assertions**, all green against the shipped parser. Mutation check (same pin, pre-bar parser `b83b62f`) turns **9 red** — one per behavioural claim these docs publish, including the P1 reversal.
- All intra-doc anchors resolve across 20 files.
- **Deferred, human-only:** on-device Quest verification. No native bridge, MonoBehaviour lifecycle or audio path touched.

> The `DocCheck` work is under gitignored `Planning~/` and so is **invisible in this diff**. Its evidence is the numbers above.

## Review

An 11-angle `review-cycle` pass plus 5 adversarial verifiers produced 30 applied findings — including **5 verified-false claims this branch had introduced**, the worst of which reinstated the exact misconception issue #47 had fixed. The audit chain under-counted three times, each miss caught by a different method than the one that missed it:

- #129 missed `scoring.md`'s glossary definition of **Winner**;
- the claim-string sweep missed *"A command stopped firing after upgrading"* — the upgrader's own door, which answered "this is coverage" for a symptom coverage cannot fix — because that section contains none of the swept strings;
- `DocCheck` was red in **two** places where the plan validator predicted one, which independently confirmed the §1 ride-along was necessary.

## Canon note — ERRATUM E-2

The G0 ruling that item 2 stays silent on the sibling over-warning **falsifies locked canon**: design §9 calls item 3 *"independently valuable and independently droppable"*, and it no longer is. Ruled erratum-plus-pointer (no design branch, no G1 re-lock), on the E-1 precedent. **Item 3 is now a hard dependency of item 4.**

## Not in scope

No behaviour change; no version number (item 4). #126 / #127 / #128 stay open — including #128's two unpinned tests, which are test work. The construction-time warning belongs to item 3.

Refs #130, #129


jinwoo1601 (commented):
## `review-pr` — full profile · head `8683bcd` · base `41ea036`

All 12 angles run (`CLAIM` + the full `angle-roster.md`). ~45 raw findings → **24 distinct**; 8 sent to adversarial verifiers, **3 refuted**. This is the second independent pass on this branch — a working-tree `review-cycle` ran before the PR opened — and it found materially more, because it was briefed to hunt what a self-briefed pass structurally cannot.

**Headline:** the measured claims hold up under independent re-derivation. The defects are in prose, and **most of the serious ones are in text written while fixing the previous round's findings.**

---

## Tier 1 — false statements this branch introduced

### 1. `troubleshooting.md:69` — the diagnostic tests for a condition that cannot occur · **VERIFIED** · found by 2 angles

> "**If the log holds no attempt at all for that utterance, this is not coverage.**"

No utterance can have an empty `attempts` array. `VoxrCommandRecogniser.cs:887-897`: when the parse yields nothing, `if (resultCount == 0)` publishes `new[] { new VoxrMatchAttempt(null, null, 0f, …, "no match", false) }` — exactly one, never zero. Where `resultCount > 0`, every branch of the Step-7 loop adds an attempt, with no `continue` skipping the add.

So the reader is told to look for something impossible, finds the synthetic `no match` row, and concludes the diagnosis does not apply — the inverse of the intent. This sentence was **added by this branch** to fix a different finding. `scoring.md:524` states the same fact correctly 455 lines later.

**Fix:** adopt the scoping `KNOWN_LIMITATIONS.md:450` already uses — *no scored attempt **for that pattern***.

### 2. `scoring.md:512` — false causal attribution, contradicting the same example 14 lines up · **VERIFIED**

> "The bar is applied *after* selection precisely so this stays true: a barred candidate still terminates the orphan run and still consumes its span."

The locked design settles it: *"orphan-run termination … is reached by a code path the bar does not touch. **A bar placed in the comparator is invisible to it.**"* (`leading-miss-bar.md:69`). `BuildCoverageTables` runs at `VoxrCommandParser.cs:2275`, before the round loop at `:2282`; `IsAdmissibleStart` reads only tokens and registered patterns. Under the *rejected* refuse-to-compete fork the orphan table would be byte-identical.

Second conjunct ("still consumes its span") is true and genuinely caused by the placement — but bound to the wrong effect. `:498`, edited in this same commit, says outright that the start test and *not* consumption holds round 1 at `1.00`.

### 3. `command-recognition.md:319` — "Nothing fires either way" falsified by the PR's own canonical example

On `"target hotel one cease fire"` the eager gate refuses (bar returns `None` at `VoxrCommandParser.cs:4255`), the buffer waits the full window — and at the flush round 1 is barred while **round 2 fires `cease_fire` at `1.00`**. A command fires; the utterance is not reported unrecognised. **Fix:** scope to a buffer whose only round is the barred one.

### 4. `KNOWN_LIMITATIONS.md:434` — a universal the code does not support

> "…so it **cannot change which later rounds occur** — only whether they produce a command."

A barred round does not advance `_resultCount`, but the loop breaks on `if (_resultCount >= _resultBuf.Length)` (`:2293`) where `_resultBuf` is sized to the command count. One-command grammar `["cease","fire"]`, utterance `"fire cease fire"`: bar **off**, round 1 emits at `0.50`, fills the 1-slot buffer, loop breaks, round 2 never runs. Bar **on**, round 1 is barred and round 2 emits at `1.00`. The bar changed which rounds occurred.

The sentence's *purpose* — defending the 28/27/1 re-sweep — survives, because the corpus grammar registers 11 commands. **Fix:** qualify to "cannot change **where** later rounds start", or add the buffer condition.

### 5. `api/batch-test-runner.md:53` — the `ActualIntent` row contradicts its own page · found by 4 angles

The row attributes a null to "**the round the runner reports**" being barred. The runner never reports a barred round — line 46 of the same page says so. In code the barred case reaches null only via `if (parseResults.Length == 0)` (`VoxrBatchTestRunner.cs:114-123`), where no round object exists, whereas `required slot unfilled` genuinely comes from a reported round. It also drops the "**every** round" condition the prose above states correctly. **Fix:** "or when every round was barred".

### 6. `scoring.md:520` — "five short-circuit before the parse"; the correct count is **three** · **VERIFIED** · 2 angles

Step 3's `_parser.ParseInternal(...)` at `VoxrCommandRecogniser.cs:717` is unconditional. The follow-up slot-fill entry (`:867`) and the refused re-score (`:825`) both publish from Step 5 (`:772`), *after* it — and Step 5's guard reads the parse's own `resultBuf`. Only the confirm/cancel resolution, the disambiguation answer (both `:695`) and the pending timeout (`:546`, no parse at all) precede it. The `// (before parsing)` comment at `:710` describes the slot-fill *attempt*, not the publication.

### 7. `scoring.md` §7 D — the rewrite exceeded DR-9's licence, and the heading now contradicts the body · **VERIFIED**

Design §5.7 licenses exactly: *"the amendment is small: **step 1 is unchanged and step 2 inverts** … The replacement keeps the same grammar and the same utterance."* Step 1 is byte-identical and step 2 inverts as specified — but the **premise** was also changed, which is neither step:

```
-Utterance: … the speaker said both commands; VOSK dropped the second one's "approach".
+Utterance: … the speaker said one command; VOSK's output also happens to fit the tail of another.
```

In the same commit the heading was made *more* drop-specific — "loses a word" → "loses **its first** word" — and `:510` still says "**its second command's** verb was never spoken", presupposing a second command the new premise denies.

The premise is also contested in canon: `scoring-model.md:168` and `coverage-in-selection/architecture.md:319` both call this utterance "a genuine two-command utterance whose second command lost its 'approach'". Flipping it silently, in a document that elsewhere insists the two cases are indistinguishable, is the issue. **Fix (author's call):** restore the base premise, *or* change the heading and `:510` to match the new one.

---

## Tier 2 — false statements the branch did not reach

### 8. `command-recognition.md:489` — the `disambiguateSiblingTies` promise · **VERIFIED** (severity lowered)

> "**`disambiguateSiblingTies` is what replaces that guess with a question.**"

False where the discriminator is the first required element: the bar `continue`s at `:2549`, above the `_tiedSiblingBuf` write (`:2590`), so no tie is recorded; `TryBuildAmbiguity` is driven by the `for (i < resultCount)` loop, which does not iterate. Nothing fires and nothing asks.

Verifier lowered severity because remedy 3 (`:440`) *does* carry the caveat — but the section itself never names the bar, unlike every other feature section in this file, and readers arrive cold from `scoring.md`'s two links. **This is also a semantic broken link:** the new remedy sends readers here.

### 9. Two surviving copies of a claim this branch narrowed elsewhere · 2 angles

`command-recognition.md:434` and `troubleshooting.md:100` both still say a lowered `minScore` makes short sibling pairs live, and both cite `KNOWN_LIMITATIONS.md` as authority — which now reads *"differing at exactly one **non-leading** required word"*, and whose repro was replaced **because** the leading shape is no longer live. Both pointers resolve; neither target supports the claim at the pointer.

### 10. `troubleshooting.md` "Commands not matching" — the first symptom door · 3 angles

`index.md` routes users here explicitly ("start here if the wrong command fires, or none does"). Its ladder still opens with *"Lower `minScore` and `minConfidence` temporarily"* — which by construction cannot reach a positional refusal. Its remedy list at `:106` also omits the leading-discriminator fix that this PR promotes to remedy 2 and pins in `DocCheck`.

This is the **same structural miss** the branch already hit twice: the site list was closed by grepping claim strings, and this section contains none of them.

### 11. The eager-flush enumeration has no bullet for the bar · 2 angles

`scoring.md` gained condition (6); `command-recognition.md:290-299` — which has a bullet for every *other* refusal — gained none, so that buffer still falls under "**Complete and unambiguous** → fires immediately". Same file's prefix-hold bullet at `:319` *does* cite the bar, so the two sections disagree. `api/command-recogniser.md:16` has the same asymmetry against its own `:17`.

---

## Tier 3 — invariant violations and consistency

### 12. Winner-vs-rival phrasing survives in four places · 3 angles

`command-recognition.md:440`, `KNOWN_LIMITATIONS.md:650` and `:660` say "**both candidates** are barred"; `scoring.md:61` says "a **pattern** whose first required element matched nothing is barred". Only the round **winner** is ever barred — `bestLeadingRequiredMissed` is assigned solely inside `if (order == CandidateOrder.Better)` (`:2383`), and every loser's flag is discarded. Whether DR-1 extends to rivals is unruled (#126). Remedy 2's own wording two lines above `:440` is already correct and is the model.

### 13. The shipped `Readme` string was not carried through · 4 angles

Two omissions in `Editor/VoxrDebugSessionLog.cs`: the `no match` gloss still reads "*nothing was extracted*" without the caveat both Markdown mirrors gained; and its one-line summary of `scoring.md`'s contents was not updated, while the two parallel copies in `index.md` and `README.md` were. This is the only description a reader of a raw exported log has.

### 14. `CHANGELOG.md` publishes 17 and the docs publish 9/39, unreconciled · 2 angles

Item 1's live entry says "17 fire nothing"; the new sites say "9 rows lose a genuine command, against 39 invented ones suppressed". `9 + 39 = 48`, and architecture §4.1 resolves it — 17 is the fire-nothing set, costs and fixes together — but that resolution is in a gitignored file, and the two numbers sit in adjacent paragraphs of the same `Changed` block.

### 15. The glossary definition is split across two rows · 2 angles

`Winner` (unedited) still reads "Only winners reach the gates, and only winners are logged as a scored attempt"; `Barred winner` was added beside it. Not false — both are one-directional — but a reader taking `Winner` at face value concludes the inverse of the bar, and `Winner` already carries its other exception inline with an explicit "One exception:" clause.

### 16. `editor-testing.md:157` — missing fallback · **VERIFIED, narrowed**

"a barred round contributes no line" is literally true in both cases; what is missing is that an all-barred case falls back to one `(no match)` line. Matters because `:212`, edited by this branch, actively invites authoring that case.

### 17. Two stale enumerations · 2 angles

`command-recognition.md:110` still says "**All three** are covered in Authoring hazards" (now four); `inspector-authoring.md:99`'s rewrite now implies all three listed scans live there, but the >12-optionals scan is documented under Eager flush.

---

## Tier 4 — PR body claims (`CLAIM` angle)

The corpus figures were **independently re-derived twice**, by two angles, from the committed arms — 48 changed / 17 fire-nothing / 9 genuine / 39 suppressed / 0 clean-`1.00` destroyed, all exact. `DocCheck`'s 66 was counted from source. Four framing defects remain:

**18.** The transparency blockquote names only `DocCheck`, but claims 1–3 rest on the same gitignored tree; "its evidence is the numbers above" is circular; and it is **over**-pessimistic — the corpus claim is fully re-derivable from committed data with no build, which is exactly what two reviewers did. It should point a reader at `issue124/*.tsv` and `analyse-arms.py` and say which claims need a build.

**19.** "~37 sites link to it **rather than restate it**" — the real count is **44**, and **28 of those lines also restate**. The branch's own architecture doc says "20-odd", matching the 24 outside `scoring.md`.

**20.** "#129 expected this to need re-deriving. **It does not**" overstates. `blocked` is structurally immovable; `recovered` is only **corpus-contingent** — the shipped text at `KNOWN_LIMITATIONS.md:618` carries that qualifier, and the PR body drops it. The figure was re-measured, not reasoned.

**21.** "nor design §7.3's `9–12`" is not a correction — 9 is the bottom of that stated range. Correct framing: *narrows `9–12` to exactly `9`*.

---

## Tier 5 — structure, and the instrument

**22.** `ALT`: the canonical section is filed as a `###` under "## 5. The two gates", while `command-recognition.md`'s diagram deliberately places the bar *within* Sequential Extraction — and both hub blurbs enumerate it as a top-level peer between §3 and §5. Three shipped surfaces, three different structural stories.

**23.** `ALT`: remedy 2 is inserted into a list introduced as "in the order worth trying" whose item 4 says "worth doing *alongside* 3" — but remedy 2 *forecloses* 3 and 4 (no pending ever opens). A reader stopping at 2 never learns that.

**24.** Gitignored rig (does not ship; affects only published figures): `stage.sh:46` uses `git diff --quiet -- Runtime`, which compares worktree against **index** — a staged-but-uncommitted `Runtime/` change is stamped clean, defeating the provenance guarantee its own comment promises. And `DocCheck.cs:894` still carries a `Main`, making this branch's new csproj comment ("Only Program and PlayerCheck still carry a Main") false. Neither is triggered by this head.

---

- **`scoring.md` §6's sibling-tie bullet** — alleged to repeat finding 8's falsehood. **Refuted:** §6 is an explicit conjunction; the bar is condition (6) under "requires **all** of", and the bullet sits under "`Commit` **additionally** requires **both** of". A leading-discriminator tie fails (6) and never reaches it. The document's structure already encodes the code's ordering (`:4255` before `:4325`).
- **The `KNOWN_LIMITATIONS` repro's "with the demo grammar" premise** — alleged unverified. **Refuted** by three angles independently against `DemoGrammar.cs`, `CommandDemo.cs` and `Cmd_SetHeading.asset`: all three `set_heading` patterns are three required elements, so `2/3` = `0.667` and all are barred. The repro is exact, and the corpus carries the row.
- **`scoring.md`'s `"target hotel one cease fire"` trace** — alleged unreachable. **Refuted:** `CompareCandidate:3165` returns on earliest start before score is read, and `cease_fire` at token 3 is charged three skipped tokens for `2/(2+3)` = `0.40` in round 1, so `approach_target` wins on both keys. Corroborated empirically: the refuse-to-compete arm yields exactly `cease_fire:0.400000` on that row.

---

## Verified clean

- **All arithmetic**: `(0+1)/2`, `(1+0)/2`, `2/3`, `6/7`, `2/(2+3)`, `5/5`→`4/5`, and the corpus figures — re-derived independently by two angles, no error found.
- **§7 D end to end**, including the contrast trace's half-open slot range `[4, 6)`.
- **All ordinals and renumberings**: remedies 1–5 with "remedy 4 above" correct, "six situations" = 6, "None of the seven" = 7, "the third of those rejection causes" correct, "but see 5 and 6" correct.
- **Precedence on both paths**: the bar precedes `allowPartialMatch`, `requiresConfirmation`, `disambiguateSiblingTies`, completeness and both thresholds; on the eager path it sits above both `HoldExtendable` returns, so `prefixHoldSeconds` is genuinely never armed.
- **179 intra-doc anchored links, 0 broken** across 21 files, including the renamed §7 D anchor and its single referrer.
- **Invariant 3**: the three `allowPartialMatch` copies are byte-identical apart from link depth.
- **Invariant 6/7**: `[Unreleased]` extended not replaced, item 1 verbatim intact, `package.json` untouched at `1.5.0`.
- **The one compiled file**: still a compile-time `const`; non-ASCII histogram byte-identical to base; zero characters reaching JSON's escape set; no test asserts it; no `SchemaVersion` bump owed (precedent cited in the 1.5.0 entry).
- **The rig's cross-revision reflection** is null-guarded and binds correctly on both staged parsers; its float comparisons mirror the shipped comparator key for key.
- **Three untouched sites checked and correctly left alone**, with reasons recorded.

## Not verifiable from here

The Unity host-project run (EditMode 133/133, PlayMode 569/569) — cannot execute here; assessed only as *consistent* with a diff whose sole compiled change is a string literal, and `Tests~/Editor` does carry exactly 133 test-attribute sites. The `Sweep`/`DocCheck` re-runs and the mutation check's "9 red" need a build. `DocCheck`'s "50" baseline is unverifiable — `Planning~/` is untracked, so no base revision exists. On-device Quest verification is human-only and remains deferred.

---

*Posted by `review-pr` (full profile). Findings are evidence for the human's G2 ruling; nothing here has been fixed.*

jinwoo1601:
## All 24 findings addressed — `c8829f9`

Every numbered finding from the review above is fixed, plus the conventions items and two stale process-doc figures. Three findings stand **refuted** and were not "fixed" — they are recorded in the review as considered and killed.

**One finding needed an author's choice.** §7 D offered two ways to make heading, body and premise agree: restore the base premise, or keep the new one and change the heading plus `:510`. I took **restoring the base premise** — it keeps DR-9's licence intact ("step 1 unchanged, step 2 inverts") rather than widening it after the fact, and canon elsewhere describes that utterance as a genuine two-command drop. The "never spoken" sentence became a statement about the transcript's ambiguity, which is what the rest of the docs say.

**Two structural findings resolved by moving the docs, not the text:**
- §5 is retitled **"The bar and the two gates"**, so the heading, its own opening sentence and both hub blurbs finally agree. Both `#5-the-two-gates` referrers updated.
- The `Winner` glossary row now carries its exception inline (matching its existing "One exception:" idiom) instead of leaving the definition split across two rows.

**The one I want to flag rather than bury:** the impossible-condition diagnostic in `troubleshooting.md` was text *added by the previous fix round* to close the upgrader-door finding. Fixing a documentation defect introduced a new one, which is the second time on this branch. That is the argument for this PR-stage pass having existed at all.

### Verification after the fixes

| check | result |
|---|---|
| EditMode | **133/133** — third consecutive run at baseline |
| PlayMode | **569/569** — unchanged |
| `DocCheck` | **66/66** green against the shipped parser |
| Mutation check | **9 red** against pre-bar `b83b62f` — unchanged, so the pins still bite |
| Anchors | **185** anchored links across 20 files, **0 broken**, across two heading renames |
| Claim sweeps | no survivor of any corrected claim; no reversed advice anywhere |

Deferred and unchanged: on-device Quest verification is human-only.

Issue #133, opened 2026-08-27
OnUnrecognisedSpeech fires in the same frame a partial-match or confirmation pending opens


jinwoo1601:
Found while writing the #128 test pins on `feat-leading-miss-warning` (#124 item 3). **Pre-existing, unrelated to that work** — verified by reading `Runtime/Commands/VoxrCommandRecogniser.cs` on `main`'s behaviour, not by a test.

## The defect

Step 7 of the result loop tracks `anyThresholdFiltered` so that a result which was *handled* rather than *rejected* does not also get reported as unrecognised. At the tail:

```csharp
if (acceptedCount == 0)
{
    if (!anyThresholdFiltered)
        OnUnrecognisedSpeech?.Invoke(text);
    return;
}
```

Three branches set the flag. The **disambiguation** branch sets it with a comment naming precisely the harm it prevents:

> Without this the utterance reaches acceptedCount == 0 with the flag clear and raises OnUnrecognisedSpeech — telling the integrator the speech was not understood in the same frame it was asked to prompt about it.

Two other branches enter a pending and `continue` **without** setting it:

1. **Partial match** — `EnterPending(cmd, partialDef, unfilled, VoxrPendingReason.PartialMatch, ...)`, inside the `cmd.Score < minScore || incomplete` block.
2. **Requires confirmation** — `EnterPending(cmd, confirmDef, Array.Empty<string>(), VoxrPendingReason.AwaitingConfirmation, ...)`.

Both `continue` without incrementing `acceptedCount`. So when such a pending is the only outcome of an utterance, the tail sees `acceptedCount == 0` with the flag clear and fires `OnUnrecognisedSpeech`.

## Impact

A subscriber is told the speech was not understood in the same frame `OnCommandPending` told it to prompt the speaker. The two obvious integrations both misbehave: a "didn't catch that" voice line plays over the confirmation prompt, and any UI that clears on `OnUnrecognisedSpeech` tears down the prompt it was just asked to show.

`allowPartialMatch` and `requiresConfirmation` are both opt-in, which is likely why this has gone unnoticed — the default grammar reaches neither branch.

## Not a regression from #124

The leading-required-miss bar refuses these candidates *before* they reach either branch, so the bar makes this fire **less** often, not more. The defect predates it.

## Repro sketch

[...]

Issue #132, opened 2026-08-27
Cancel-collision report is a false advisory where the tie can never be asked about


jinwoo1601:
Found while building #124 item 3 (`feat-leading-miss-warning`). **Pre-existing — not a regression from that branch**, which is why it was ruled out of scope there rather than fixed. Filed so the evidence is not lost.

## The claim

`BuildCancelCollisionWarning` tells an author that a discriminating value is also cancel vocabulary, so if the ambiguity were routed back to the speaker, answering with that word would cancel rather than choose. The remedy it offers is to rename a grammar literal.

The report's own comment in `VoxrCommandParser.cs` states the standard it is held to:

> Reporting a collision on a value the runtime would never ask about is a knowingly false advisory, the class issue #81 spent a whole feature reversing.

## Where it fails that standard

The collision loop gates on exactly two things: the value is in the effective cancel vocabulary, and `HasAnswerableCoMember(set, m)` finds a co-member with both a different value and a different intent. It inherits **neither** of the sibling warning's two reachability gates.

Two shapes therefore get the advisory although no question can ever be posed:

1. **Below the score gate (older).** A two-element set scores `(2-1)/2 = 0.5`, under the default `minScore`, so both siblings are rejected and nothing fires. The sibling warning is withheld; the collision report is not.
2. **Leading discriminator (since #124).** Where the discriminating word is every member's first required element, dropping it bars whichever member wins the round. No result is emitted, so no pending opens and no "which did you mean?" is asked. Item 3 added the anchor condition to the sibling warning; the collision report does not inherit it.

In both cases the author is asked to rename a literal to protect an answer that can never be given.

## Why it was not fixed in item 3

Item 3's requirements scope the work as *"the sibling scan is narrowed, not redesigned"*, and the collision report is narrowed **deliberately differently** — per-PAIR rather than per-SET, because a set can be cross-intent overall while a particular pair inside it shares an intent. Making it inherit both gates is a design decision about that split, not a one-line extension of item 3's condition. Ruled out of scope by the human at item 3's implementation, 2026-08-28.

## Sketch of the fix

Shape 2 alone is cheap and exact: if every member is leading-anchored, no pair inside the set can be asked about, so the existing set-level anchor predicate applies unchanged. Shape 1 is the harder half — it needs a ruling on whether a per-PAIR report should inherit a per-SET score gate, and that gate is itself known-limited because the parser is never handed the recogniser's configured `minScore`.

[...]

PR #134, opened 2026-08-27
Narrow the sibling over-warning; pin three bar guarantees (#124 item 3, Deliverable A)

### LINKED ISSUE #128: Two leading-miss-bar guarantees are unpinned: DR-2's slot half on the pending path, and two CHANGELOG claims

jinwoo1601:
Deliverable A of #124 backlog item 3, plus #128's three test pins. **Editor-only — no runtime behaviour change.** Backlog tracked at #130.

Closes #128.

## What was wrong

The construction-time sibling warning concludes its diagnosis with *"so the wrong intent can fire"* and then prescribes three remedies. Since the leading-required-miss bar (#125), that is **false** for a whole class of grammar: where the discriminating word is **every** member pattern's first required element, dropping it leaves every candidate leading-missed, so whichever wins the round is barred, the round yields nothing, no command fires and no pending opens for that set.

Authors with such a grammar were being sent to rewrite it against a hazard that does not exist — the class of claim #81 spent an entire feature reversing. The code had deferred this deliberately, with a comment explaining why the gate was *conservative rather than wrong*.

## The quantifier is the whole point

**Every member anchored** → nothing fires; warning correctly withheld.
**Mixed set** (the word leads one pattern, not another) → the candidates still tie exactly, `MatchedRequired` is not a `CompareCandidate` key, so registration order can hand the round to a candidate the bar does not touch, which fires. The tie is live, `disambiguateSiblingTies` works, and the warning correctly still fires.

That distinction **did not exist before the bar**, so prose written near it silently encoded the any-rule. Six sites are corrected. The one that mattered most was prescriptive: a remedy telling authors to make the word *"the pattern's"* first required element — singular — which if half-applied produces exactly the still-hazardous mixed set.

The new condition is **positional and reads no configuration**, unlike the existing `minScore`-based exclusion, so it is exact at every threshold rather than judging against the default.

## Implementation

- `DiscriminatorIsEveryMembersAnchor(SiblingSet)` — walks each member's **authored** pattern for the first element crediting a required literal *or slot* (`CreditsRequired`, matching the runtime latch's own guards, per DR-2), and compares against that member's **own** `AuthoredDiscriminatorIndex`.
- One added conjunct on the gate. `BuildSiblingWarning`'s message text is **unchanged** — the claim is true of everything that still reaches it.
- Two pre-existing tests used a leading discriminator incidentally and were **re-fixtured to interior discriminators**, not deleted or relaxed. One is the only pin on the remedy string this PR promises to preserve; the other is the deliberate intent-dedup guard.

Three ways to get the predicate wrong, each pinned by a test that a plausible wrong implementation fails:
- using the literal-only predicate (skips a leading required slot → wrongly withholds)
- comparing against the frame's `DiscriminatorIndex` rather than the member's authored one
- requiring *any* member anchored rather than *every*

## #128's pins

All three published-but-unguarded guarantees now tested, each with a **control** on a near-identical grammar so it cannot pass incidentally, and and each non-vacuous for a reason stated in its own comment — these are **two different arguments**, not one. The `requiresConfirmation` and `disambiguateSiblingTies` pins sit below the `minScore` filter and need `D >= 3` to clear it (`2/3 = 0.667`); at two elements they would score `0.5`, be rejected a step earlier, and pass with the bar deleted. The **slot-anchor** pin needs no score floor at all and scores exactly `0.50`: a missed required *slot* takes `RequiredSlotMissPenalty = -1.0` where a missed required *literal* takes `0`, and the partial-match branch it exercises sits inside the rejection body and reads no threshold. Its own control asserts a pending opening at `0.50`.

**Mutation-verified.** With the bar disabled at both adoption sites, all three fail in the **barred** half while their controls hold; the construction-time tests stay green, confirming the mutation reached only the runtime bar. 14 tests fail under that mutation, every one bar-dependent.

## Verification

**133/133 EditMode, 578/578 PlayMode.** On-device Quest verification does not apply — Editor-only, no native-bridge or lifecycle change.

## Filed along the way

- #132 — the cancel-collision report has the same false-advisory problem, pre-existing, ruled out of scope here
- #133 — `OnUnrecognisedSpeech` fires in the same frame a partial-match or confirmation pending opens; verified in source, predates #124

## Not in this PR

Deliverable B — the new construction-time scan for the bare-pattern-tail shape — is a later phase whose detection rule is deferred to a human ruling at a measurement (design §9's clause is one sentence and specifies no firing condition). Early evaluation suggests the candidate rules fire on 17 of 32 shipped demo patterns, so it may be cut.

Canon note: design §9 scoped item 3 to the new scan alone while ERRATUM E-2's non-droppability argument required this narrowing. Recorded as **ERRATUM E-3** (erratum-plus-pointer, no G1 re-lock) on the E-1/E-2 precedent.

---

**Post-review.** An independent 12-angle `review-pr` pass at the full profile found **no blockers and no behaviour defect**, and ten confirmed findings — every one a claim wider than the change rather than a logic error. Three further claims were refuted and are named in the review. G2 accepted with all ten ruled fix-before-merge; the two corrections above are the findings against this body.


jinwoo1601 (commented):
# Independent review — full profile

**Profile:** full (feature lane, G2 at stake). All 12 angles run, none dropped: `CLAIM` + `LINE`, `GONE`, `XFILE`, `PIT`, `WRAP`, `REUSE`, `SIMP`, `EFF`, `ALT`, `CONV`, `GAP`.
**Reviewed:** head `d8c3db23fe94ea5bb9b684330123e01a11929add`, against merge-base `747c8dfa3f5227a19433a7456658ddd729f3d3cd`.
**Finders inherited no context** from the session that wrote the code. Every numbered finding below carries an independent verifier verdict; three claims were refuted and are named at the end rather than quietly dropped.

## Verdict in one line

**No blockers. No behaviour defect.** Six angles came back clean, and the diff's correctness was confirmed mechanically. Every one of the ten confirmed findings is about a **claim being wider than the change** — comments, docs, and the PR body asserting more than the code supports. Which is the same failure mode this PR exists to fix.

## What was verified clean

- **Invariant 1 (no runtime behaviour change) — confirmed mechanically.** Stripping comments and whitespace, the entire 100/31-line production diff reduces to **exactly two code changes**: one added conjunct and one new method body. No parse, score, selection, eager-gate or flush code differs by a token (`EFF`).
- **The predicate agrees with the runtime latch exactly.** `CreditsRequired` is the precise complement of `TryMatchScored`'s `!isOptional` guards, verified across required/optional literals and slots and three malformed element shapes `IsRequiredLiteral` declines (`WRAP`, `LINE`).
- **Index space, quantifier, precedence, and the fixture arithmetic for all nine new and re-fixtured tests** re-derived independently (`LINE`).
- **Both re-fixtured tests still pin what they guarded** — the remedy-string regex is byte-unchanged and still fires; the intent-dedup assertion still names both intents once (`GONE`).
- **No shipped-sample warning silently vanished.** The four leading-discriminator demo sets were already withheld by the score gate; `mode_weapons@3` still warns (`GONE`).
- **All three consumers of the withheld warning traced**, and the claim that no shipped doc described the construction-time cap advisory was independently verified (`GONE`).
- **`[Conditional("UNITY_EDITOR")]` semantics**, struct copy costs, allocation, and the A/B rig's continued compilability (`PIT`).

## Findings

Ranked severity, then confidence. Ten confirmed; none blocks merge on correctness grounds.

### 1 — moderate · CONFIRMED — the mixed-set rationale is published in five documents and pinned by nothing

`Runtime/Commands/VoxrCommandParser.cs:1672-1677` states the justification for the every-not-any quantifier: the two candidates "tie EXACTLY, because MatchedRequired is not one of CompareCandidate's keys… registration order still decides the round, and one order hands it to the member that is NOT barred, which fires." That claim now also ships in `KNOWN_LIMITATIONS.md` (×2), `CHANGELOG.md:14`, `command-recognition.md:436`, `troubleshooting.md:99` and `scoring.md:240`.

The verifier confirmed **the claim is true**, analytically: on the fixture's own grammar both candidates reach 2/3 with identical `LiteralCount`, `ConsumedEndIdx` and coverage, and `MatchedRequired` is structurally not passed to `CompareCandidate` — so `Tied` holds and registration order decides.

What is unpinned is the rationale. `SiblingWarning_MembersDisagreeingOnTheAnchor_StillWarns` asserts only that the construction-time warning fires; it never calls `Parse`. No test in the suite parses a mixed-anchor pair — the named near-misses use a trailing discriminator or an all-anchored set. The tie rests on two contingent facts: a matched optional slot crediting exactly `MatchScore` on both sides of the ratio, and `MatchedRequired` staying out of the comparator. **Break either and the barred member always wins the mixed round, the quantifier becomes wrong, five author-facing documents become false, and nothing in the suite fails.**

This PR exists to close that class — a published claim with no guard is what #128 *is* — and it publishes a new one on its own headline claim.

*Fix:* one runtime pin. Parse a mixed set with the discriminator dropped and assert the unbarred member fires.

### 2 — minor (top of) · CONFIRMED — two new production comments generalise the bar past the round's winner

- `VoxrCommandParser.cs:1801-1803` — "a candidate that missed its anchor may compete and consume but may not produce a result, **on the flush path and the eager one alike**."
- `VoxrCommandParser.cs:1658-1659` — "lets **a candidate** that missed its first required element compete and consume, but not fire."

Both are new in this PR (`git show 747c8df:… | grep "compete and consume"` → empty). The bar reads only `bestLeadingRequiredMissed`; `TiedSiblingRival` carries no leading-miss field and `TryBuildAmbiguity` applies no positional filter, so a leading-missed **rival** can be offered as a disambiguation choice and fire when picked — issue #126, open.

Reachable on this PR's own grammar shape, with a correction worth stating precisely: **with `cease_fire` declared first the barred member wins the round and is discarded; reversing the declaration order opens the path.** The PR's own test comment concedes that order exists ("Reverse the two definitions and a different intent fires").

Severity sits at the top of minor because the over-general clause is not incidental prose — it is the stated **warrant** for a newly added production gate, so a future author extending the bar's reasoning to the choice path reads a false premise as justification.

*Fix:* scope both to the round's winner, as `:2593-2599` already does.

### 3 — minor · CONFIRMED — the `capNote` rationale claims an exactness the score gate does not have

`VoxrCommandParser.cs:2246-2248`: "Each of the three withholds a set that can never put its values to the speaker: a same-intent tie is never routed at all, and neither of the other two shapes produces a result to route."

False of the `(D-1)/D` conjunct, and the same file says so at `:1653-1655` ("An author who lowers it below (D-1)/D makes these ties live and gets no warning — a real limitation"), as does `KNOWN_LIMITATIONS.md:772-776` ("so on the lowered threshold this entry is about, a short tie the warning stayed quiet about is still routed to the speaker").

Counterexample verified concretely: six cross-intent commands `["fire","alpha"] … ["fire","foxtrot"]`, frame `fire|*`, `(2-1)/2 = 0.5 < 0.6` so the warning and cap note are withheld — and at a configured `minScore` of 0.4 the tie is live, all six tie, and `values.Count = 6 > 1 + MaxDisambiguationRivals` is exactly the set that never gets the cap note.

Note this PR gets the identical justification **right** 500 lines away at `:2744-2749`, where it is correctly scoped to the anchor shape alone. The base comment confined the strong claim to same-intent; this PR widened it.

### 4 — minor · CONFIRMED — the PR body's non-vacuity rationale is wrong for one of the three pins

The body says all three #128 pins are non-vacuous because "each at `D >= 3` so `minScore` cannot mask the bar (at two elements the score is 0.5 and the filter rejects it a step earlier)". `LeadingRequiredMiss_SlotAnchor_OpensNoPending` has D = 4 and scores **exactly 0.50** — the value the parenthetical calls fatal.

The arithmetic the body assumes is the literal-miss one: `RequiredLiteralMissPenalty = 0f` gives the other two pins `(3-1)/3 = 0.667`, but a missed required **slot** takes `RequiredSlotMissPenalty = -1.0f`, giving `2/4 = 0.50`.

The test is still non-vacuous, for a different reason it states in its own comment: the partial-match branch sits inside the rejection body and routes on `Score > 0f && AllowPartialMatch && unfilled.Length > 0`, reading no threshold. Its own control asserts a pending opening **at 0.50** — an in-repo counterexample to the body's claim, inside the test the claim describes.

*Nothing shipped is wrong here* — brief invariant 9 already allows the second disjunct. The defect is confined to the PR narrative.

### 5 — minor · CONFIRMED — "the warning ends …" is false, in a user-facing file

`CHANGELOG.md:14` (new in this diff) and the PR body both say the sibling warning **ends** "so the wrong intent can fire". Joining the split literals at `VoxrCommandParser.cs:2263-2267`, that clause is followed by a full remedy sentence naming three remedies, plus the optional cap note.

Correction to the finding as originally written: the clause is *sentence-terminal*, not mid-sentence. "The warning ends its hazard sentence with X" would be true; "the warning ends X" is not.

The same file already contradicts it — `CHANGELOG.md:32`, under 1.5.0, tells users the message "offers only remedies that exist today".

### 6 — minor · CONFIRMED (narrowed) — "three cases where the tie is not a hazard" is wrong about one of the three

`Documentation~/inspector-authoring.md:96`, new in this PR. `"not a hazard"` appears exactly **twice** in the shipped doc set: this clause covering all three exclusions, and `command-recognition.md:432` where it is scoped to exclusion 1 alone. The docs deliberately use different language elsewhere — exclusion 2 is "a **different problem** from the wrong command firing" (still a problem, and `KNOWN_LIMITATIONS.md` documents its symptom as "the wrong command fires").

The finding **overreached on exclusion 3** and the verifier corrected it: canon rules that withholding exact ("the third costs nothing, since it reads no threshold"), and the "silence, the speaker repeats themselves" cost belongs to *remedy 2*, an author's deliberate choice, not to the withholding. Exclusion 3 is fine. The word **"three"** is what makes the sentence wrong.

*Fix:* "withheld in three cases where the warning's claim does not hold".

### 7 — minor · CONFIRMED as clarity (severity lowered from major) — an unresolved quantifier ambiguity at `KNOWN_LIMITATIONS.md:751`

"**The discriminating word must not be the pattern's first required element.**" Pre-existing and untouched by this PR (byte-identical at base), but it is now the lone un-quantified statement in a region where every other one carries an explicit universal — including a line **27 lines below that this PR corrected from "either" to "every" for exactly this reason.**

Both prior readings are wrong. The "harmless over-strict repro constraint" defence fails because the sentence's justification is a **liveness** argument ("the tie is never live"), not a recipe argument — and that is false for a mixed set realisable inside this entry's own silent zone (`{?ship} cease` / `{ship} resume` scores 0.5, so the warning is withheld on the `minScore` ground this entry is about, yet the `{ship} resume` candidate is unbarred and fires at 0.4). But the flat "contradicts" reading is also wrong: in the repro's own two symmetric patterns the singular reads distributively and both readings coincide.

The sentence has one consistent reading and one contradictory one, and the artifact does not choose.

### 8 — nit · CONFIRMED as clarity — `command-recognition.md:491` states the carve-out in pair-scoped terms

"where **the two patterns** differ at their first required element" — the link target of the `disambiguateSiblingTies` API row this PR corrected to "every competing pattern's". Not false (the distributive reading is the every-rule for a pair), but narrow against "A sibling set is n-ary" 33 lines below in the same section. A reader with a three-member all-anchored set reads it cardinally, concludes the exception misses them, and expects a pending they will not get.

The finding's framing that this was "not brought onto the corrected wording" overstates: the six swept sites were sites the bar made *false*; `:491` was never false, so it was never owed the sweep.

### 9 — nit · CONFIRMED — the CHANGELOG entry describes #128's work but links only #124

The bullet's penultimate sentence is #128's exact deliverable, and the PR says `Closes #128`. The file's practice is to co-link: `:16` carries `[#124], [#129]`, and three other multi-issue footers exist (`:34`, `:44`, `:46`). Three refutations were hunted and failed, including "test-only work isn't changelogged here" — disproved by two docs-only issues that do carry footers.

Also absent: **#130**, the tracking issue. Neither #128 nor #130 appears anywhere in `CHANGELOG.md`, `Documentation~/`, or `KNOWN_LIMITATIONS.md`.

### 10 — nit · CONFIRMED (one of two claims) — `:364` mis-dates the silent cap to "#124 item 3"

Base `:2177-2180` already reasoned about withholding the cap note from a set "too long to ask about", and a one-intent 6-synonym set or a 2-element cross-intent set with 6+ values was already capped in silence before this PR. Item 3 added a third way in; it did not introduce the exception. The mis-dating is replicated in `Planning~/features/leading-miss-warning/architecture.md:33`.

## Considered and refuted — named so they are not re-found

- **"`?please cease fire` vs `please resume fire` breaks the tie mechanism, so the CHANGELOG's 'tie exactly' is wrong."** REFUTED. `NormalizeElement` strips `?` from optional **slots** only, never optional **literals**, so that pair keys into different buckets and is not a sibling set. The divergent score is precisely *why* optional literals are not folded — `:1205-1213` says so, and `SiblingSets_OptionalLiteralInTheFrame_IsNotASet` pins it. The CHANGELOG's mechanism is correct for every reachable mixed set.
- **"The gate comment at `:1668-1670` claims something structurally impossible, since `&&` means conjunct 3 can only subtract warnings."** REFUTED. The comment says the conjunct "stays correct" — a soundness claim about its withholdings, which is true — not an emission claim. Clarity risk at most.
- **"`:2239`'s F19 citation now conflicts with the third gate."** REFUTED on three grounds: wrong line, untouched by this PR, and already reconciled in the same block.

- **The `133/578` counts and the mutation-run results are author claims.** No angle reproduced them; this environment cannot run the host-project Unity suites. They were audited as claims, and `CLAIM` found the *arithmetic* consistent (569 + 6 + 3) while finding the mutation narrative's stated rationale wrong for one pin (finding 4).
- **On-device Quest behaviour.** The author's reasoning that it does not apply — Editor-only, no native-bridge or lifecycle change — was checked against the diff and holds.

## One process note, raised by `CONV` and not a code finding

The project `CLAUDE.md` binds the close-out order as: *PR opened **without** product-doc changes → `review-pr` → human rules G2 → **then** product docs pushed to the same branch → human merges.* This PR carries `CHANGELOG.md`, `KNOWN_LIMITATIONS.md` and five `Documentation~/` files **pre-G2**. There is a real argument they belong here — Deliverable A's whole subject is that a shipped doc claim is false, and the doc corrections were written as acceptance criteria — but that argument was owed to the human before pushing, not assumed. Flagged for the G2 ruling.

jinwoo1601:
**Correction to finding 1 in the review above, and the fixes are pushed shortly.**

The review stated that admitting `MatchedRequired` as a `CompareCandidate` key would make *"the barred member always win the mixed round"*. That direction is **wrong**. On the mixed fixture the unbarred `{ship} resume fire` carries `MatchedRequired = 2` against the barred `{?ship} cease fire`'s `1` — the optional slot matches but credits neither counter — so a higher-is-better key would hand the round to the **unbarred** member in *both* registration orders.

The finding's substance is unaffected: the shipped documents claim *registration order* decides a mixed set, and under that change it would not. Only the failure direction was mis-stated, which matters because it changes which assertion catches it — the guard is the **barred-member-registered-first** direction, where the round is expected to yield nothing and would instead fire `resume_fire`.

Both directions are now pinned, so the regression is caught either way. Recording the correction because the review was the evidence weighed at G2, and it was cited from a verifier verdict that carried the same inversion.

PR #136, opened 2026-08-28
Record the barred-rival fire path rather than gating it (#126)

### LINKED ISSUE #126: Leading-missed candidate can still fire as a disambiguation choice
Found by the `review-pr` pass on #125 (the leading-required-miss bar). Not a regression — #125 neither introduced nor widened it — but #125 is what makes it worth naming, because it is the one remaining way a command whose verb was never spoken can still reach `OnCommandRecognised`.

## The path

The bar added in #125 is applied to the **round winner**:

```csharp
// VoxrCommandParser.ParseInternal
if (bestLeadingRequiredMissed)
{
    searchStart = bestEndIdx;
    continue;
}
```

Tied *rivals* are recorded separately, by `RecordTiedSiblingRival`, and that record carries no leading-miss information:

```csharp
_tiedSiblingRivalBuf[slot] = new TiedSiblingRival
{
    CommandIndex = ci,
    PatternIndex = pi,
    Value = rivalValue,
    SlotCount = matchResult.SlotCount,
    EndIdx = matchResult.EndIdx,
};
```

`matchResult.LeadingRequiredMissed` is in scope at that site and is dropped. With `disambiguateSiblingTies` set, the recogniser builds the choice list from those rivals (`VoxrCommandRecogniser.TryBuildAmbiguity` → `BuildSiblingRivalCommand`), and a chosen alternative fires through `PendingCommandHandler`'s `Complete(pending.Choices[i], …)`.

So a candidate whose first required element matched nothing may not fire **as a winner**, but may still be **offered as an alternative and fire when picked**.

## Preconditions — all three required

1. `disambiguateSiblingTies` is enabled (off by default).
2. A sibling set whose discriminator is at **element 0** — the two patterns differ at their first element and agree everywhere else.
3. The speaker drops that first element, so both candidates are leading-missed and tie exactly.

Then the "which did you mean?" prompt offers a command nobody uttered, indistinguishable from the legitimate option.

## Why it was not fixed in #125

DR-1 rules on a *candidate that wins its round*. Extending it to tied rivals is a scope change canon did not decide, and #125 was scoped to build exactly what G1 ruled. It is filed here so the scope question gets ruled on its own evidence.

## The argument for accepting it as-is

The speaker must hear the question and deliberately utter the discriminating word to select the phantom. That is materially different from issue #124's complaint, which was a command firing **with nothing marking it as invented**. This one asks first.

## The argument against

DR-1's text is *"may not fire"*, and this fires. The person answering the prompt cannot tell that one option was never spoken — from their side both look like things they might have said. It is #124's shape (the system proposing an action nobody requested) with a confirmation step in front of it.

## Sketch of the fix

Carry `LeadingRequiredMissed` into `TiedSiblingRival` and skip barred rivals when building the choice list — or record explicitly, beside `PendingCommandHandler`'s existing note on "the property is asserted of one candidate and used on another", why the bar deliberately stops at the winner. Either way it wants a test: no existing disambiguation fixture can reach the bar, because `ConfigureAsking` uses an interior discriminator.

## Related

- #124 — the leading-required-miss bar
- #125 — item 1, where this was found
- The deferred **fork C** design topic revisits this area and may collapse the two bar sites into one; whatever is ruled here should be visible to it.

🤖 Filed from a `review-pr` finding on #125.

jinwoo1601:
Closes #126

Ruled **recorded, not gated**. No behaviour changes — comments, tests and docs only.

## What

- `VoxrCommandParser.RecordTiedSiblingRival` — records why `matchResult.LeadingRequiredMissed` is deliberately not carried into `TiedSiblingRival`, and what the consequence is.
- Two existing comments (`:1671`, `:1820`) that said "issue #126 … still open" now state the ruling.
- `PendingCommandHandler` — the #73 note ("the property is asserted of one candidate and used on another") now says explicitly that its *unreachability* argument does **not** extend to the leading-miss property, because that one carries no score penalty and is not a `CompareCandidate` key.
- Three tests in `VoxrCommandRecogniserInjectionTests.cs` pinning the shape. None existed: no prior fixture ever *parsed* a **mixed-anchor** set with `disambiguateSiblingTies` on, so nothing in the suite could reach a leading-missed rival that is actually **offered as a choice**. (`Disambiguation_LeadingDiscriminator_AsksNothingAndFiresNothing` does record a leading-missed rival — `RecordTiedSiblingRival` runs inside the scan loop, above the bar — but its round is barred before the choice list is built.)
- `KNOWN_LIMITATIONS.md` entry, plus one clause each in `scoring.md`, `command-recognition.md`, `api/command-recogniser.md`, and a `CHANGELOG` bullet.

## Why

The bar is applied to the round **winner** only. A tied sibling rival is recorded *before* it and carries no leading-miss information, so with `disambiguateSiblingTies` on a barred candidate is offered as a choice and fires when picked. Measured against the real recogniser:

```
choice[0] value='at' intent='fire_at' score=0.6
choice[1] value='to' intent='fire_to' score=0.6
afterAnswer fired=1 intent=fire_to
control (fire_to alone, same utterance) resultCount=0
```

**Two claims on the issue were wrong, and both change the ruling.**

1. **Its preconditions occur — its outcome does not.** It asks for a discriminator at element 0 with *both* candidates leading-missed; that grammar and utterance are constructible, and `Disambiguation_LeadingDiscriminator_AsksNothingAndFiresNothing` already pins them. What cannot follow is the offered rival: a discriminator must be a required literal, so at element 0 it is every member's anchor by construction — the uniform case, where whoever wins the round is barred and no rival is ever offered. The only shape that reaches the choice list is a **mixed** set, possible because `{?ship}` and `{ship}` normalise alike, so two siblings can disagree about which element is first-required.

2. **It is not "the one remaining way a verb-less command fires."** In the reachable shape the *winner* is missing the same words — siblings differ at one element and agree everywhere else. `fire_at` survives only because a matched leading **slot** precedes its verb. Both choices fire a command whose verb went unheard.

So gating it would narrow the **question**, not the hazard: in a two-member set `TryBuildAmbiguity` falls below two choices, no question is asked at all, and the equally unheard survivor fires anyway. In the shape where the discriminator *is* the barred member's own anchor, gating would also delete a question whose answer **supplies the missing word** — `Disambiguation_MixedAnchorSet_OffersTheBarredRivalAndFiresItWhenPicked` pins that.

The asymmetry underneath — DR-1's "first required element" stops standing in for "the verb" once a required slot leads the pattern — is the finding worth acting on, and belongs to the deferred **fork C**, which canon already says may collapse the bar's two sites into one.

## Validation

- [x] EditMode 133/133, PlayMode 583/583 (580 + 3 new), Unity 6000.4.7f1, host project `VoXR TestGround`. Result XML `start-time` checked fresh against wall clock.
- [x] All three new tests pass; each accounts for its construction-time sibling warning via `LogAssert`.
- [x] Runtime diff is comment-only — every added line in `Runtime/` starts with `//`, verified mechanically.
- [x] The `requiresConfirmation` mitigation claimed in the docs is real: `Complete` re-enters pending on the **chosen alternative's own** definition (`PendingCommandHandler.cs:337-354`), already covered by `Disambiguation_AnswerRequiringConfirmation_AsksWhichThenAsksAreYouSure`.
- No on-device (Quest) check needed — no native, audio, or lifecycle code is touched.

## Reviewer's attention

Doc wording is the risk surface, not the code. An earlier draft of the three doc clauses said "where only *some* members are anchored on that word" — true of one pinned shape and false of the other, since in the `fire_at`/`fire_to` repro **neither** member is anchored on the discriminator. They now key off the complement of the documented exception instead. Worth checking I have not reintroduced that anywhere, and that `scoring.md`'s addition still reads as compatible with the "does not raise a 'which did you mean?' about it" sentence it follows — that one is about a barred *winner*, so it is not an exception, but it can read like one.

---

### Review follow-up

`review-pr` (slim profile) returned **7 confirmed findings, all of them claims wider than the code supports — no code defect**. All seven are addressed in the follow-up commit on this branch: the runtime comment's definition of the mixed-anchor shape, the `KNOWN_LIMITATIONS` rationale and its missing ordering precondition, the unhedged registration-order claims across four sites, a test that could pass vacuously, a wrong-direction cross-reference, and the two PR-body claims above (corrected in place).

One correction cannot be made in place: the message of commit `e06b720` repeats *"Its stated preconditions cannot occur."* The follow-up commit records the correction rather than rewriting history, so a **squash merge** — which takes this body, not that message — lands the accurate wording.


jinwoo1601 (commented):
## `review-pr` — independent review

**Depth profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE`. No angles added or dropped: the diff crosses no contract boundary (`XFILE`) and touches no MonoBehaviour lifecycle, coroutine or main-thread-only API (`PIT`).
**Head reviewed:** `e06b7209fcb7e6460d706476c4182d25184e3db8`. Base: merge-base `1c892eb`.
**8 findings after dedup → 7 CONFIRMED, 1 REFUTED.** Finder agents inherited nothing from the session that wrote the diff; every numbered finding was then passed to a separate verifier instructed to refute it.

The pattern is worth stating plainly: this PR is almost entirely *claims* — comments, docs, a CHANGELOG bullet — and **every confirmed finding is a claim wider than what the code supports.** No defect was found in the three added tests' mechanics, which verified clean in detail (tie keys, admission arithmetic, `3/5 = 0.60` against `minScore`, warning element numbering, choice ordering, no undrained buffers).

---

### 1. `KNOWN_LIMITATIONS.md:517-521` — major, CONFIRMED
The entry's whole *"Why this is recorded rather than fixed"* rationale is stated as a universal over mixed sets and is **false of one of the two fixtures this PR itself pins**: *"in a mixed set **both** members missed the same words … So both choices fire a command whose verb went unheard."* For `{ship} resume fire` / `{?ship} cease fire` on `"alpha fire"`, the shared literal `"fire"` **matched for both**; each missed only its own discriminating value (`resume` vs `cease`) — different words — and the answer that selects the barred member *is* its missing anchor. The PR's own test comment says the opposite at `VoxrCommandRecogniserInjectionTests.cs:2102-2105`. The stated reason ("siblings differ at one element and agree everywhere else") is a property of *every* sibling set, which makes the conclusion universal by construction; agreeing on an element does not entail it went unmatched.
**Same defective universal also at `CHANGELOG.md:18`** ("so both members of such a round missed the same words").
**Fix:** scope the argument to the shape where the discriminator is *not* the barred member's anchor, and name the other shape as the case where the answer restores the anchor.

### 2. `Runtime/Commands/VoxrCommandParser.cs:2835-2841` — major, CONFIRMED
The canonical comment — the one three other sites now point at — defines the reachable shape as *"one member's first required element **is the discriminator**"*, then credits all three new tests to it. True of one fixture only. For `fire_to ["{?ship}","fire","to","{target}","now"]` the first required element is `"fire"` (index 1) while the discriminator is index 2, so in the fixture used by **two of the three** tests *neither* member is anchored on the discriminator. `CreditsRequired` (`:1244`) and the 1-based warning index (`:2253` vs the tests' `element 3`) confirm it. The block also contradicts itself fifteen lines later, calling that trait *"the shape where the discriminator IS the barred member's own anchor"* — a qualifier that would be vacuous if the opening were the definition.
**Fix:** use the property `KNOWN_LIMITATIONS.md` and the test header already use — members *disagree about which element is first required* — with anchor-is-discriminator named as one instance.

### 3. Registration order asserted away — major, CONFIRMED
The path additionally requires the **unbarred member to win the tie**. `Tied` never displaces the incumbent (only `Better` writes the winner, `VoxrCommandParser.cs:2475-2485`), so the first-registered member wins; if that is the barred one, the bar at `:2651` skips the result and `_tiedSiblingBuf` write entirely and nothing is offered — pinned on `main` by `VoxrCommandParserTests.cs:5233` `MixedAnchorSet_BarredMemberRegisteredFirst_NothingFires`. Sites:
- `VoxrCommandParser.cs:2838` — *"registration order hands the round to the member the bar does not touch"*, unhedged, inside a method that runs in **both** orders — and contradicting `:1685` in the same file (*"**one order** hands it…"`).
- `CHANGELOG.md:18` — *"the round **is handed** to a member the bar does not touch"*, where the untouched bullet above hedges *"**can** hand"*.
- `KNOWN_LIMITATIONS.md:504-513` — Root cause offers mixed anchoring as the only precondition; the repro fixes an order without saying the order is load-bearing.
- `Documentation~/command-recognition.md:491`, `api/command-recogniser.md:27` — **partially qualified**; only *"the round resolves as usual"* is flat.

### 4. `VoxrCommandRecogniserInjectionTests.cs:2249-2300` — minor, CONFIRMED
`Disambiguation_MixedAnchorSet_TheUnbarredWinnerIsEquallyUnheard` can pass vacuously. Its only assertion runs *after* `Answer("at")`, with no pending or fired-count check before it. In the counterfactual its own comment claims to pin (`TryBuildAmbiguity` returning false), `fire_at` would fire straight off the flush — it already clears every filter, since the pending opening today proves it — `firedIntent` would already read `"fire_at"`, `Answer("at")` would parse to nothing, and the pass would be silent (no subscriber on `OnUnrecognisedSpeech`, no parse-path logging, same construction warning either way). *Severity lowered from major:* the sibling test at `:2206` asserts `PendingAmbiguity.HasValue` on the identical fixture, so a real regression cannot escape the suite.
**Fix:** one line before `Answer("at")` — `Assert.IsTrue(_recogniser.PendingAmbiguity.HasValue, …)`.

### 5. PR body, *Why* §1 — minor, CONFIRMED
*"**Its preconditions cannot occur.**"* All three of #126's stated preconditions co-occur in a fixture already on `main`: `VoxrCommandRecogniserInjectionTests.cs:1997-2027` sets the flag, registers two patterns differing at element 0, and drops it. What cannot occur is the **outcome** the issue predicted — which the next sentences state correctly, then concede with *"That case is already closed and tested."* A case that is tested is a case that occurs. Same wording in the commit message of `e06b720`.

### 6. PR body, *What* §4 — minor, CONFIRMED
*"every prior disambiguation fixture … uses an **interior** discriminator, so nothing in the suite could reach a leading-missed rival."* `Disambiguation_LeadingDiscriminator_AsksNothingAndFiresNothing` (`:1997`) configures **inline** with a leading discriminator and the flag on, and a leading-missed rival **is** recorded there — `RecordTiedSiblingRival` is called at `:2601`, inside the scan loop, while the bar sits at `:2651` *after* it. It is simply never offered, because the round-level record at `:2692` is skipped. The supportable claim is the narrower one the test block's own header makes: nothing previously *parsed a mixed-anchor set* with the flag on.

### 7. `KNOWN_LIMITATIONS.md:529` — minor, CONFIRMED
*"see *Sibling patterns that differ at one word fire the first-registered one* **above**"* — that heading is at `:649`, 120 lines **below**. The file has zero intra-file anchor links (every in-file reference is positional prose), so the direction word carries the whole navigation burden. The entry's *other* "above" (`:494`) is correct. One-word fix: `below`.

---

### Refuted

- **CHANGELOG condition keys on the complement of the uniform exception, sweeping in ordinary interior ties where nothing is barred** — REFUTED. The scoping clause is inside the same sentence (*"and a member can therefore lose its own anchor while its sibling keeps one"*), before the definite description it licenses, and the bullet restates that scope twice more downstream.

### Notes (unverified, not numbered)

- Test 3 restates test 2's 30-line fixture verbatim instead of sharing a helper in the style of `ConfigureAsking`; drift risk, not correctness.
- The `KNOWN_LIMITATIONS` workaround list omits registering the barred-prone member first, which suppresses the path outright — though it also suppresses the legitimate sibling, so omitting it is editorially defensible.

- **The test-count claims were not independently re-run.** `EditMode 133/133, PlayMode 583/583` comes from a separate harness run against the host Unity project; review agents were instructed not to launch Unity. Statically confirmed: exactly three `[Test]` methods added, consistent with 580 + 3.
- On-device (Quest) behaviour — not applicable; no native, audio, or lifecycle code is touched, and the runtime diff is comment-only (verified mechanically: every added line under `Runtime/` begins with `//`).

Per the issue lane's rule, nothing below blocker has been fixed — all seven stand unruled for the human. None was rated a blocker.

jinwoo1601:
Pushed `98009ef` addressing **all seven confirmed findings** from the review above.

| # | finding | fix |
|---|---|---|
| 1 | `KNOWN_LIMITATIONS` rationale false of the anchored fixture (and same universal in `CHANGELOG`) | scoped to the non-anchored shape; the anchored shape now stated as the case where gating would be **worse**, since answering supplies the anchor |
| 2 | canonical `RecordTiedSiblingRival` comment defined the shape by a trait only one of two fixtures has | redefined as *members disagree about which element is first required*, with anchor-is-discriminator named as one instance |
| 3 | registration order asserted away at four sites | the unbarred member must **win** the round — now stated in the comment, `CHANGELOG`, `KNOWN_LIMITATIONS` root cause and repro; the two doc clauses conditioned; added as a workaround with its cost |
| 4 | `…TheUnbarredWinnerIsEquallyUnheard` could pass vacuously | asserts `PendingAmbiguity.HasValue` **before** the answer |
| 5 | PR body: "its preconditions cannot occur" | corrected in place — they occur; the outcome does not |
| 6 | PR body: "every prior disambiguation fixture … interior discriminator" | narrowed to: no prior fixture *parsed* a mixed-anchor set with the flag on |
| 7 | `KNOWN_LIMITATIONS` cross-reference "above" → target is 120 lines below | `below` |

The refuted finding was left alone.

Re-verified after the fixes: **EditMode 133/133, PlayMode 583/583**, Unity 6000.4.7f1, result-XML `start-time` checked fresh against wall clock (the totals are unchanged from the previous run, so the timestamp is the only thing that distinguishes it). `Runtime/` is still comment-only across the whole branch, verified mechanically.

One correction that could not be made in place: the message of `e06b720` repeats *"Its stated preconditions cannot occur."* `98009ef` records the correction rather than rewriting history — a **squash merge** takes the (corrected) PR body, not that message.

PR #137, opened 2026-08-28
Pin the HoldExtendable -> None transition the leading-miss bar introduces (#127)

### LINKED ISSUE #127: Leading-miss bar suppresses the shortened eager hold, lengthening the wait before OnUnrecognisedSpeech

jinwoo1601:
Closes #127

## What

Test-only, plus one comment. **No runtime behaviour changes.**

Two pin tests in `Tests~/Runtime/VoxrEagerCommitTests.cs`, both PlayMode, one per `HoldExtendable` return in `TryEagerCommit`:

- `TryEagerCommit_LeadingRequiredMiss_ExtendablePattern_StillReturnsNone` — guards hold return B (`!CanCommitEarly(...)`, issue #32)
- `TryEagerCommit_LeadingRequiredMiss_UnanalysableGrammar_StillReturnsNone` — guards hold return A (`_canCommitEarly == null`, issue #44)

Each pairs the barred assertion with a **control on the same grammar** whose leading element is present. The control rules out *"this grammar can never hold at all"* — the case where a bar refusing everything would pass the barred assertion. It does **not** by itself establish that the barred utterance reached a hold, since that is a different input and the whole-buffer span gate between the bar and the hold returns is input-dependent. What establishes the transition is the **mutation** recorded below, and each test's comment now carries that mutation and its observed values, so the evidence lives beside the assertion rather than only here.

One comment paragraph (two sentences) appended to the sibling-tie refusal's comment in `Runtime/Commands/VoxrCommandParser.cs`, noting that its principle — *"a refusal must not lengthen a wait it is not responsible for"* — is not absolute, that the leading-miss bar ~60 lines above is a ruled exception, and why the two cases differ.

## Why

#127 reports that the bar sits above both hold returns, so a barred winner that would have been *held* is refused outright, stretching the wait before `OnUnrecognisedSpeech` from `prefixHoldSeconds` to the full `bufferWindow`.

**That behaviour was already ruled and is deliberately left alone.** Option 1 — accept permanently — was ruled at item 3 of the #124 backlog (`Planning~/features/leading-miss-warning/requirements.md:125`, and recorded in #130). Options 2 and 3 are runtime timing changes needing a design amendment, and are not in scope here.

What was left unbuilt is the issue's **"Also unpinned"** section, which the same ruling explicitly kept open: *"#127's 'also unpinned' note — that no test covers the `HoldExtendable -> None` transition — stays open with the issue."*

The transition is also a **published** guarantee: the shipped `[Unreleased]` CHANGELOG entry for #124 states *"the eager gate refuses a leading-missed candidate before it can arm the shortened hold, so such an utterance now waits the full `bufferWindow`"*. Nothing tested it. This is the same family as #128's three published-but-unguarded guarantees.

Two tests rather than one because the two hold returns are separate lines and each test guards only one — which takes **two** mutations to demonstrate, since the between-returns relocation alone reddens the un-analysable test too.

## Validation

- [x] **Unity EditMode — 133/133 passed**, `failed="0"` (host project `VoXR TestGround`, Unity 6000.4.7f1, package referenced by local path)
- [x] **Unity PlayMode — 585/585 passed**, `failed="0"`; both new tests confirmed present by name in the results XML and `result="Passed"` (not merely absent from the failure list)
- [x] **Mutation-verified — the pins can be made red, and each guards a distinct return.** Moving the bar below *both* hold returns failed *both* tests, each reporting `Expected: None / But was: HoldExtendable`, with no other test in the 585 affected. Moving it *between* the two returns failed only the un-analysable-grammar test and left the extendable-pattern test passing.
- [ ] On-device (Quest) check: **not applicable.** The only `Runtime/` change is a comment; no parsing, scoring, firing, lifecycle or native-bridge behaviour moves.

Two limits worth stating rather than glossing:

- NUnit aborts a test method at its first failing assertion, so under the both-returns mutation each test stopped at its barred assertion and its *control* assertion did not execute in that run. The controls are observed green in the unmutated suite and under the between-returns mutation; they are not separately observed under the first mutation.
- The narrowing mutation now cited in the second test's comment — `bestLeadingRequiredMissed && _canCommitEarly == null`, which would spare that test and redden the extendable one — is **derived, not run**. The two mutations that were actually executed are the two reported above.

No CHANGELOG entry — the behaviour is already published there under #124 and is unchanged by this PR.

---

*Updated after the `review-pr` pass: four findings, all confirmed, all of them prose claims wider than their evidence rather than defects in the tests. Fixes in `acce4fa`.*

### DIFF
--- a/Runtime/Commands/VoxrCommandParser.cs
+++ b/Runtime/Commands/VoxrCommandParser.cs
@@ -4462,6 +4462,13 @@ namespace VoXR.Commands
             // commit early anyway. There is nothing to refuse when nothing was being offered,
             // and a refusal must not lengthen a wait it is not responsible for. Here the only
             // transition it introduces is Commit -> None.
+            //
+            // That principle is not absolute, though: the leading-required-miss bar above does
+            // convert HoldExtendable into None, and that is a ruled exception (issue #127,
+            // accepted permanently). A barred candidate produces no result at any length of
+            // wait, so the only cost there is how soon the speaker is told — whereas a tied
+            // candidate may still fire after the wait, which is what makes lengthening it
+            // pointless here.
             if (bestTiedSiblingCommandIdx >= 0)
                 return EagerCommitVerdict.None;
 
--- a/Tests~/Runtime/VoxrEagerCommitTests.cs
+++ b/Tests~/Runtime/VoxrEagerCommitTests.cs
@@ -969,6 +969,105 @@ namespace VoXR.Tests.Runtime
             );
         }
 
+        [Test]
+        public void TryEagerCommit_LeadingRequiredMiss_ExtendablePattern_StillReturnsNone()
+        {
+            // Pins the HoldExtendable -> None transition the bar introduces, at the second of
+            // the two hold returns: the extendable-pattern one (issue #32). The bar sits ABOVE
+            // both hold returns, so a barred winner that would otherwise have been HELD is
+            // refused outright instead — which lengthens the wait before OnUnrecognisedSpeech
+            // from prefixHoldSeconds to the full bufferWindow.
+            //
+            // That transition is deliberate and ruled (issue #127, accepted permanently). DR-5
+            // places the bar beside the #66/#70 completeness refusals — they are all refusals
+            // about the candidate itself — and moving it below the hold returns to preserve the
+            // shortened hold would be a design change, not a bug fix. The cost is latency only:
+            // a barred candidate produces no result at any length of wait, so the only thing the
+            // longer wait delays is how soon the speaker is told. At the shipped default
+            // prefixHoldSeconds = 0 there is no shortened hold to lose, so it is inert there.
+            //
+            // The fixture: "advance"'s only pattern is a proper prefix of "advance_more"'s, so
+            // CanCommitEarly reports false for it and the winner reaches the extendable hold
+            // return. Everything between the bar and that return is cleared deliberately —
+            // score (1 + 1) / 3 = 0.667 clears the 0.6 gate; there are no slots at all, so the
+            // issue #66 condition cannot bite; the miss is LEADING, so "bravo" and "charlie"
+            // match after it and no required TAIL remains (issue #70); and the match spans the
+            // whole buffer. "advance_more" cannot steal the win on the barred utterance — its
+            // unspoken "delta" drags it to (1 + 1) / 4 = 0.5, under the gate and under 0.667.
+            var parser = new VoxrCommandParser(
+                Slots(),
+                Commands(
+                    Cmd("advance", P("alpha", "bravo", "charlie")),
+                    Cmd("advance_more", P("alpha", "bravo", "charlie", "delta"))
+                )
+            );
+
+            Assert.AreEqual(
+                EagerCommitVerdict.None,
+                parser.TryEagerCommit(Tok("bravo charlie"), null, 0.6f, 0.4f),
+                "the bar outranks the extendable hold — a barred winner is refused, not held"
+            );
+
+            // Control, on the same grammar, and the whole point of the test: with the leading
+            // element present the very same winning pattern reaches the extendable hold return
+            // and answers HoldExtendable. Without this the assertion above could go green from
+            // any refusal at all, and would not show that the fixture reaches a HOLD — i.e.
+            // that the bar really is converting HoldExtendable into None, rather than sitting
+            // above a None that would have happened anyway.
+            Assert.AreEqual(
+                EagerCommitVerdict.HoldExtendable,
+                parser.TryEagerCommit(Tok("alpha bravo charlie"), null, 0.6f, 0.4f),
+                "unbarred, the same winner is held because its pattern is a prefix of a longer one"
+            );
+        }
+
+        [Test]
+        public void TryEagerCommit_LeadingRequiredMiss_UnanalysableGrammar_StillReturnsNone()
+        {
+            // The same HoldExtendable -> None transition, at the FIRST hold return: the
+            // un-analysable-grammar degrade (issue #44). Both tests are needed because the two
+            // hold returns are separate lines — an edit that moved the bar to sit BETWEEN them
+            // would leave one of these green and flip the other, so neither one pins the
+            // transition on its own.
+            //
+            // Deliberate and ruled exactly as in the test above (issue #127, accepted
+            // permanently): DR-5 places the bar beside the #66/#70 completeness refusals, and
+            // moving it below the hold returns would be a design change, not a bug fix. The cost
+            // is latency only — a barred candidate produces no result however long the wait, so
+            // only the moment the speaker is told moves — and it is inert at the shipped default
+            // prefixHoldSeconds = 0.
+            //
+            // OverLimitCommand carries one optional past MaxOptionalExpansion, so the
+            // eligibility analysis is abandoned for the whole parser and _canCommitEarly is
+            // null; that is what puts hold return A in play here rather than B. It matches
+            // nothing in either utterance ("noisy" is never spoken), so "advance" still wins.
+            // The construction warns at authoring time (issue #44), hence the expectation
+            // before it and the sweep at the end.
+            LogAssert.Expect(LogType.Warning, new Regex("more than the 12"));
+
+            var parser = new VoxrCommandParser(
+                Slots(),
+                Commands(OverLimitCommand(), Cmd("advance", P("alpha", "bravo", "charlie")))
+            );
+
+            Assert.AreEqual(
+                EagerCommitVerdict.None,
+                parser.TryEagerCommit(Tok("bravo charlie"), null, 0.6f, 0.4f),
+                "the bar outranks the un-analysable degrade — a barred winner is refused, not held"
+            );
+
+            // Control: the same grammar with the leading element present cannot commit either,
+            // because the analysis that would have vetted a commit was abandoned — so it
+            // degrades to the hold instead of None. That is what makes the barred utterance's
+            // None attributable to the bar, and not to the missing analysis.
+            Assert.AreEqual(
+                EagerCommitVerdict.HoldExtendable,
+                parser.TryEagerCommit(Tok("alpha bravo charlie"), null, 0.6f, 0.4f),
+                "unbarred, the un-analysable grammar degrades to the hold rather than None"
+            );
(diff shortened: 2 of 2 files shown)


jinwoo1601 (commented):
## `review-pr` — slim profile

**Head reviewed:** `3599f26342cb78c120bb82335972cc8c60eaa415` · **Base:** merge-base `7df16e2` · **Angles:** `CLAIM`, `LINE`, `GONE` (issue-lane slim profile; none dropped, none added — the diff crosses no contract boundary and touches no MonoBehaviour lifecycle, so `XFILE` and `PIT` were not warranted).

**No blockers.** The two pins are sound and non-vacuous — `LINE` independently re-derived all four assertions from the parser source and established that with the bar deleted, both barred assertions reach a `HoldExtendable` return and go red. `GONE` confirmed the diff is pure addition by reconstructing the base from the head (deleting only the inserted line ranges yields a byte-identical file), and that the gate order in `TryEagerCommit` is unchanged.

All four findings below are defects in **prose** — the PR body and the new test comments — not in the tests' behaviour. All four were adversarially verified; none were refuted.

---

### 1. The control assertion is necessary but not sufficient, and both the PR body and the merged comment claim it is sufficient — `minor`, CONFIRMED

The body says the control is "what shows the fixture actually reaches a hold — i.e. that the bar converts `HoldExtendable` into `None`, rather than sitting above a `None` that would have happened anyway", and `Tests~/Runtime/VoxrEagerCommitTests.cs:1013-1016` repeats it as "the whole point of the test".

The control is a **different input** (`Tok("alpha bravo charlie")` vs `Tok("bravo charlie")`), and the whole-buffer span gate at `Runtime/Commands/VoxrCommandParser.cs:4416` is input-dependent. The counterexample exists in this very file: `VoxrEagerCommitTests.cs:819-830` pairs a span-gate `None` with a grammar that holds — both green with the bar deleted, pinning nothing.

The control rules out "this grammar can never hold at all", which is exactly how the #125 pin words its own control (`:963-964`). What actually establishes the transition is the mutation. **Structural consequence worth noting: that mutation evidence is durable in this PR body and *not* in the merged file** — so the file's most emphatic claim points at the one thing that cannot support it.

### 2. The gate enumeration mis-locates three of four gates and omits the confidence gate — `nit`, CONFIRMED

`VoxrEagerCommitTests.cs:991` frames a list as "Everything between the bar and that return is cleared deliberately", then enumerates score, #66, #70 and whole-buffer. Actual order: minScore `4281`, #66 `4337`, #70 `4364`, **bar `4395`**, whole-buffer `4416`, confidence `4420`, hold A `4432`, hold B `4437`.

The list is a *hybrid* — three gates above the bar (so the barred assertion is refused by the bar, not earlier) plus one below it (so the control reaches hold B) — described as one span. The below-the-bar half omits the confidence gate at `:4420`, which this fixture clears only because `wordConfidence` is `null` and `ComputeConfidence` returns `-1f` (`:4002-4003`). The sibling pin at `:1476,1483` already lists that gate explicitly, so the file sets the standard this misses.

### 3. "Neither one pins the transition on its own" is true but its cited mutation demonstrates only half of it — `nit`, CONFIRMED (as under-justification, **not** as a false claim)

`VoxrEagerCommitTests.cs:1027-1031`. Verified: the conclusion is **true** — the fixtures cover disjoint input classes, and an edit exists that spares the un-analysable test while breaking the extendable one. But the between-returns mutation it cites reddens the un-analysable test too, so it never exercises a break that test misses. The `so` is carried by "the two hold returns are separate lines", not by the illustration.

Caution for whoever fixes this: the reviewing agent's suggested example (`_canCommitEarly != null`) is **inverted** and would not demonstrate the missing half — the direction needed is `== null`. Because hold A precedes B and keys off a whole-parser flag, most single-site relocations redden the un-analysable test first; the mutation class that spares it is narrow.

### 4. "One sentence appended" describes a two-sentence addition — `nit`, CONFIRMED

`Runtime/Commands/VoxrCommandParser.cs:4465-4471` is two independent declaratives; the body describes only the first. "One comment paragraph" would be exact. No consequence beyond precision — the undescribed sentence's assertion ("a barred candidate produces no result at any length of wait") was checked against the flush path at `:2651` and holds.

---

### Notes (unverified, not findings)

- The appended comment at `VoxrCommandParser.cs:4466-4471` largely restates the pre-existing `ACCEPTED, and unlike the sibling condition below…` block at `:4384-4394`. A simplify pass may want one of the two.
- `..._ExtendablePattern_...` omits the `LogAssert.NoUnexpectedReceived()` sweep its sibling test uses.
- Whether two tests earn their place, given hold A precedes B and reddens first under most mutations, is a scope question this review did not rule.

- **The review agents could not execute the suite.** This is a bare UPM package; `Tests~` compiles only via the host project. The `133/133`, `585/585` and mutation results in the PR body come from the authoring session's runs, not from this review — the angles checked them by reading and found no counterexample, which is weaker than execution.
- On-device (Quest) behaviour — not applicable here, and the diff's only `Runtime/` change is a comment.

*Per the issue lane, nothing below blocker is fixed by the automated pass. All four findings are left for the human to rule on.*

jinwoo1601:
## All four review findings addressed — `acce4fa`

Ruled by the human: address all four, not just blockers. None were blockers; all four were prose claims wider than their evidence, in text I wrote. Comment-only commit — every changed line begins with `//`, and the non-comment lines are byte-identical to `3599f26`.

| # | finding | fix |
|---|---|---|
| 1 | control claimed as sufficient evidence for the transition | Both control comments now state what the control *does* establish ("this grammar can never hold at all", as the #125 pin words it) and what it does not, cite `TryEagerCommit_LeadingRecognisedLeftover_StillReturnsNone:815` as the span-gate shape that proves the point, and **record the mutation and its observed `Expected: None / But was: HoldExtendable` in the test file itself.** The PR body's version of the same claim is corrected too. |
| 2 | gate list mis-framed as one span, confidence gate omitted | The list is now explicitly two-sided — gates above the bar (so the bar is the first refusal) and below it (so the control reaches the hold) — and names the confidence gate, inert here because both calls pass a null `wordConfidence` and `ComputeConfidence` returns `-1f`. |
| 3 | "neither pins it on its own" under-justified by one mutation | Now cites **both** mutations: the between-returns relocation (reddens the un-analysable test, spares the extendable one) and a narrowing to `bestLeadingRequiredMissed && _canCommitEarly == null` (the reverse). The verifier caught that the reviewing agent's suggested direction was inverted; the version committed is the correct one, and the PR body flags it as derived rather than run. |
| 4 | "one sentence appended" was two | PR body now says "one comment paragraph (two sentences)". |

**Finding 1 is the one that mattered**, and its fix is structural rather than editorial: the mutation evidence existed only in this PR body, where no future reader of the test would find it. It now sits beside the assertion it supports.

### Re-verified at `acce4fa`

- **PlayMode 585/585**, `failed="0"`, no `error CS` in the Unity log, both pin tests present and passing, all 71 tests in `VoxrEagerCommitTests` green.
- EditMode not re-run — the change is comments inside two PlayMode test methods. Its 133/133 at `3599f26` stands.
- The mutation runs were **not** repeated at `acce4fa`; they were measured at `3599f26`, whose non-comment lines are byte-identical to this commit.

Still open and unruled: the three `### Notes` from the review (the appended parser comment overlapping the pre-existing `ACCEPTED` block at `VoxrCommandParser.cs:4384-4394`; the missing `LogAssert.NoUnexpectedReceived()` sweep in the extendable test; whether two tests earn their place). Those were explicitly not findings, so I left them alone.

PR #138, opened 2026-08-28
Suppress OnUnrecognisedSpeech when a partial-match or confirmation pending opens (#133)

### LINKED ISSUE #133: OnUnrecognisedSpeech fires in the same frame a partial-match or confirmation pending opens

jinwoo1601:
Closes #133

## What

One of the three Step 7 branches that enter a pending and `continue` set `anyThresholdFiltered = true` — the disambiguation branch; the partial-match and requires-confirmation branches did not. They now do, so all three set it.

- `Runtime/Commands/VoxrCommandRecogniser.cs` — the flag is set in the partial-match branch (inside the `unfilled.Length > 0` check, so a candidate that opens *no* pending still falls through to the reject below and is still reported) and in the requires-confirmation branch.
- `Tests~/Runtime/VoxrPendingCommandTests.cs` — two pins, one per branch, each asserting `OnCommandPending` fired **and** `OnUnrecognisedSpeech` did not.
- `Documentation~/` — three files. The shipped docs described the defect as intended behaviour, so they move with the fix (see below).
- `CHANGELOG.md` — a new `### Fixed` block under `[Unreleased]`.

## Why

At the tail of the result loop, `acceptedCount == 0` with the flag clear raises `OnUnrecognisedSpeech`. Both branches `continue` without incrementing `acceptedCount`, so when such a pending was the only outcome of an utterance the integrator was told the speech was not understood in the same frame `OnCommandPending` asked it to prompt the speaker. The disambiguation branch already set the flag, with a comment naming precisely that harm.

`allowPartialMatch` and `requiresConfirmation` are both opt-in and neither is reached by the default grammar, which is why this survived.

### The docs asserted the old behaviour

This is not incidental scope. `Documentation~/scoring.md`s `OnUnrecognisedSpeech` table literally read `| A candidate was diverted to pending (partial match or requiresConfirmation) | **fires**, alongside OnCommandPending |`, and `command-recognition.md` listed the diversion as situation 4 of six. Leaving those in place would have shipped documentation contradicting the code. The first sweep for this was a grep for `OnUnrecognisedSpeech`, and the review caught that being insufficient: a doc can be falsified by a path it never names. `command-recognition.md:200` and `:676` describe *dynamic slot exclusion* without mentioning `allowPartialMatch` at all, yet an excluded value **is** an unfilled required slot, so it routes into the branch this PR changed. Those two, and two more in `scoring.md` (`:240`, `:282`) whose unconditional wording this change adds counterexamples to, are fixed in the follow-up commits rather than claimed correct.

## Validation

Both suites run through the host project `VoXR TestGround` (Unity 6000.4.7f1), EditMode and PlayMode, per the project verification bindings.

- [x] **Red before green — the pins were proven to catch the defect.** The two tests were written and run *before* the production fix was applied. Both failed, `Expected: 0 / But was: 1`, with the rest of the run green (EditMode 133/133, PlayMode 585/587). They are not vacuous pins.
- [x] **Green after the fix:** EditMode 133/133, PlayMode 587/587. Compiles clean, no `error CS` in either log.
- [x] **No regression:** total test count unchanged at 133 + 587 across both runs, and only the two target tests changed state (fail → pass). No previously-passing test asserted the old behaviour.

No on-device (Quest) check is claimed or, as far as I can tell, needed: the change is two assignments inside the pure command-parsing result loop, with no native-bridge, MonoBehaviour-lifecycle, mic, or audio-capture surface touched.


jinwoo1601 (commented):
## Independent review — PR #138

**Depth profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE` — **with `XFILE` added**, because the diff edits mutually cross-referencing statements across three documentation files and the code those files describe. No angle was dropped.
**Head reviewed:** `e7923feeb25992eb19fb0b3fb18609c6329b0216` · **merge-base:** `41d5b45`
Four finder agents ran against the diff with no context from the session that wrote it; the PR body was treated as a claim to audit. Each numbered finding below was then passed to a separate verifier prompted to refute it. The code fix itself verified clean on every angle — all findings below are about documentation and about this PR's own claims.

---

### 1. `command-recognition.md:200` and `:676` are now false, and the PR body's doc-sweep claim is overstated — CONFIRMED, major

`:676` (untouched): *"When a value provider excludes a target, speech that names the excluded target fires `OnUnrecognisedSpeech` instead of `OnCommandRecognised`. You can use this to explain why the command failed"* — followed by a worked handler. `:200` says the same.

An excluded slot value leaves the slot unfilled, so with `allowPartialMatch` on it now routes into the fixed branch and is suppressed. The verifier staged the head parser into the repo's A/B rig and got a candidate **bit-identical at the branch input** to this PR's own new pin:

```
EXCLUDED: 'launch missiles target hotel two'  n=1  launch_weapon=0.5000 [weapon=missiles] hasTarget=False
control   'launch missiles target'            n=1  launch_weapon=0.5000 [weapon=missiles] hasTarget=False
```

So `PartialMatchPending_DoesNotAlsoReportUnrecognised` asserting `unrecognisedCount == 0` is itself the proof these passages are false. True before this diff, false after. The same file qualifies carefully at `:643` ("and the command does not set `allowPartialMatch`"), so the unconditional form is an inconsistency, not shorthand.

The body's *"Every passage … was checked by grep; the ones describing the old behaviour are updated"* does not hold: 17 mentions in `Documentation~/`, 3 touched, and two of the 14 untouched describe behaviour this diff invalidated.

**Addressed** — see the follow-up commit.

### 2. `scoring.md:240` — named in the PR body as "remains correct", but this diff adds two counterexamples — CONFIRMED, minor

*"…and where nothing else in the utterance fired, the utterance reports through `OnUnrecognisedSpeech`."* Reachable trace: grammar `approach_target: ["approach","target","{target}"]` + `cease_fire: ["cease","fire"]` with `RequiresConfirmation`, utterance *"target hotel one cease fire"* — round 1's winner is barred and consumes the head (`VoxrCommandParser.cs:2651` re-bases `searchStart`), round 2 emits `cease_fire`, which enters a confirmation pending. Nothing fired, and it is now silent. Base fired.

**Pre-existing sentence, new counterexamples.** It was already false for disambiguation, `minConfidence` and debounce; this diff moves two more configurations onto the false side. Left unfixed deliberately: qualifying a sentence that was already an approximation is a judgement about how much to say, not a correction this run should make unilaterally. The verifier notes `scoring.md:282` sits on identical footing.

### 3. PR body's "## What" opening sentence is inverted — CONFIRMED, minor

*"Two of the three Step 7 branches that enter a pending and `continue` set `anyThresholdFiltered = true`; the partial-match and requires-confirmation branches did not."* — two of three set it, then two are named that did not: four branches out of three. Ground truth: **1 of 3 before** (disambiguation, `:1012` at base), **3 of 3 after** (`:946`, `:1022`, `:1059`). "Two" is true in neither tense. The commit message and the test comment at `VoxrPendingCommandTests.cs:1830` both state the correct polarity ("*left* `anyThresholdFiltered` clear"); only the PR body inverted it. Correction for the record, since the body cannot be edited from here: **one** branch set the flag before this change; three do now.

---

### Refuted

- **`api/command-recogniser.md:44` — "those three are silent" reads as scoping the silence to pendings only.** REFUTED: the following parenthetical says "The pending **case** is deliberate" (singular), collapsing the pending kinds to one item, so "those three" = {`minConfidence`, debounce, pending}. Both readings are true of the code; the wrong one is incomplete, never false.

- `VoxrCommandRecogniser.cs:1088-1090`: the tail comment still enumerates only "(confidence, debounce)" as the flag's setters; 3 of 5 are now unlisted. Pre-existing gap this diff widens.
- `troubleshooting.md:47` and `:121` recommend `OnUnrecognisedSpeech` as the raw-transcript diagnostic; that advice is narrower for the two opt-in flags now.
- `command-recognition.md:643` (situation 3) frames the reject as "does not set `allowPartialMatch`", omitting the `unfilled.Length == 0` reject.
- `scoring.md:555`'s remedy pair ("set `allowPartialMatch` … **or** re-prompt off `OnUnrecognisedSpeech`") is now mutually exclusive rather than complementary; the "or" keeps it literally true.
- `PendingCommandHandler.cs:280` comment claims "EnterPending refuses anything at or below zero"; it contains no such refusal. Pre-existing, untouched.
- A multi-result utterance where one candidate opens a pending now also silences a genuinely-unmatched sibling candidate. `LINE` judged this the intended per-utterance semantics — byte-identical to what the confidence, debounce and disambiguation branches have always done — not a defect.

### Verified clean

`LINE` returned zero findings: all four no-pending escape paths (`Score <= 0f`, `TryLookupCommand` false, `AllowPartialMatch` false, `unfilled.Length == 0`) still reach the reject with the flag clear, so the fix does not over-suppress; `EnterPending` has no failure return, so "flag set" and "`OnCommandPending` fired" are equivalent at both sites; the two pins are mutually independent — each fails if only its own branch is reverted. `XFILE` confirmed every markdown anchor resolves, no heading changed, all three doc files agree on five-fires/three-silences, and no fourth place in the repo asserts the old behaviour.

- **The test runs.** This is a bare UPM package with no test CI (`.github/workflows/` holds only `release.yml`); the suites run only through the host Unity project. The claimed counts (133 EditMode, 585→587 PlayMode) are **internally consistent** — the diff adds exactly two `[Test]` methods, both under `Tests~/Runtime/` i.e. PlayMode — but were not independently re-run by this review.
- **On-device (Quest) behaviour.** Human-only. The body's argument for not needing it was audited and found supported: both assignments sit in the result loop, not in a lifecycle callback, and no native/plugin file is in the diff.

jinwoo1601:
Pushed `309b440` addressing **finding 1** — `command-recognition.md:200` and `:676` now carry the `allowPartialMatch` qualifier the same file already uses at `:643`. Findings 2 and 3 are left unfixed for your ruling; note that while writing the fix I caught myself asserting the unfilled slot name is readable from `PendingCommand` — it is not (`UnfilledSlots` is `internal`), so the paragraph directs `HasSlot("target")` instead. Docs-only change; no test re-run.

jinwoo1601:
Both remaining findings addressed.

- **Finding 2** — `b27a33c`: `scoring.md:240` and `:282` now name the utterance-wide suppression instead of asserting the report unconditionally. Both were already approximations before this PR (false for a co-occurring disambiguation pending, `minConfidence` rejection, or debounce); this change added two more counterexamples, so they are corrected rather than merely re-scoped.
- **Finding 3** — the PR body is corrected in place: **one** branch set the flag before this change (disambiguation), three do now. The doc-sweep sentence in the same section was also stale after the finding-1 fix and now records what the review actually established — that a grep for the changed symbol cannot find a doc falsified by a path it never names.

Docs and PR-body only since `309b440`; no code change, so no test re-run. The suites were last green at `e7923fe` (EditMode 133/133, PlayMode 587/587).

PR #139, opened 2026-08-28
Withhold the cancel-collision report on an every-member anchor (#132)

### LINKED ISSUE #132: Cancel-collision report is a false advisory where the tie can never be asked about

jinwoo1601:
Addresses #132 — **one of its two shapes**. Deliberately does *not* close it; see *Scope* below.

## What

`WarnOnSiblingDiscriminator`'s per-member cancel-collision loop now skips a set where `DiscriminatorIsEveryMembersAnchor(set)` holds — the same set-level predicate the sibling-tie filter beside it already consults, reused rather than copied.

- `Runtime/Commands/VoxrCommandParser.cs` — the guard, placed *after* the cancel-vocabulary and `HasAnswerableCoMember` tests so the walk over members and their patterns runs only for a member that would otherwise be reported.
- `Tests~/Runtime/VoxrCommandParserTests.cs` — one new PlayMode test.
- `Documentation~/command-recognition.md` — the collision paragraph listed one reachability narrowing; it now lists two, and says applying remedy 2 silences this warning alongside the sibling one.
- `CHANGELOG.md` — new bullet under `[Unreleased]` → `Fixed`.

## Why

The report tells an author a discriminating value is also cancel vocabulary, so answering with that word would cancel rather than choose — and its remedy is to rename a grammar literal. Where that word leads **every** member pattern, the leading-required-miss bar (#124) refuses whichever member wins the round: nothing fires, no pending opens for that set, and no question is ever posed for cancel to swallow. The author was being sent to rename a literal to protect an answer that can never be given — the false-advisory class #81 spent a feature reversing, and the standard the report's own comment holds itself to.

The sibling warning already withholds on this shape; the collision report did not inherit it because it is narrowed per **pair** where the sibling filter is narrowed per **set**. Inheriting the anchor condition alone is sound set-wide: if no question is asked for the set, no pair inside it can be asked about either.

## Scope — why this does not close #132

#132 names two unreachable shapes. This PR fixes the issue's **shape 2** (leading discriminator). The one it leaves alone is the issue's **shape 1** — a set that cannot clear the default `minScore` — and the guard for it is **deliberately not inherited**. That test is judged against `DefaultMinScore` because the parser is built from the grammar alone, while the runtime gate *is* handed the configured threshold — the asymmetry `KNOWN_LIMITATIONS.md` records under *"The sibling warning is silent below the default minScore"*. An author who lowered `minScore` can still have such a tie routed to the speaker, so inheriting the score gate would suppress an advisory that is **true** for exactly that author. #132 itself flags that half as needing a ruling; it remains open for it.

## Validation

- [x] **Green run** — host project `VoXR TestGround` (Unity 6000.4.7f1), **both** platforms: EditMode **133/133**, PlayMode **588/588**, zero failed/inconclusive/skipped. EditMode is unchanged at 133; PlayMode is `main`'s baseline of 587 plus this PR's one new test = 588. Every test that existed before still ran, and the only new one is this PR's.
- [x] **Mutation-verified, because a "no warning logged" test can pass vacuously.** With the guard commented out and nothing else changed, the suite was re-run: PlayMode went to 587/588 with **exactly one** failure — the new test — reporting `Unhandled log message: '[Warning] [VoxrCommandParser] Intent '"'"'answer_negative'"'"' carries the discriminating value "negative" at element 1, which is also in the default cancel vocabulary. …'`. EditMode stayed 133/133 and no other test moved. The guard is what withholds the message, and the test is load-bearing.
- [x] The new fixture uses **three** elements, not two, so the `(D − 1) / D` = `0.67` reading clears the default `minScore` and the anchor condition is the only thing withholding either message — a two-element grammar would have gone quiet for the wrong reason. `SiblingWarning_DiscriminatorCollidingWithCancel_IsAlsoReported` is the control: same intents, same two values, same collision, discriminator on the **last** element, and it still expects both warnings.
- [x] Regression neighbours confirmed still passing: `SiblingWarning_DiscriminatorCollidingWithCancel_IsAlsoReported`, `SiblingWarning_CancelCollisionReachableOnlyThroughOnePair_StillReports`, `SiblingSets_DemoGrammar_VolumeAndOrderAreStable` (the demo-grammar volume pin — no shipped-grammar warning changes hands).
- [x] Doc sweep for the **outcome** rather than the symbol: `Documentation~/`, `KNOWN_LIMITATIONS.md`, `README.md` and the released CHANGELOG blocks were searched for claims about when the collision report fires. `command-recognition.md:448` was the only prose the change falsified. Released `[1.5.0]` entries are historical and left as written.
- **No on-device (Quest) check applies.** The call site is `[Conditional("UNITY_EDITOR")]`, so a player build is unchanged; no `NativeBridge~/`, no MonoBehaviour lifecycle, no mic path is touched. Nothing about parsing, scoring, or which commands fire moves.


jinwoo1601 (commented):
## Independent review — PR #139

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE`. No angles dropped; none added — the diff crosses no contract boundary (`XFILE`) and touches no MonoBehaviour lifecycle, coroutine, or main-thread-only API (`PIT`).
**Head reviewed:** `0780fb09c497997d338b59424ea9a6f0b2b77709` · **merge-base:** `9bd465dd62cf262a7e028e21829c75c814e2366f`

Three findings, all CONFIRMED, all **minor** — every one is a claim written wider than the code, and none touches behaviour. **No blockers.** The guard itself, its soundness, and the test's non-vacuity were all attacked and held; `GONE` returned zero findings.

---

### 1 · The placement comment claims a guarantee the placement does not give · minor · CONFIRMED

`Runtime/Commands/VoxrCommandParser.cs:1753-1754` states the guard is *"Placed after both member tests so the walk over members and their patterns runs only for a member that would otherwise be reported."* The per-value dedup sits **below** the guard (`:1767-1783`), so a member that clears both member tests and is then discarded as a duplicate value has already paid for the walk.

Not hypothetical — `SiblingWarning_TwoPatternsCarryingOneCancelWord_ReportItOnce` (`Tests~/Runtime/VoxrCommandParserTests.cs:4821`) pins exactly that shape: `ident_hostile` and `ident_unknown` both carry `"negative"`, both are answerable against `ident_friendly`, and the second reaches `:1764` before being deduped at `:1782`.

The charitable reading — *"reported"* meaning *"reported by the two tests above"* — was tried and fails: `:1728` already uses *reported* in the emission sense (*"counts only members that would themselves be reported"*) twenty-six lines earlier.

**Fix:** reword to claim only what holds, or hoist the set-invariant predicate out of the member loop and say that instead.

### 2 · The PR body's "the second" indexes into #132's list and lands on the wrong shape · minor · CONFIRMED

The body says *"#132 names two unreachable shapes. **The second** is a set that cannot clear the default `minScore`."* The issue numbers them the other way: its **1** is *"Below the score gate (older)"*, its **2** is *"Leading discriminator (since #124)"*. The ordinal's antecedent is the clause that attributes the enumeration to #132, and the PR introduces no ordering of its own — so a reader resolving *"the second"* against the issue lands on the shape this PR **fixed**.

Chronology does not rescue it (the issue tags the minScore shape *"(older)"*). Kept minor because the appositive names the shape unambiguously in the same breath, and both merged artifacts avoid ordinals.

**Fix:** drop the borrowed ordinal — *"the shape that cannot clear the default `minScore` (#132's shape 1)"*.

### 3 · Validation box 1's stated reason is false — the totals do *not* match the baseline · minor · CONFIRMED

The box claims *"PlayMode **588/588** … **Totals match the pre-change baseline**, so nothing went undiscovered."* The baseline on `main` is **587**: PR #138, whose merge commit is this PR's merge-base, reports *"PlayMode 587/587"*. The diff adds exactly one PlayMode test with no `[TestCase]`/`[ValueSource]`, so 588 = 587 + 1.

The number is right and the assurance survives; only the justification is false. The reading that rescues it — *"baseline"* meaning this session's mutation run — fails, since that run postdates and contains the change. It misleads in the direction that matters at a merge gate: a reader reconciling 588 against `main` finds a mismatch and cannot tell whether a test was added or one silently vanished.

**Fix:** state the relation — *"PlayMode is the baseline 587 plus this PR's one new test = 588."*

---

### Refuted

None. All three numbered findings survived adversarial verification.

### Notes (unverified, not numbered)

- The new test asserts the *absence* of warnings but never pins that its grammar forms a sibling set at all; a future regression stopping a leading discriminator forming a set would leave it silently green. `Tests~/Runtime/VoxrCommandParserTests.cs:4962` shows the stronger in-test pattern (`Assert.AreEqual(1, VoxrCommandParser.FindSiblingSets(commands).Count, …)`). The adjacent `SiblingWarning_LeadingDiscriminatorInEveryMember_IsWithheld` omits it too, so this is a pre-existing house split rather than a new deviation.
- The set-invariant predicate is re-evaluated once per surviving member rather than once per set. No behavioural effect; cold path.
- This bullet lands under `[Unreleased] → Fixed` while its sibling-warning counterpart (`CHANGELOG.md:14`) sits under `Changed`, though both withhold a warning on the same predicate.
- `Documentation~/command-recognition.md:517` states the construction warning unqualified where `:448` now lists two narrowings. It was already loose under the same-intent narrowing shipped in 1.5.0, so this PR compounds rather than falsifies it.

### Verified clean (attacked, held)

- **Suppression is exact, not over-wide.** A question reaches the speaker only through the single emission site (`VoxrCommandParser.cs:2737`), and the leading-required-miss bar (`:2679`, with the eager path's copy at `:4423`) sits above it unconditionally — no flag, no threshold. `TryFindSiblingRival` (`:2194-2205`) only names a `setId` where winner and rival share a set, so the winner of any askable tie is itself a set member. The remaining hole — could an anchored set's winner have *matched* its anchor and still tie — is closed by `CompareCandidate` (`:3333-3349`) requiring equal `startIdx`, with `RequiredLiteralMissPenalty = 0f` against `OptionalLiteralScore = 0.5f` leaving no compensating route. **The issue #126 rival path cannot route a question for an every-member-anchored set.**
- **Invariant 2 holds.** A mixed set still warns (`:1863-1884` returns false on the first unanchored member); trailing and medial discriminators still warn; below-`minScore` sets still warn, because no score test entered the collision loop. All four existing collision-expecting fixtures discriminate at element 3 and are untouched.
- **Invariants 3 and 4 hold.** The per-pair narrowing still decides which member is named, and the dedup's asymmetry is harmless: it is only reached when the anchor predicate is false, so adding it there would be a vacuously-true conjunct.
- **Call-site safety.** `_commands` (`:223`) is assigned at `:444`, well before `:673`; members come from `FindSiblingSets(_commands)`, so the index chain cannot go out of range, and `anchor` cannot remain `-1`.
- **The claimed arithmetic is exact.** Frame `[null,"mark","contact"]` → `(3−1)/3 = 0.667 ≥ 0.6`, so the score gate does not withhold and the anchor guard really is the only thing suppressing both messages. `"negative"` is in `DefaultCancel` (`VoxrPendingCommand.cs:64`); `"friendly"` is not.
- **The `minScore` asymmetry the PR reasons from is real.** The parser constructor takes only slots + commands and judges against `const float DefaultMinScore = 0.6f` (`:149`), while `VoxrCommandRecogniser.cs:31`'s serialized `minScore` reaches the runtime gate (`:639`, `:759`, `:929`).
- **"Remedy 2 above" resolves correctly**, and no shipped-grammar warning changes hands — none of the demo grammar's discriminating values is cancel vocabulary.
- **Editor-only holds.** The sole call site carries `[Conditional("UNITY_EDITOR")]` (`:1609`); the only Runtime hunk is inside it.

### Not verifiable from here

- **The Unity run counts were not independently re-run** by the review agents (multi-minute operation, deliberately not launched). EditMode 133/133 and PlayMode 588/588, and the mutation run's 587/588, are the author's claims — everything static about them was checked, and finding 3 is what that check produced.
- On-device (Quest) behaviour: not applicable to this diff, and that reasoning was audited rather than assumed.

jinwoo1601:
All three review findings addressed.

- **Finding 1** (`VoxrCommandParser.cs:1753`, the only one in shipping code) — fixed in c8a0f4a. The comment now claims only what holds: the walk runs for a member whose value collides and is answerable, *not* for one that would ultimately be reported. It also records why the guard stays above the dedup rather than below it — hoisting it to spare a deduped member would charge every set a walk to save the rare colliding one. Comment-only; no executable statement changed. Re-verified: EditMode **133/133**, PlayMode **588/588**, no compile diagnostics.
- **Finding 2** — PR body corrected: it now names the issue's shape 2 as the one fixed and shape 1 as the one deferred, instead of borrowing an ordinal that indexed the other way.
- **Finding 3** — PR body corrected: the green run is stated as `main`'s baseline 587 plus this PR's one new test = 588, replacing the false claim that the totals matched the baseline.

The four **Notes** in the review remain unactioned by design — they are improvements rather than defects, and the test-premise one names a pre-existing house split. Still unmerged and unapproved; the ruling is yours.

PR #141, opened 2026-08-28
Judge the construction-time sibling scan against the configured minScore (#140)

### LINKED ISSUE #140: Cancel-collision report still fires below the default minScore, where nothing can fire
Successor to #132, which was closed by #139. That PR fixed one of #132's two shapes — the leading-discriminator one — and **deliberately declined the other**. This issue carries the declined half so the record survives the close.

## The shape

A sibling set whose members cannot clear the default `minScore` once the discriminator is dropped, where a member's discriminating value is cancel vocabulary and it has an answerable co-member.

Losing one required element from a frame worth `D` leaves `(D − 1) / D`, so a two-element pattern drops to `0.5` — under the shipped `DefaultMinScore` of `0.6`. **Both** siblings are rejected, nothing fires, no pending opens, and no "which did you mean?" is ever asked. The cancel-collision report fires anyway, telling the author to rename a grammar literal to protect an answer that can never be given.

That is the same false-advisory class #81 spent a feature reversing, and the standard the report's own comment holds itself to (`Runtime/Commands/VoxrCommandParser.cs`, the collision loop in `WarnOnSiblingDiscriminator`).

## Why #139 did not just inherit the score gate

The sibling-tie filter beside it already carries `ScoreAfterDroppingDiscriminator(set.Frame) >= DefaultMinScore`. Copying that conjunct into the collision loop is a one-liner, and it was considered and **rejected on the merits** — not overlooked:

> the parser is never handed the recogniser's configured `minScore`, so it judges against the default, while the **runtime gate is** handed the configured value.

`KNOWN_LIMITATIONS.md` records that asymmetry under *"The sibling warning is silent below the default `minScore`"*, including that the runtime still routes such a tie to the speaker for an author who lowered the threshold. So gating the collision report on `DefaultMinScore` would **suppress an advisory that is true** for exactly the author who most needs it — trading a false positive at default settings for a false negative at lowered ones.

The anchor condition #139 did inherit has no such problem: it is positional, reads no configuration, and holds at every threshold.

## What the real fix probably is

Not the score gate — the **root cause**, which is that the parser is constructed from the grammar alone and never sees the configured threshold. Plumbing `minScore` into the parser would let the sibling filter *and* the collision report judge against the real value, and would retire the `KNOWN_LIMITATIONS` entry above rather than extending it to a second message.

That is a wider change than either message, so it wants its own scoping. Anyone picking this up: do **not** close it by adding the `DefaultMinScore` conjunct to the collision loop. That is the reverse error, and it has already been declined once.

## Test gap

No fixture combines a cancel-vocabulary discriminator with the below-`minScore` shape. All four existing collision-expecting fixtures in `Tests~/Runtime/VoxrCommandParserTests.cs` discriminate at element 3 of a three-element pattern — `(3 − 1) / 3` = `0.67`, clear of the gate. #139 added a fixture for the anchor shape only.

## Related

- #132 — the parent, closed by #139 with this half unfixed
- #139 — the anchor-half fix, and where this exclusion is argued in the shipped code comment and the CHANGELOG
- #81 — the feature that established the false-advisory standard
- #124 — the leading-required-miss bar, which made the anchor half tractable

jinwoo1601:
Closes #140

## What

`VoxrCommandParser` now receives the recogniser's **configured** `minScore` instead of judging against a hardcoded copy of the shipped default, so both of its construction-time Editor warnings reason about the gate that will actually run.

- `VoxrCommandParser`'s public constructor gained a **final optional** `float minScore = DefaultMinScore`, stored in a new `readonly float _minScore`. Appended last, so every existing positional call site compiles and behaves exactly as before — additive and non-breaking.
- The sibling-tie filter now compares `ScoreAfterDroppingDiscriminator(set.Frame) >= _minScore`.
- The cancel-collision loop **inherits that score condition**, which is the fix for #140: `ScoreAfterDroppingDiscriminator(set.Frame) < _minScore || DiscriminatorIsEveryMembersAnchor(set)`. The score test is placed first so it short-circuits the anchor predicate's walk over members and their authored patterns.
- `VoxrCommandRecogniser` passes its configured `minScore` at both parser-construction sites.
- `VoxrBatchTestRunner` passes through the threshold **it already accepted and previously dropped** — both its constructors took a `minScore`, stored it for their own gating, then built the parser without it. That latent inconsistency (harness gating on one threshold, its parser predicting against another) is closed here.

Nothing about parsing, scoring, or which commands fire moves. Both messages are Editor-only via `[Conditional("UNITY_EDITOR")]` on the call site, and `_minScore` is read by nothing else — `Parse`, `ParseInternal` and the scoring helpers never touch it, and `TryEagerCommit`'s gate stays on the `minScore` it is handed per call.

## Why

#140 is the surviving half of #132. A sibling set that cannot clear the gate once the discriminator is dropped — a two-element pattern falls to `(2 − 1) / 2 = 0.5`, under the `0.6` default — has **both** members rejected. Nothing fires, no pending opens, no question is ever asked. The cancel-collision report fired anyway, telling the author to rename a grammar literal to protect an answer that can never be given: the same false-advisory class #81 spent a feature reversing.

**The one-line fix was forbidden, and rightly.** PR #139 weighed adding the score conjunct and declined it on the merits — twice. The parser was blind to the configured threshold, so a test judged against `DefaultMinScore` would have silenced this advisory for the author who lowered `minScore`, the one author for whom it is **true**. That objection was correct on the facts it had.

This PR removes the blindness rather than overruling the objection. Judged against `_minScore`, the conjunct is sound; judged against the constant it would still be wrong. The shipped comment at the guard makes that distinction explicit, because it is the thing a future reader is most likely to get backwards.

The root-cause fix also retires the documented asymmetry it created: the sibling-tie warning no longer goes silent below the default, so the `KNOWN_LIMITATIONS.md` entry recording that is retired rather than extended to a second message.

## Validation

- [x] **EditMode — 133/133 passed**, 0 failed / 0 inconclusive / 0 skipped. `VoxrCommandParserDiagnosticTests` ran all 20 cases; it constructs `VoxrCommandParser` at eight sites, so the changed constructor signature is exercised, not merely compiled.
- [x] **PlayMode — 590/590 passed**, 0 failed / 0 inconclusive / 0 skipped. `Tests~/Runtime` is where the parser tests live.
- [x] Both platforms run via the host project `VoXR TestGround` (Unity 6000.4.7f1) per the project's verification bindings — this is a bare UPM package and nothing compiles or tests standalone.

### Mutation-verified (the negative test is not vacuous)

`LogAssert.NoUnexpectedReceived()` passes trivially if the grammar never formed the shape under test, so a green negative test proves nothing on its own. The fix was disabled — the guard reverted to `if (DiscriminatorIsEveryMembersAnchor(set))` with everything else left in place — and PlayMode re-run:

**589/590, exactly one failure:**

```
VoxrCommandParserTests.SiblingWarning_TwoElementCancelCollision_IsWithheldAtTheDefault

Unhandled log message: '[Warning] [VoxrCommandParser] Intent 'mark_negative' carries
the discriminating value "negative" at element 2, which is also in the default cancel
vocabulary. ...'. Use UnityEngine.TestTools.LogAssert.Expect
  at VoxrCommandParser:WarnOnSiblingDiscriminator () VoxrCommandParser.cs:1848
```

No other test failed, so the change's blast radius is exactly the two new fixtures. `SiblingWarning_TwoElementCancelCollision_ReportsAtALoweredMinScore` still passed under the mutation, as expected — at a configured `0.4` the report is owed either way. The source was then restored from backup and verified identical by checksum.

This evidence does not survive in the merged test, which is why it is recorded here.

### New test coverage

No existing fixture covered #140's shape — all four collision-expecting fixtures discriminate at element 3 of a three-element pattern (`2/3 = 0.667`), clear of the gate.

- `SiblingWarning_TwoElementCancelCollision_IsWithheldAtTheDefault` — the fix itself.
- `SiblingWarning_TwoElementCancelCollision_ReportsAtALoweredMinScore` — the same grammar at `minScore: 0.4f`, where **both** messages are owed and fire.

The pair moves together by design: the negative alone is satisfied by an implementation that simply stopped reporting collisions, the positive alone by one that always reports. Only together do they pin that the report is conditional on the *real* threshold. The positive also pins **where** the configured value arrived — the sibling-tie warning firing proves it reached that filter, the collision report firing proves it reached the collision loop.

The grammar is `["mark","negative"]` / `["mark","friendly"]`, deliberately discriminating at **element 2 and not element 1**. The existing `SiblingWarning_TwoElementFrame_IsDetectedButNotWarnedAbout` uses `["cease","fire"]`/`["resume","fire"]`, where the discriminator leads and the set is withheld twice over — by score *and* by anchor — so it cannot isolate the score condition. Moving the discriminator to the front would make the new test pass for the #132 reason instead, proving nothing about #140.

### Checked and unmoved

- No existing test depends on a collision report below `0.6`; none of the four collision fixtures carries an optional element that would shift the denominator.
- No pinned demo-grammar count moves. `SiblingWarning_DemoGrammar_WarnsOnlyOnTheReachableTie` constructs at the default and the new condition can only *suppress*; no cancel-collision report fires on the demo grammar today.
- No test anywhere configures a recogniser `minScore` below `0.6`, so routing the configured value into the sibling filter cannot newly light up warnings in the recogniser-level suites.
- `SiblingWarning_ReachabilityGateTracksTheRecogniserDefault` keeps its assertion unchanged — `DefaultMinScore` still mirrors the recogniser's serialized default, now as the parameter's fallback. Only its comment changed.

## Docs

Swept for the **outcome**, not the symbol — a behaviour change falsifies prose that never names the changed identifier.

- **`KNOWN_LIMITATIONS.md`** — *"The sibling warning is silent below the default `minScore`"* is **retired**. Its root cause is precisely what this change removes. The file carries no "Fixed in"/archive convention, so clean removal is what it already does. Its still-true positional content (the anchor withholding, which no threshold changes) already existed in the neighbouring entry, which picked up three surgical updates instead.
- **`Documentation~/command-recognition.md`** — four passages, including the PR #139 cancel-collision site (now "rules out three shapes") and the "what is frozen when the parser is built" enumeration, which was incomplete without the `minScore` copy.
- **`Documentation~/troubleshooting.md`** — a sentence asserting the warning under-reports below `0.6`, plus a pointer that would have dangled at the deleted entry.
- **`Documentation~/api/command-recogniser.md`, `api/batch-test-runner.md`** — the two lifetimes: the runtime gate is read fresh per parse; the construction-time copy is frozen until the next rebuild.
- **`CHANGELOG.md`** — the existing `#132` bullet is **rewritten, not repointed**: its "the companion exclusion is deliberately not inherited" argument described an interim state that never shipped to users, its "remains open" sentence is gone, and it also carried one of the two copies of the now-falsified "the parser is built from the grammar alone and never sees your configured threshold" claim. A new `Fixed` entry covers #140. Two further `[Unreleased]` bullets were corrected for **different** reasons: one carried the other copy of that falsified claim, while the other changed only because the `KNOWN_LIMITATIONS.md` entry it referenced is retired here, so its reference was re-pointed rather than a false claim removed. #140 ships inside 2.0.0, so leaving any of them would have shipped release notes contradicting their own neighbours.

## Notes for review

- **Not deferred to device.** This is a construction-time Editor advisory — no mic capture, no MonoBehaviour lifecycle, no native-bridge ABI. `NativeBridge~/` is untouched, so no `.so` rebuild is implicated.
- **Formatting:** all four `.cs` files were already non-csharpier-compliant at `HEAD` at lines unrelated to this change. Verified before/after that these edits introduce no new violations; running `csharpier format` would have produced a large unrelated diff, so it was deliberately not run.
- The one-line fix `ScoreAfterDroppingDiscriminator(...) >= DefaultMinScore` remains wrong and is **not** what this does. That distinction is the load-bearing part of the change.

## Post-review fixes

`review-pr` (slim + `XFILE`) ran on this PR and posted its verdict: **3 confirmed minor findings, 3 refuted, no blockers.** All three confirmed findings have since been fixed on this branch:

1. **`CHANGELOG.md`** promised non-breaking compat to a caller that cannot exist. `VoxrCommandParser` is `internal` (`:137`), with `InternalsVisibleTo` naming only this package's own test and editor assemblies, so the "additive and non-breaking / a caller that constructs the parser directly" language addressed an audience with no members. The clause now states the type is internal and no integrator code is touched, and points at the one consequence a consumer *can* observe — public `VoxrBatchTestRunner` propagating a threshold it previously dropped.
2. **`Tests~/Runtime/VoxrEagerCommitTests.cs`** carried a comment asserting the warning "must judge against a default minScore **the constructor cannot see**" and that the gate "judges exactly rather than predicting" — both falsified by this PR. Rewritten: the pair is withheld by **position** at any threshold (issue #132 / PR #139), and additionally on score because that test's `Build()` supplies no threshold and takes the `0.6` fallback; the surviving gate-vs-scan distinction is **lifetime** (per-call parameter vs captured at construction), not blindness. Prose only — stripping whole-line comments leaves the file byte-identical, so the assertions and `Build()` helper are provably untouched and the suite was not re-run.
3. **This PR body** misattributed a CHANGELOG edit. The falsified blindness claim lived at base lines **14 and 24**, not 14 and 16; line 16 changed because its referent was deleted. Corrected in the Docs section above.

The three refuted findings were left alone: a `troubleshooting.md` lifetime contrast that carries its own scope clause, a NaN `minScore` guard asymmetry that is a documented trade-off and already present in the runtime being predicted, and a "only its comment changed" phrasing whose substance the body already spelled out.


jinwoo1601 (commented):
## Independent review — PR #141

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE` — **plus `XFILE`**, added because the diff changes a constructor signature consumed at four call sites across three files. `PIT` deliberately not added: no MonoBehaviour lifecycle, coroutine, or main-thread-only API is touched.

**Head reviewed:** `e056939cec233f1b11069b79aa41c86665ea6c93` · **merge-base:** `9b11738`

Four finder agents ran against the diff with no context from the sessions that wrote it; six candidate findings were adversarially verified. **Three confirmed, three refuted, no blockers.** The core change — the guard at `VoxrCommandParser.cs:1828` comparing against the configured `_minScore` rather than `DefaultMinScore` — was checked against issue #140's explicit prohibition and is **not** the forbidden fix.

---

### 1. CHANGELOG promises a non-breaking guarantee to a caller that cannot exist — `CHANGELOG.md:26` · minor · CONFIRMED

The entry says the constructor change is *"additive and non-breaking: every existing call site compiles and behaves exactly as before, and a caller that constructs the parser directly without supplying a threshold still gets the `0.6` fallback."*

The parameter genuinely is last (`VoxrCommandParser.cs:468`), so the compile half is delivered. But the type is `internal class VoxrCommandParser` (`:137`), and `Runtime/AssemblyInfo.cs:9-11` grants internals only to `Tests.Runtime`, `Tests.Editor`, `Editor` — all first-party. The package's own API reference already says so: `Documentation~/api/batch-test-runner.md:5`, *"Drives an internal parser instance directly."* No public subclass, factory, or wrapper surfaces the constructor. The only callers omitting `minScore` are the 223 sites under `Tests~/`, which this same changelog records as stripped from the published package (`CHANGELOG.md:48`) — so the fallback guarantee addresses an empty set, in a file whose header declares SemVer adherence and which `package.json` `changelogUrl` points consumers at.

Secondary, and arguably the larger point: the bullet bolds the non-change while the one genuinely consumer-visible consequence — public `VoxrBatchTestRunner` now propagating a threshold it previously dropped, changing which Editor warnings a consumer's batch run logs — gets a subordinate clause.

**Fix:** scope the compat language to first-party call sites, or drop the "caller that constructs the parser directly" clause.

---

### 2. Test comment asserts the constructor "cannot see" `minScore` — `Tests~/Runtime/VoxrEagerCommitTests.cs:1716-1721` · minor · CONFIRMED

`TryEagerCommit_LeadingTwoElementTie_RefusesForScoreAtDefaultAndForTheBarBelowIt` opens: *"The warning is SILENT about this pair because it must judge against a default minScore **the constructor cannot see** … This gate is handed the real configured threshold, so it judges exactly rather than **predicting**."*

Both clauses are falsified here, and the new comment at `VoxrCommandParser.cs:1789` states the opposite in the same merged tree: *"Judged against `_minScore`: the threshold the recogniser was CONFIGURED with, the same value its runtime gate applies."* The asymmetry this test is **named** for is the one #140 closes. The situational defence — that this test's parser is built without a `minScore` and so does fall back to `0.6` — does not hold, because the sentence asserts a capability, and the "rather than predicting" contrast makes the capability reading unavoidable.

The test still passes for the right reason (`_minScore` is read only inside the Editor-only scan), so this is stale reasoning, not a red test.

**One leg of this finding was refuted:** the `— documented in KNOWN_LIMITATIONS` pointer does **not** dangle. It names no heading, and the surviving entry still documents the case at `KNOWN_LIMITATIONS.md:753-755`.

**Fix:** restate the reason as the anchor/positional withholding, or as the fallback threshold.

---

### 3. PR body misattributes a CHANGELOG edit — `CHANGELOG.md:16` · minor · CONFIRMED

The body says two `[Unreleased]` bullets, *"lines 14, 16"*, asserted "the parser is built from the grammar alone and never sees your configured threshold" and were corrected. At base that string occurs at lines **14 and 24** — never 16. Line 16's only threshold text was the *title* of the `KNOWN_LIMITATIONS` entry it references, and the word-diff shows that clause kept verbatim; its edit re-points a reference whose referent was deleted, which is a different rationale.

Both edits are themselves correct. Only the body's account of line 16 is wrong — and it correspondingly under-credits line 24, which did carry the second copy.

**Fix:** correct the body. No code change.

---

- **`troubleshooting.md:75` lifetime contrast** — the passage carries its own scope clause ("if you are tuning live") inside a section about what *fires*; the construction-time copy governs only Editor warnings. The finder's "one other place in the docs" premise was false: `troubleshooting.md:99` and `command-recognition.md:436` both state it. Incompleteness with no failure path.
- **NaN `minScore` splits the two guards** — derivation sound (both `x >= NaN` and `x < NaN` are false; NaN is reachable via public `VoxrBatchTestRunner`), but it is a documented, accepted trade-off at `VoxrCommandParser.cs:492-494`, which reasons about NaN explicitly and bounds harm to "admit or withhold a warning." Decisively, the same `>=`/`<` split already exists untouched in the runtime being predicted (`VoxrCommandRecogniser.cs:761` vs `:931`), so at NaN the recogniser is already broken upstream of any advisory.
- **"Only its comment changed"** — the body's preceding clause ("now as the parameter's fallback") *is* the new assertion message, so nothing was concealed; an NUnit `message` argument is inert prose. Pedantry about defensible phrasing.

### Verified clean

`GONE` returned no findings on a deletion-heavy diff, walking the retired ~60-line `KNOWN_LIMITATIONS` entry clause by clause and locating a surviving home for every still-true statement; no dangling anchor links repo-wide. `XFILE` traced all 227 parser construction sites and ruled out a positional transposition (the two `float` parameters are separated by two `string[]` and a `bool`, so a swap could not compile). Both new test regexes match the shipped format strings verbatim. The four rebuild triggers named in the docs resolve. The two new fixtures do isolate the score condition: `DiscriminatorIsEveryMembersAnchor` returns false for `["mark","negative"]`/`["mark","friendly"]`, so the score gate is the sole cause of withholding.

### Notes — unverified, no action implied

- `VoxrBatchTestRunner`'s `minScore` default is its own `0.6f` literal, not `DefaultMinScore`; the sync test pins only the recogniser's field, so the runner's literal can drift out from under the parser scan it now feeds.
- `VoxrBatchTestRunner` still omits `effectiveCancelVocabulary`, so a batch run's collision report predicts against the default cancel vocabulary regardless of a configured override. Pre-existing.
- No `KNOWN_LIMITATIONS` entry records the successor residual — the parser's copy is frozen at construction, so an Inspector edit leaves the scan stale until a rebuild. Documented in three other places, but not as a limitation entry.
- The mutation account's "no other test failed, so the blast radius is exactly the two new fixtures" is weaker than stated: an unexpected `LogWarning` fails a fixture only where it calls `NoUnexpectedReceived()`.
- `VoxrEagerCommitTests.cs:1716` was already partly stale at base — #139's anchor test independently silences that pair at any threshold, so "0.5 is under it" was never the sole cause. That residue predates this PR.

The Unity results (EditMode 133/133, PlayMode 590/590) and the mutation run (589/590, one failure) were produced in the host project `VoXR TestGround` and **could not be re-performed by the review agents** — this is a bare UPM package that does not build standalone. They were assessed for internal consistency only, and are consistent: the described revert removes exactly three lines, which places the collision `LogWarning` at the line the quoted stack frame names, and the quoted assertion text matches `BuildCancelCollisionWarning`'s format string verbatim for that grammar. On-device (Quest) behaviour is not implicated by an Editor-only construction-time advisory.

*Posted by `review-pr` · slim + XFILE · no approval implied, no merge performed.*

jinwoo1601:
All three CONFIRMED findings from the review above are now addressed — pushed as `5cfa05a`, plus a PR-body correction.

- **Finding 1** (`CHANGELOG.md:26`) — the compat clause is rewritten. It now states that `VoxrCommandParser` is internal with first-party-only `InternalsVisibleTo`, so this is not a public-API change, and points at the one consequence a package consumer *can* observe: public `VoxrBatchTestRunner` propagating the threshold it previously dropped.
- **Finding 2** (`Tests~/Runtime/VoxrEagerCommitTests.cs:1714`) — the comment no longer claims the constructor "cannot see" `minScore`. It now says the pair is withheld by **position** at any threshold (#132/#139), and on score only because that test's `Build()` supplies no threshold and takes the `0.6` fallback; the surviving gate-vs-scan distinction is **lifetime**, not blindness. One narrowing against the review's wording: the anchor predicate is described as withholding *the sibling warning* here rather than "both messages", because neither "cease" nor "resume" is cancel vocabulary — saying "both" would imply the collision report was otherwise live, which is the defect class being fixed.
- **Finding 3** (PR body) — corrected. The falsified blindness claim lived at base lines **14 and 24**; line 16 changed only because its referent was deleted.

The three refuted findings were deliberately left alone.

**Not re-run, and why:** both edits are prose-only. Stripping whole-line `//` comments makes `VoxrEagerCommitTests.cs` byte-identical to the previously-verified commit, with unchanged brace balance, so no assertion, helper, or compiled behaviour moves. The EditMode 133/133 and PlayMode 590/590 results, and the mutation evidence, all still stand against `e056939`'s code, which this commit does not touch.

PR #149, opened 2026-09-06
Warn when a grammar word is not in the model vocabulary (#143)

### LINKED ISSUE #143: Warn at authoring time when a grammar word is not in the model vocabulary
## Summary

A grammar word the VOSK model does not know is dropped from the decoder grammar silently. Nothing in the package, the bridge, or the Console tells the author. The failure surfaces only in the field as `[unk]` (see `KNOWN_LIMITATIONS.md`, "Abbreviations and letter sequences map to [unk]", where `cqb` is the documented case).

Add an authoring-time check: every entry the grammar generator emits is looked up in the loaded model's vocabulary, and each unknown word produces an Editor warning that names the word, the slot or pattern it came from, and the documented remedies (spell it out, or add a phonetic alias).

## Why it is silent today

- The bridge calls `vosk_set_log_level(-1)` (`NativeBridge~/src/vosk_bridge.cpp:215`), which suppresses the Kaldi warning `Ignoring word missing in vocabulary: '...'` that VOSK emits for every unknown grammar entry. The warning string is present in the shipped `libvosk.so`.
- `vosk_model_find_word(VoskModel*, const char*)` is exported by both shipped libraries (verified with `nm -D` on `Runtime/Plugins/Android/arm64-v8a/libvosk.so` and `strings` on `Runtime/Plugins/x86_64/libvosk.dll`) and is already declared at `NativeBridge~/include/vosk_api.h:21`. Nothing calls it.

## Proposed change

- Editor path: P/Invoke `vosk_model_find_word` in `Runtime/Native/VoxrNative.cs` and run the check over the output of `VoxrCommandParser.GenerateGrammarJson` (split multi-word entries into words; skip `[unk]`) whenever the grammar is built and a model handle is available (`EditorMicBackend` owns one).
- Warn once per unknown word per grammar build, in the same style as the existing construction-time authoring warnings (Editor only, `Debug.LogWarning`, never reject).
- Device path is optional: the check is an authoring aid, and the authoring happens in the Editor. If a bridge export (`vosk_bridge_find_word`) is added anyway, the rebuilt arm64 `.so` must be committed and on-device verification reported as deferred (human-only).

## Acceptance

- EditMode or PlayMode test in the host project: a grammar containing `cqb` produces exactly one warning naming `cqb`; a grammar containing only `cease` and `fire` produces none. (The sample `namedRange` slot at `Samples~/CommandRecognition/CommandDemo.cs:65` and `Tests~/Runtime/DemoGrammar.cs:40` already carries `cqb`.)
- Multi-word entries are checked word by word (`safe range` checks `safe` and `range`).
- No warning when no model is loaded (text-injection-only sessions).
- Docs: one sentence in `Documentation~/command-recognition.md` (What the grammar contains) and the `KNOWN_LIMITATIONS.md` `[unk]` entry pointing at the new warning.

## Out of scope

Automatic alias generation (G2P), and any change to what the decoder does with the unknown word.

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), critique item R7, ranked 1 of 12. Ruling 3 (2026-09-06): the leading-required-miss bar is absolute; this check changes nothing at match time.

jinwoo1601:
Closes #143

## What

Binds `vosk_model_find_word` and checks every word of a generated grammar against the loaded VOSK model. Each unknown word logs one Console warning naming it and the two documented remedies. Advisory only — the grammar is applied unchanged, and nothing on the parse/score/fire path moves.

| File | Change |
|---|---|
| `Runtime/Native/VoxrNative.cs` | one new `[DllImport]`, `vosk_model_find_word` |
| `Runtime/Native/VoxrGrammarVocabulary.cs` | **new** — `ExtractWords` (pure) + two `WarnOnUnknownWords` overloads |
| `Runtime/EditorMicBackend.cs` | one call added inside `SetGrammar` |
| `Tests~/Runtime/VoxrGrammarVocabularyTests.cs` | **new**, 6 PlayMode tests |
| `CHANGELOG.md`, `KNOWN_LIMITATIONS.md`, `Documentation~/command-recognition.md` | docs |

**Windows Editor only.** `VoxrNative` is gated `#if UNITY_EDITOR_WIN`, not `UNITY_EDITOR` — narrower than the issue implies. Player builds and non-Windows Editors are unchanged. No public API added, no native rebuild, `NativeBridge~/` untouched.

## Why

A grammar word the model does not know is discarded when the grammar is compiled, and the Kaldi warning is suppressed by `vosk_set_log_level(-1)` (`NativeBridge~/src/vosk_bridge.cpp:215`). The only symptom was a phrase that never matched, with nothing in the transcript saying why.

### Two deviations from the issue text, both deliberate

**1. Placement — the issue's suggestion would have missed the common case.** The issue proposes running the check "whenever the grammar is built". Grammar generation (`GrammarManager.Rebuild`) can run *before* a model exists — that is precisely why `ApplyIfReady` exists — so a check hung off generation would silently skip the ordinary startup ordering. It runs instead in `EditorMicBackend.SetGrammar`, the only site where a grammar and a live model handle meet. This also covers runtime rebuilds (`SetActiveSets`, `RebuildGrammar()`) and needs no accessor for the private `_model`. ~~`NotifySlotChanged`~~ — corrected during review: it rebuilds only the parser and never re-applies a grammar, so it does not reach the check (finding 2).

**2. No slot/pattern provenance in the message — please rule on this.** The issue asks the warning to name "the slot or pattern it came from". At the check site the grammar is already a flat, deduped, sorted list of entries — mostly multi-word phrases, not single words — and no slot or pattern attribution survives into it; `EditorMicBackend` holds no slot or command definitions to recover it from. Recovering it means either plumbing slots and commands into the Editor mic backend or duplicating the parser's grammar walk, both of which felt disproportionate for an authoring aid. The issue's own acceptance criteria only require naming the word, so that is what shipped. Say the word and I will do it as a follow-up.

## Measured facts behind the implementation

Probed directly against `vosk-model-small-en-us-0.15`. Upstream **does** document the contract (`vosk_api.h`: the word symbol if present, `-1` otherwise, symbol 0 being `<epsilon>`) — the trimmed copy this repo carries at `NativeBridge~/include/vosk_api.h:21` drops that comment, which is what the original claim here got wrong (finding 3). The measurements stand as corroboration:

- **Unknown → `-1`, not `0`.** Known words return positive ids (`cease` 66827, `fire` 146905, `target` 55604). `0` is a valid symbol id, so a `!= 0` test would have been wrong.
- **`[unk]` is itself in the vocabulary** (id 152208). It is still skipped explicitly — it is the package's own token, not author vocabulary, so a future model lacking it must not warn about a word the author never wrote.
- **Running the real check over the real generated grammar** (the `grammar` array in `NativeBridge~/harness/expectations.json`, genuine `GenerateGrammarJson` output): 133 entries → 97 author words → **exactly 2 unknown: `cqb` and `railgun`**.

### Two doc corrections this measurement forced

- **`railgun` was undocumented.** `KNOWN_LIMITATIONS.md` recorded only `cqb`. Both live in the sample's `range` slot. The sample vocabulary is **not** changed — the point is to warn about it, not quietly fix it.
- **`KNOWN_LIMITATIONS.md` named `pdc` as out of vocabulary, which is false** — `pdc` → id 87893, it is present. That line is corrected and `pdc` is now called out as an abbreviation-shaped word that the model *does* know.

## Review outcome

`review-pr` ran at the slim profile plus `XFILE` and posted its verdict on this PR: **6 confirmed, 2 refuted**, every one about text rather than logic. All six confirmed findings have now been addressed at the maintainer's instruction, across three commits. Two were blockers I had introduced about my own feature — a warning that misquoted the remedy it cited, and a changelog claim that `NotifySlotChanged` reaches the check when it cannot. The rest corrected a false provenance comment, a "the model this package vendors" clause the package contradicts in two other files, an API-reference example that now trips the new warning, and two `[unk]`-exemption imprecisions.

A separate pre-existing hazard found while verifying is filed as #150: `vosk_recognizer_new_grm` **segfaults** on malformed grammar JSON instead of returning NULL, on both the Editor and Android paths.

## Validation

- [x] **Unity EditMode — green, 133/133** (`failed=0`, `inconclusive=0`), host project `VoXR TestGround`, Unity 6000.4.7f1.
- [x] **Unity PlayMode — green, 596/596** (`failed=0`, `inconclusive=0`).
- [x] **The six new tests are confirmed present and `Runnable` in the PlayMode result XML**, each with a non-zero duration — checked because a file wrapped in `#if UNITY_EDITOR_WIN` that compiled to nothing would otherwise be indistinguishable from a pass.
- [x] **The production path is exercised end to end by the existing suite.** The PlayMode log carries 34 warnings arriving via `RebuildGrammarInternal` → `ForceApply` → `SetGrammar` → the `IntPtr` overload — 17 WAV-replay tests × the 2 genuinely unknown words — so the new `[DllImport]` resolves and the `-1` sentinel is correct against a real loaded model in Unity, not only in my offline C probe.
- [x] **Mutation-verified.** Inverting the predicate to `if (!isKnown(word))` (warn on known words, stay silent on unknown) turned PlayMode to 593/596 with exactly the three warning-path tests red and the three `ExtractWords_*` tests green. In particular `WarnOnUnknownWords_AllKnown_WarnsNothing` — a pure negative assertion — **did** fail, so it is not inert. Mutation reverted; the committed file is sha256-identical to the one that ran green.
- [ ] **On-device (Quest) verification — not applicable and not claimed.** The change is `UNITY_EDITOR_WIN`-only and compiles to nothing on Android.

### Residual gaps, stated plainly

- The mutation made tests 4 and 6 fail via Unity's *unhandled-log* check (an unexpected warning naming a known word) rather than via their unmet `LogAssert.Expect`. So their expectation-not-received path is **not** exercised by this run. A mutation that only suppressed the `cqb` warning without adding noise would be needed to prove that half; I judged a third full Unity pass not worth it, since ~100 existing call sites in the suite rely on the same mechanism.
- **New log noise:** a PlayMode run now emits 34 of these warnings, and the sample scene will warn twice on every grammar build until `cqb`/`railgun` are addressed. That is the feature working, but it is a visible change to Console output and worth knowing before merge.
- Warnings re-fire on every grammar rebuild rather than once per session, matching how the existing construction-time warnings behave (each rebuild constructs a fresh parser).


jinwoo1601 (commented):
## Independent review — PR #149

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE` — **plus `XFILE`**, added because the diff introduces a native-interop contract (a new `[DllImport]` that must agree with a third-party C header and resolve against a shipped binary). No angles dropped.

**Head reviewed:** `58e97ce6f138cfa13c81d7a9c7d08720cd4ae789`. Four finder agents ran with no context from the session that wrote the code; eight surviving findings went to one adversarial verifier each.

**Verdict:** 6 confirmed, 2 refuted. Two were ruled blockers and are **fixed in this run** (see the follow-up commit); the rest are posted unfixed for the maintainer to rule on. The implementation itself came through clean — every numbered finding below is about *text*, not logic. `LINE` re-derived the whole new file and found no defect in it.

---

### 1. `[blocker · CONFIRMED-WITH-CORRECTION]` The warning text misquoted the remedy it cites — **FIXED**

`Runtime/Native/VoxrGrammarVocabulary.cs:85` told authors to "spell it out as separate letters (e.g. `c q b`)" and cited `KNOWN_LIMITATIONS.md, "Abbreviations and letter sequences map to [unk]"` as its source. That section (`:104-108`) prescribes something different: "Spell out **the phrase** in the slot value (e.g. `close quarters` instead of `cqb`)". Grep for "separate letters" in `KNOWN_LIMITATIONS.md`: zero hits. The misquote had propagated to `Documentation~/command-recognition.md:722` and `CHANGELOG.md:12`.

**Corrected by the verifier:** the sharper charge — that following the advice trips the parser's single-character warning — is **false**. `VoxrCommandParser.cs:756` tests `value.Length == 1` on the whole slot value, so `"c q b"` (length 5) never trips it, and `c`/`q`/`b` are all in vocabulary (64266 / 30216 / 95510). This was a false citation, not a trap.

**Fixed** in all three files: the remedy now quotes what the cited section actually says.

### 2. `[blocker · CONFIRMED]` `NotifySlotChanged` cannot reach the check — **FIXED**

`CHANGELOG.md:12` claimed the check "also covers runtime rebuilds — a slot repopulated by `NotifySlotChanged`, or a `SetActiveSets` switch". The first half is false. `VoxrCommandRecogniser.cs:340` `NotifySlotChanged()` calls only `RebuildParser()`, which constructs a parser and touches no grammar — the package's own `Documentation~/api/command-recogniser.md:61` already says "**Does not touch the grammar or VOSK recogniser**".

The verifier killed all three escape routes: `NotifySlotChanged` never sets `GrammarRebuildDeferred` (written only at `:371`, `:391`, `:526`, `GrammarManager.cs:59`), the model-ready path returns early on `IsApplied`, and `RebuildGrammarInternal:382` passes raw `_slots` rather than `BuildEffectiveSlots` — so the grammar could not surface a new unknown word by that route even if re-applied. Not imprecise; the named mechanism is incapable.

**Fixed:** the changelog now names `SetActiveSets` and `RebuildGrammar()`, which do reach it.

### 3. `[minor · CONFIRMED]` A new code comment says upstream documents no return contract. It does.

`Runtime/Native/VoxrGrammarVocabulary.cs:94` asserts "The upstream header documents no return contract at all". The genuine upstream header vendored in the tree, `NativeBridge~/vendor/vosk-linux-x86_64-0.3.45/vosk_api.h:69-74`, documents it exactly: "*@returns the word symbol if @param word exists inside the model or -1 otherwise. Reminding that word symbol 0 is for <epsilon>*". Only the repo's **trimmed** copy at `NativeBridge~/include/vosk_api.h:21` is bare — and its own banner calls itself "Copied from the upstream vosk-api repository".

The steelman fails on the PR's own wording, which separates the two ("the upstream header documents **no** return contract **and the repo says nothing**"). Behaviour is unaffected: `-1` and "0 is a valid id" are both correct, and upstream independently confirms the `0` caveat this PR derived by measurement. **Fix:** cite the upstream contract, keep the measurement as corroboration.

### 4. `[minor · CONFIRMED]` "the model this package vendors" — the package vendors no model

`KNOWN_LIMITATIONS.md:93`, newly added text. Contradicted by `README.md:75` ("The SDK does not bundle a VOSK model") and `THIRD_PARTY_NOTICES.md:13` ("not included in this package"). `NativeBridge~/vendor/` is gitignored (`.gitignore:41`) and `git ls-files` returns zero tracked files there, so the model is not vendored even in the checked-in-dependency sense.

The clause is a removable appositive — "measured against `vosk-model-small-en-us-0.15`" already does the sentence's whole job. **Note for the maintainer:** the same inaccuracy pre-exists at `KNOWN_LIMITATIONS.md:16` ("the small English VOSK model ... we ship with"), 77 lines above; this PR propagates rather than originates it. Left unfixed pending your ruling on whether to close both.

### 5. `[note-tier · CONFIRMED]` `railgun` is still taught unremarked in the API reference

`Documentation~/api/command-definitions.md:64` — `new VoxrSlotDefinition("weapon", new[] { "rocket", "laser", "railgun" })`. This PR establishes `railgun` as out of vocabulary for the first time, so a reader copying the example now gets a warning about a word the reference just recommended.

The verifier lowered this below numbered-finding tier and I am reporting it anyway because it is one word to fix: the file is **byte-identical to base**, asserts nothing false, and ships in no built grammar (the sample's real `weapon` slot is `missiles`/`torpedoes`/`jackal`, all in vocabulary). It swept every `.md` in the repo and confirmed this is the **only** unremarked site. **Fix if you want it:** swap for `torpedo` (56964), or add a cross-reference.

---

- **Existing "validation warnings re-emit" entry (`KNOWN_LIMITATIONS.md:317`) should list the new check.** REFUTED. That entry is bound to *parser-constructor* passes in all four of its fields — its repro, its root cause ("the parser ctor unconditionally re-runs its validation passes"), its "five passes" count, and its proposed fix ("once per `Configure()`, not per parser rebuild"). The new check is not one: it fires on grammar *application*, including the initial apply where no parser is constructed, and being `UNITY_EDITOR_WIN` it can never appear in the logcat the entry describes. The proposed extension would have introduced three false statements.
- **The PR hides that its tests substitute a predicate instead of a real model.** REFUTED. The body already says the `IntPtr` path is exercised "by the **existing** suite" and that the 34 warnings are what "the PlayMode log **carries**" — an observation, stated as one. The issue's acceptance criterion does not require a model. The residue is a pure coverage gap, which this lane's materiality floor excludes.

---

### Notes — unverified, one line each

- **The remedy example in the message is a fixed literal, so it mismatches every word except `cqb`.** Observed in the re-verification log: `Grammar word "railgun" is not in the loaded VOSK model's vocabulary. ... Remedies: spell the phrase out in the slot value (e.g. "close quarters" instead of "cqb")`. It is marked `e.g.` so it reads as an illustration rather than a false claim, and it was equally true of the wording finding 1 replaced. Left alone deliberately: fixing it is a third edit to production text and would need another full suite run, so it is better bundled with whatever you rule on findings 3-5.
- **Pre-existing hazard found while verifying, worth its own issue:** `vosk_recognizer_new_grm` **SIGSEGVs** on malformed grammar JSON rather than returning NULL — measured on the vendored libvosk: `["a"b", "c"]` → exit 139, and `[]` → exit 139 too. The `_recognizer == IntPtr.Zero → ModelLoadFailed` path at `EditorMicBackend.cs:290` is therefore unreachable for that input; the Editor process dies. Unrelated to this diff and older than it.
- `ExtractWords`'s quote-pair scan mis-tokenises if a grammar word contains `"` (returns `["a", ","]` for `["a"b", "c"]`, silently skipping `b` and `c`). Only reachable via an authoring error that already crashes per the note above; slot values with punctuation are already warned at `VoxrCommandParser.cs:747`, though **pattern literals are not**. The comment at `:27-28` overstates ("scanning for quote pairs is sufficient") — it is sufficient only while no word contains a quote.
- The PR body calls the grammar "a flat, deduped, sorted **word list**"; it is a list of *entries*, most of them multi-word phrases. The conclusion drawn from it (no provenance survives) is unaffected and stands.
- `[System.Diagnostics.Conditional("UNITY_EDITOR")]` on both overloads is redundant inside a file already gated `#if UNITY_EDITOR_WIN` — inert, matches house convention.
- `Documentation~/command-recognition.md:722` says "every word the grammar emits is looked up"; `[unk]` is deliberately exempt.
- That same paragraph links `../KNOWN_LIMITATIONS.md` with no anchor, while the warning names the entry by title.
- Nothing asserts the `!= -1` sentinel, the UTF-8 marshalling, or the `IntPtr.Zero` guard. The guard is unreachable from its sole call site (`SetGrammar` returns early unless `IsInitialised`), so it is defence-in-depth.

---

### Confirmed clean (checked, not assumed)

The P/Invoke signature agrees with `vosk_api.h:21` on return width, arity, handle representation, calling convention and UTF-8 encoding, and matches its siblings. `vosk_model_find_word` is in the export table of both shipped libs (`objdump -p` ordinal 15645; `nm -D` `T`). Conditional-compilation guards agree across the definition, its call site and the tests in all four platform configurations. `InternalsVisibleTo` wiring permits the test access. The `.meta` GUID is unique repo-wide. Three agents independently reproduced the measurements against the vendored model: 133 entries → 97 author words → exactly `cqb` and `railgun` unknown, `pdc` → 87893 (so the `pdc` correction is right), `<eps>` → 0 (so "0 is a valid id" is right). No public API added; `NativeBridge~/` untouched; the diff is insertion-only in both changed source files, with all `SetGrammar` guards and the free/recreate sequence intact; nothing on the parse/score/fire path moves; `ZeroAllocPollPathTests`' pinned paths are untouched.

### Not verifiable from here

- **The Unity test runs.** No Unity is reachable from a review agent, so `133/133`, `596/596`, the six-tests-`Runnable` claim and the mutation counts were audited for internal consistency, not re-executed. They are internally consistent, and the 34 = 17 × 2 warning count is discriminating: an unresolved `[DllImport]` would have thrown, and a `!= 0` sentinel would have warned on all 97 words.
- **On-device Quest behaviour** — human-only, and not applicable: this code is `UNITY_EDITOR_WIN`-only and compiles to nothing on Android.

jinwoo1601:
Follow-up: pushed `f95aa32`, which addresses findings **1** and **2** (both ruled blockers — false statements this PR introduced about its own feature, in text that ships). Re-verified after the fix: PlayMode 596/596, all six `VoxrGrammarVocabularyTests` green, and the log confirms the corrected message renders cleanly with the old wording fully displaced.

Findings 3, 4 and 5 are left unfixed and unruled for you, per the issue lane's rule that only blockers are this run's own work. Each is a one-line edit and they would bundle into a single verified pass with the message-example note if you want them.

jinwoo1601:
All review findings are now addressed, per your instruction. Pushed `5cee944` on top of `f95aa32`.

| # | Finding | Status |
|---|---|---|
| 1 | Warning misquoted the remedy it cited | fixed in `f95aa32` (3 files) |
| 2 | `NotifySlotChanged` cannot reach the check | fixed in `f95aa32` |
| 3 | Comment claimed upstream documents no return contract | fixed in `5cee944` |
| 4 | "the model this package vendors" | fixed in `5cee944`, **plus** the pre-existing instance at `KNOWN_LIMITATIONS.md:16` |
| 5 | `railgun` taught in the API reference | fixed in `5cee944` — swapped for `torpedo` (in vocabulary) |
| — | Notes: hardcoded `cqb` example in the message; quote-scan comment overstated its precondition; two "every word is looked up" claims where `[unk]` is exempt; missing link anchor | all fixed in `5cee944` |

Two notes were deliberately **not** taken: `[Conditional("UNITY_EDITOR")]` inside a `#if UNITY_EDITOR_WIN` file is redundant but matches house convention, and the PR-body wording fix was applied to the description rather than the code. The two refuted findings stand refuted.

**Final verification at `5cee944`:** EditMode 133/133, PlayMode 596/596, `failed=0 inconclusive=0` on both, all six `VoxrGrammarVocabularyTests` green. The log confirms the corrected message on all 36 emissions with clean concatenation seams and zero occurrences of the old wording; CSharpier reports the file already formatted. On-device Quest remains not applicable (`UNITY_EDITOR_WIN`-only) and is not claimed.

The pre-existing crash found while verifying is filed separately as #150 — `vosk_recognizer_new_grm` segfaults on malformed grammar JSON instead of returning NULL, on both the Editor and Android paths. Nothing in this PR causes or worsens it.

PR #151, opened 2026-09-06
Record barred rounds and the runner-up in the session log (#144)

### LINKED ISSUE #144: Session log: record barred rounds and the runner-up; invariant number formatting
## Summary

The session debug log (`Library/VoxrDebugLogs/`, written by `Editor/VoxrDebugSessionLog.cs`) cannot show the two things the 2.0.0 scoring rules made important:

1. **A barred round leaves no trace.** When a round's winner missed its first required element, `VoxrCommandParser` skips the span and continues before any command or attempt is built (`Runtime/Commands/VoxrCommandParser.cs:2742-2746`). The log's synthetic `no match` attempt therefore covers two different outcomes, "nothing matched" and "everything matched and was barred", and the log header already has to explain that ambiguity in prose.
2. **Only round winners are logged.** The runner-up's intent and score are never recorded, so the near-miss distribution (how close a losing candidate came) is not measurable from the field.

Also: numbers embedded in `rejectReason` text are formatted with the Editor's current culture (documented at `Editor/VoxrDebugSessionLog.cs:67`), so the log is not machine-readable across locales.

## Proposed change

Bump the log schema version and add to `AttemptDto` (`Editor/VoxrDebugSessionLog.cs`, ~line 307):

- `barred` (bool): true for an attempt recorded from a round whose winner was refused by the leading-required-miss bar. Such a round must now write an attempt (intent, pattern, score, span) with `accepted=false`, `barred=true`, `rejectReason="barred"`, instead of nothing.
- `runnerUpIntent` (string) and `runnerUpScore` (float): the second-ranked candidate of the same round, empty / `-1` when there was none. Distinct from the existing `tiedRival` fields, which only cover exact ties.
- Format every number written into `rejectReason` with `CultureInfo.InvariantCulture`, and update the header text at `:67` accordingly.

The data has to come from `VoxrMatchDiagnostics` (Runtime), which is the only source the Editor log reads, so the runtime diagnostics record gains the same three fields.

## Constraints

- The parser's zero-allocation poll path is pinned by `ZeroAllocPollPathTests`. Recording a barred attempt or a runner-up must not allocate on that path; extend the pooled diagnostics buffers rather than allocating per round. If it cannot be done allocation-free, the fields go behind the existing diagnostics-enabled flag.
- No behaviour change: what fires, and when, is untouched. The bar itself is not modified (ruling 3, 2026-09-06: absolute).

## Acceptance

- Replay a fixture known to be barred (`target hotel one` against the demo grammar: `approach_target` wins and is barred) and assert the log contains an attempt with `barred=true` and no fired command; the existing "no match" synthetic entry is not written for that round.
- A round with two scored candidates records the loser as `runnerUpIntent`/`runnerUpScore`.
- A `rejectReason` produced under a comma-decimal culture contains `0.5`, not `0,5`.
- `Documentation~/scoring.md` "Reading a session log" and the log header describe the new fields; `CHANGELOG.md` under Unreleased.

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), critique item R2, ranked 2 of 12. These two fields are what an audit of the bar's field cost needs.

jinwoo1601:
Closes #144

## What

The Editor session log now records the rounds the leading-required-miss bar refused, and names each round's runner-up.

**Barred rounds.** A round whose winner missed its first required element previously returned before any command or attempt was built, so it left no trace at all. It now publishes its own attempt — `barred: true`, `accepted: false`, `rejectReason: "barred"`, an empty `slots` array, and the intent, pattern and score of the candidate the bar refused — **in the order that round actually ran**. Because such an utterance now carries those attempts, the synthetic `no match` entry is no longer written for one whose every round was barred, so `no match` recovers its plain reading: no round produced a result *and* no round was barred.

**Runner-up.** Every attempt gains `runnerUpIntent` / `runnerUpScore`: the round's second-ranked candidate, what would have won had the winner not been there. Ranked by the same `CompareCandidate` order selection itself uses (earliest start, then score, then consumed span, then literal count) — not by score alone, so "second-ranked" means second by every key.

**Invariant formatting.** Numbers embedded in `rejectReason` now use `FormattableString.Invariant`, so the decimal separator is always `.` whatever the Editor's locale.

`schemaVersion` 2 → 3. All three fields are additive.

## Why this shape

Two design points a reviewer should check rather than take on trust:

1. **Barred rounds ride a parallel array, not `LastParseDiagnostics`.** The i-th `ParseDiagnosticEntry` describes the i-th entry of `_resultBuf`, and both `VoxrCommandRecogniser.BuildAttempt` and `VoxrBatchTestRunner` index it that way. A barred round produces no `VoxrCommand`, so putting it in that array would break the contract. Instead `VoxrCommandParser` gains `BarredRoundEntry[] LastBarredRounds`, each entry carrying `ResultsBefore` — the number of results emitted when the bar fired — which is what lets the recogniser re-interleave them into round order via a head drain inside the Step 7 loop and a tail drain after it.

2. **The runner-up is tracked with the standard best/second-best algorithm.** The outgoing incumbent always outranks the current runner-up (it outranked every candidate seen so far), so promotion on `Better` needs no comparison; every non-`Better` candidate is ranked against the runner-up slot with one extra `CompareCandidate` call. The slot is seeded exactly as the incumbent slot is, so `CompareCandidate`'s own `bestScore <= 0f` → `Better` rule reads `float.MinValue` as "empty".

## The allocation constraint

The issue asked that this not allocate on the pinned poll path, with a fallback of putting the fields behind the diagnostics-enabled flag. Two findings:

- `ZeroAllocPollPathTests` measures `VoxrJsonParser`, not the command parser. The command parser's two alloc pins (`EagerSelection_OverASiblingTie_AllocatesNothingPerCall`, `LeadingMissLatch_AllocatesNothingPerCall`) both measure `TryEagerCommit`, which builds no diagnostics and was **not touched**.
- There is no runtime diagnostics-enabled flag in this codebase — the gate *is* the `UNITY_EDITOR` compile symbol, and the whole `VoxrMatchDiagnostics` apparatus already exists only there. So everything here is `#if UNITY_EDITOR`: a player build allocates nothing new, and the Editor path pays what it already paid. The issue's fallback is satisfied by construction.

The comment above the bar's early return previously asserted "a barred round should allocate nothing"; it has been amended to say that this still holds for a player build and that the Editor now deliberately records one entry.

## Validation

- [x] Compiles clean via host project `VoXR TestGround` (Unity 6000.4.7f1).
- [x] **EditMode 146/146 passed**, 0 failed, 0 skipped (baseline was 133; 13 new tests).
- [x] **PlayMode 596/596 passed**, 0 failed, 0 skipped — unchanged from baseline, no regression.
- [x] 13 new EditMode tests: the barred round is recorded without becoming a result and without entering `LastParseDiagnostics`; `LastBarredRounds` is per-parse and never stale; the runner-up is recorded, is absent when there is one candidate, and does not leak between rounds; runner-up and `tiedRival` are shown naming **different intents on the same round**, proving neither is derived from the other; a barred utterance publishes a barred attempt and no `no match`, fires no command, and still raises `OnUnrecognisedSpeech`; `no match` still happens when nothing matched; barred-then-emitting and emitting-then-barred both come out in round order (the latter is the tail drain, and would fail without it); two consecutive barred rounds sharing one index both appear, in order; the session-log DTO carries and serialises all three fields; `schemaVersion` is pinned at 3.
- [x] Documentation swept for the **outcome**, not the field names — this change falsified 15+ sentences across `scoring.md`, `editor-testing.md`, `troubleshooting.md`, `command-recognition.md`, `api/batch-test-runner.md` and `KNOWN_LIMITATIONS.md` that never mention any new symbol (e.g. "It records no session-log attempt", "Round 2 left no entry at all", "a pattern's absence has two readings"). All corrected.
- [x] No measured number in `Documentation~/scoring.md` was changed. The one figure added to a worked example's JSON block (`"score": 0.67`) restates the `2 / 3` already asserted in that example's own prose.
- [x] `CHANGELOG.md` under `[Unreleased]`; the released `[2.0.0]` entry describing the old behaviour was left as written.
- Quest device check: **not applicable** — Editor-only diagnostics, no runtime or native path touched.

## Worth a reviewer's attention

- **The culture test passed rather than self-ignoring**, which confirms `CultureInfo.CurrentCulture` does reach `float.ToString("F2")` in this Editor runtime — so it has real coverage. But it was **not mutation-verified**: nobody reverted the invariant formatting to confirm it goes red. Treat that as the weakest assertion in the batch.
- **Deliberate omissions, both named in the changelog.** The batch test runner's per-case report still does not surface barred rounds (its rows describe one round per case; a round producing no command has no place in that shape) — though its attempts *do* carry the runner-up fields, so the two copies of the assembly logic stay in step. And a barred attempt's `slots` array is always empty, because the bar refuses above the point where a winner's slot array is built.
- **Barred rounds are recorded on the ordinary parse path only.** The Step 5 follow-up slot-fill path returns with its own single synthetic attempt before Step 7, so a barred round in an utterance that resolved a pending is not recorded. Out of scope here, but a real hole if the audit this issue exists to enable hits it.
- `Runtime/Testing/VoxrBatchTestRunner.cs` `ToCsv` has an unrelated `{r.Score:F2}` interpolation that is a genuine culture hazard in exported CSV (a comma decimal separator would corrupt the column count). It is not a `rejectReason` and was left alone; flagging it as a possible follow-up issue.


jinwoo1601 (commented):
## Independent review — PR #151

**Profile:** slim (issue lane), plus `XFILE` — added because this diff crosses two contract boundaries: a new `internal struct` produced in Runtime and consumed across the `#if UNITY_EDITOR` seam, and a `[Serializable]` DTO whose field names are the exported log's public shape. Angles run: `CLAIM`, `LINE`, `GONE`, `XFILE`.

**Head reviewed:** `4576ce8394eef2dfebb798ca727c0cf89f64f3c9` (base `1c595bf`). Four findings survived dedup and were adversarially verified; three CONFIRMED (one in part), one REFUTED.

The mechanism itself came back clean under line-by-line re-derivation: the runner-up's best/second-best invariant holds inductively, the head and tail drains partition `ResultsBefore` with no gap and no double-emission, Step 6's non-barred branch reproduces the base attempt argument-for-argument, `OnUnrecognisedSpeech` still fires everywhere it fired before, all four parse exits leave `LastBarredRounds` non-null, and all 19 `VoxrMatchAttempt` construction sites still bind correctly. Findings 1–3 are about what the PR and the docs *say*, not what the code does.

---

### 1. The "Editor-only" claim was false — `CHANGELOG.md` — CONFIRMED, **fixed in `2f905af`**

The changelog told users "**Editor-only, and nothing about behaviour moves:** all of it sits behind `UNITY_EDITOR`", and the PR body said "a player build allocates nothing new". Two of the four invariant-formatting edits are in `Runtime/Testing/VoxrBatchTestRunner.cs:196-197` and `:214-215`, inside `PassesThresholds` — and that file's only guarded region is `#if UNITY_EDITOR` at `:255`…`:305`, inside `MakeResult`. `Runtime/Jinwoo1601.VoXR.Runtime.asmdef` has empty `includePlatforms`/`excludePlatforms`/`defineConstraints`, so `public class VoxrBatchTestRunner` compiles into a player build, and the lines are reachable there via the unguarded `Run` → `:133 if (!PassesThresholds(...))`.

The code is fine — the invariant formatting is an improvement and sits off every hot path. The claim was not. Amended to state the exception, including that `FormattableString.Invariant` lowers to `FormattableStringFactory.Create(string, params object[])` and so allocates an `object[]` where the base's plain interpolation used `string.Format(string, object, object)`.

### 2. The doc sweep over-corrected on the follow-up path — `Documentation~/troubleshooting.md:49`, `:70` — CONFIRMED IN PART, not fixed

The base warned that a pattern's absence from the log has two readings — it lost selection, or it won its round and was barred. This PR replaces that with one reading. `troubleshooting.md:70` now says flatly: "if the log holds no attempt naming the pattern at all, it is **neither coverage nor the bar**: the pattern lost selection to something else." Same at `:49`.

That is false on the follow-up slot-fill path, which the PR itself discloses as a known hole. `ParseInternal` runs unconditionally at `VoxrCommandRecogniser.cs:719`; the Step 5 arms at `:833-851` and `:877-889` publish one synthetic attempt and return, discarding that parse's barred rounds. A barred round emits no result, so it cannot set `hasCompleteNewCommand`, and the arm's guard is unobstructed — the case is reachable. `troubleshooting.md` contains no mention of "follow-up" or "ordinary parse path" to scope the claim.

**Two sub-claims refuted, and the finding is narrower than first reported.** `scoring.md:520` is *not* affected: it sits under "On the ordinary parse path", and the pre-existing paragraph at `:522` already tells the reader the follow-up entries publish a single synthetic attempt instead. And the symptom-table row was not deleted — it was rekeyed in place (base `:551` → head `:559`), with `:561` still serving a reader who observes no attempt.

Fix would be to re-scope the two `troubleshooting.md` sentences to the ordinary parse path.

### 3. `Match n/N` is the wrong header for a lone barred round — `Documentation~/editor-testing.md:24` — CONFIRMED, not fixed

The new sentence promises, unhedged, that a barred round "gets its own `Match n/N` block". `Editor/VoxrDebugWindow.cs:389-392` emits that header only when `diag.Attempts.Length > 1`; otherwise it draws "Last Match". An every-round-barred utterance goes through Step 6 and publishes exactly one attempt — the PR's own test asserts it (`Tests~/Editor/VoxrCommandRecogniserDiagnosticTests.cs:399`). So the canonical repro, the case a reader reproduces first, renders under "Last Match".

The rest of that sentence checks out against `VoxrDebugWindow.cs:397, 408-412, 421-425, 459-460, 163-164`: `[FAIL]`, the pattern and score lines, `Rejected: barred`, no slots section, and `<intent>: barred` in Match history are all correct.

---

### Refuted, so you can see what was considered and killed

- **"The issue asked for a `span` and the diff does not record one."** Refuted. Issue #144 names and types every field it wants added (`barred` bool, `runnerUpIntent` string, `runnerUpScore` float); "span" appears with neither, inside the `barred` bullet, as a gloss on the noun *attempt* — and its acceptance criteria assert no span. Decisively: **no** attempt has ever carried a round span, on base or head, so the parenthetical describes the round rather than prescribing a fourth field. `StartIdx`/`EndIdx` are consumed into `aggregateConfidence`, which the log header already documents as the minimum over the matched span, giving a barred attempt exact field parity with an emitting one.

### Notes — unverified, below the materiality floor

- `VoxrCommandRecogniser.cs:1110`'s `entered pending (awaiting disambiguation, {choices.Length} choices)` is the one reject-reason number not converted. Harmless — a small non-negative `int` formats identically under every culture — but it makes "every number" fractionally wider than the code.
- `Editor/VoxrDebugWindow.cs:162, 483` and `VoxrBatchTestWindow.cs:268` still format culture-sensitively. Display-only, not `rejectReason`, out of scope — as is `VoxrBatchTestRunner.ToCsv`'s `{r.Score:F2}`, which the PR already flags as a possible follow-up issue.
- No test uses the `approach_target` demo-grammar fixture the issue's acceptance text names; an equivalent fixture is used instead.
- The runner-up docs say the fields are empty/`-1` "when the round had a single candidate", where the code means a single *admissible* candidate. `scoring.md` defines admission, so this reads correctly there; the log's own header const does not.
- `DrawThresholdLine` will render a barred attempt's score as clearing its threshold while the intent line reads `[FAIL]`.
- The three new DTO fields were inserted before `slots`, changing JSON key order. Immaterial for JSON readers.

No Unity build or test run was possible from here. The PR's **EditMode 146/146** and **PlayMode 596/596** results are an author claim this review could not re-run; what was checked is that the diff supports the arithmetic — 13 added `[Test]` methods (6 / 5 / 2), and 133 + 13 = 146. The culture test's own red state remains unverified by mutation, as the PR itself states. On-device Quest behaviour was not checked and is not applicable to this diff.

jinwoo1601:
Pushed `2f905af` addressing finding 1 (the false "Editor-only" claim in the changelog). Findings 2 and 3 are left unfixed for your ruling.

jinwoo1601:
Pushed `59a6cf2`, addressing findings 2 and 3.

**F2** — `troubleshooting.md:49` and `:70` now scope the absence claim to the ordinary parse path and name the exception: an utterance that resolved a pending command (a confirm or cancel word, a disambiguation answer, a follow-up slot-fill) publishes one synthetic entry for the whole parse and discards every round of it, barred or not, so an absence there says nothing. `editor-testing.md:24` carried the same unscoped claim about the same window — the one `troubleshooting.md:49` links to — so it is scoped too; leaving it would have made the fix self-inconsistent.

The refuted half is untouched: `scoring.md:520` is already scoped by "On the ordinary parse path", and the symptom row was rekeyed in place rather than deleted.

**F3** — `editor-testing.md:24` no longer promises a `Match n/N` header. It now names both, and says a lone barred round — the canonical repro — renders under `Last Match`, which is what `VoxrDebugWindow.cs:389-392` does when `Attempts.Length == 1`.

Documentation only. No `.cs` file has changed since the verified run, so EditMode 146/146 and PlayMode 596/596 still stand.

All three findings from the review are now addressed. The underlying code hole behind F2 — barred rounds not recorded on the follow-up slot-fill path — is unchanged and remains documented in the PR body as a known limitation; closing it would be a separate issue.

PR #152, opened 2026-09-06
Re-extract the model when the archive's contents change (#145)

### LINKED ISSUE #145: ModelExtractor ignores a changed model or model.conf on existing installs
## Summary

`ModelExtractor` never re-extracts a model whose contents changed. On any device that has already run the app, a new model zip under the same file name, or the same model with an edited `conf/model.conf`, is silently ignored.

## Where

`Runtime/ModelExtractor.cs`:

- `:24-26` — the cache key is `Path.GetFileName(modelRelativePath)` joined to `persistentDataPath/VoxrModels`. Nothing about the zip's contents enters the key.
- `:33-36` — if the directory exists and `ValidateModelDirectory` passes, the existing extraction is returned and the zip is never opened.
- `:118-128` — `ValidateModelDirectory` checks only that `am/final.mdl` and `conf/mfcc.conf` exist and `graph/` is a directory. `conf/model.conf` is not checked, so an absent or corrupt decoder config also passes.

## Why it matters now

`conf/model.conf` is the decoder's live configuration (beam, lattice beam, endpointer trailing-silence rules). Any tuning shipped through it, and any model upgrade, is a no-op for existing installs. Field data collected after such a change would be attributed to the new configuration while the device still runs the old one.

## Proposed change

- Write a stamp file (for example `.voxr-model-stamp`) into the extracted directory at extraction time, containing a version key derived from the source zip. `ReadStreamingAsset` already loads the whole zip into a `byte[]`, so a content hash of those bytes costs one pass over memory the code already holds. Package version alone is not enough: the same package version can ship with a different model.
- On start: compute the key for the source zip, compare with the stamp; on mismatch (or missing stamp) delete and re-extract, using the existing temp-directory-then-rename path.
- Add `conf/model.conf` to `ValidateModelDirectory`.
- Design the key so it also covers a model obtained by download rather than from StreamingAssets (ruling 6, 2026-09-06: shipped builds will download models on demand at a language-selection screen). A hash of the archive bytes satisfies both sources.

## Acceptance

- Test: extract; modify `conf/model.conf` inside a copy of the source zip; run again; assert the extracted `conf/model.conf` changed and the stamp updated.
- Test: unchanged zip does not re-extract (stamp matches; directory mtime unchanged).
- Test: a directory missing `conf/model.conf` fails validation.
- Existing installs upgrade cleanly: a missing stamp triggers one re-extraction, not an error.
- `Documentation~/getting-started.md` (model setup) notes the stamp and the re-extraction on model change; `CHANGELOG.md` under Unreleased (Fixed).

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), critique verification (c) and item R9, ranked 3 of 12. Hard prerequisite for any `model.conf` tuning.

jinwoo1601:
Closes #145

## What

The extracted-model cache is now keyed on the **contents** of the source archive, not just its file name.

- `ExtractModelAsync` reads the archive on every launch and computes a SHA-256 hash of its bytes (`ComputeArchiveKey`, run on a thread-pool thread via `Task.Run`).
- Extraction writes that key into a `.voxr-model-stamp` file inside the extracted model directory — into the **temp** directory, before the atomic rename, so the existing "an interrupted extraction cannot leave a half-written cache" guarantee still holds and a stamped cache can only ever appear complete.
- A cached extraction is reused only when it passes structural validation **and** its stamp matches the current archive. On mismatch, or with no stamp, the model is re-extracted through the same temp-then-rename path — and the cache already on the device stays in place as the working model until the replacement has been extracted, validated and stamped.
- `ValidateModelDirectory` now also requires `conf/model.conf`.

Testability: `ExtractModelAsync` is split into a wrapper that keeps today's exact signature and resolves the Unity-owned paths, plus an `internal` seam overload taking `(modelName, basePath, Func<Task<byte[]>> archiveSource, archiveDescription, onError)`. No test targeted the extraction path *directly* before this — the existing `ModelExtractor` tests all covered the pure `ValidateModelDirectory`, and the extractor's only other coverage was incidental, through `VoxrSpeechRecogniserOwnershipTests`.

## Why

`conf/model.conf` is the decoder's live configuration (beam, lattice beam, endpointer trailing-silence rules). The cache key ignored archive contents entirely: if the directory existed and validated, the archive was never opened. So any decoder tuning shipped through `model.conf`, and any model swap under the same file name, was a **no-op on every device that had already run the app** — and field data gathered after such a change would be attributed to a configuration the device was not running. This is a hard prerequisite for tuning `model.conf` at all.

A hash of the archive bytes was chosen over cheaper keys deliberately: package version is insufficient (the same version can ship a different model), and a byte hash is the one key that works identically for a model shipped in StreamingAssets and one obtained by download.

## Behaviour changes worth reviewing

1. **A `conf/model.conf`-less archive now fails validation** and raises `ModelLoadFailed` where it previously loaded. Stock VOSK models all ship the file (verified against the host project's real `vosk-model-small-en-us-0.15.zip`), but this is a real tightening.
2. **Existing installs pay exactly one re-extraction** on the first launch after upgrading — no stamp reads as a mismatch. Not an error, does not repeat.
3. **Startup cost:** the archive (~40 MB for the small model) is now read and hashed on **every** launch rather than only the first. Only the *hash* is reliably off the main thread. The read joins it on a thread-pool thread in the Editor and in standalone builds; on Android it does not — the archive comes through `UnityWebRequest`, which is polled on the main thread and materialises the whole array there. The transfer itself is asynchronous, so the main thread is not blocked for its duration, but the polling and that array copy are main-thread work. See the unchecked box below.
4. **A cache is never surrendered before its replacement exists.** A stamp mismatch does not delete the old extraction up front; the delete happens immediately before `Directory.Move`, once the replacement has been extracted, validated and stamped. A corrupt replacement archive therefore costs an error rather than the model the device already had.
5. **An unreadable archive falls back to a valid cache** — including a read that *throws* (locked or permission-denied file, OOM materialising the bytes), not only one that is absent. `ModelLoadFailed` is raised only when there is no usable cache, and the two failures stay distinguishable in the message: `Model archive not found in StreamingAssets: <name>.zip` versus a "could not be read" message naming the underlying reason. That path also deletes an invalid cache and sweeps a stale `.tmp_` directory before raising, as the pre-stamp code did.

## Review

`review-pr` (slim profile plus `PIT`) ran against `c6ff8e6` and posted five CONFIRMED findings. The maintainer ruled that all five be fixed; every one is addressed in this branch. Items 4 and 5 above and the `Documentation~`/`CHANGELOG` threading wording are the result.

## Validation

- [x] EditMode suite: **147/147 pass** (baseline on `main` 146; +1 new test)
- [x] PlayMode suite: **606/606 pass** (baseline on `main` 596; +10 new tests)
- [x] Both run via the host project `VoXR TestGround` (Unity 6000.4.7f1), per the project's verification bindings
- [x] Ten new PlayMode tests cover the issue's acceptance criteria and the review fixes: a changed archive re-extracts and restamps; an unchanged archive leaves the cache untouched (witnessed by a sentinel file, not a directory mtime); a missing stamp re-extracts exactly once; a missing archive with a valid cache returns the cache silently; a missing archive with no cache still raises `ModelLoadFailed`; a corrupt archive leaves neither a published nor a temp directory behind; **a corrupt archive with a valid cache leaves that cache intact, valid and still stamped**; **a throwing archive source with a valid cache returns the cache with no error**; **a throwing source with no cache raises `ModelLoadFailed` with the "could not be read" message**; **a missing archive with an invalid cache sweeps both that cache and a stale temp directory**
- [x] `Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.FailedInitialise_ReleasesClaimForNextRecogniser` depends on the old `Path.GetFileName` keying and on the exact missing-archive message; both are preserved and it still passes
- [ ] **On-device (Quest) check needed.** This code runs on device and `ReadStreamingAsset` has an Android-specific `UnityWebRequest` branch that cannot be exercised from WSL. Three things need a headset: the main-thread cost of polling and materialising the ~40 MB array out of the APK on **every** launch (item 3 — unmeasurable here), peak memory at that array copy, and the one-time upgrade re-extraction on a device holding an existing pre-stamp cache.


jinwoo1601 (commented):
## Independent review — PR #152

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`, **plus `PIT` added** because the diff reorders `await`s, introduces a `Task.Run` hashing step, converts `ExtractModelAsync` from `async Task<string>` to a non-async wrapper, and adds six async `[UnityTest]`s.
**Head reviewed:** `c6ff8e60da9d0efd1f6bd41db6493e60a10acc4e` · **merge-base:** `d437156`.

Five numbered findings, all CONFIRMED by adversarial verification. **No blockers**, so nothing was fixed on the branch — all five are posted unruled. Three majors cluster on one root cause: moving the archive read above the cache check changed what happens on a *cache-hit* launch, and the PR body and docs describe the intended behaviour rather than the implemented one. The core fix — hash the archive, stamp the extraction, re-extract on mismatch — is correct, and every acceptance criterion in #145 is met.

---

### 1. A stamp mismatch destroys a valid cache before its replacement exists — `Runtime/ModelExtractor.cs:73-76` · major · CONFIRMED

```csharp
if (ValidateModelDirectory(finalPath) && ReadStamp(finalPath) == sourceKey)
    return finalPath;

Directory.Delete(finalPath, true);
```

At base this delete was reachable only after `ValidateModelDirectory` returned **false** — the deleted cache was already worthless. Adding the stamp term makes it fire on a cache that validates perfectly, and it fires before the unzip (`:83`) and before `ValidateModelDirectory(tempPath)` (`:116`). From `:76` until `Directory.Move` (`:131`) the device holds no model. If the replacement archive is corrupt, the temp is deleted and `ModelLoadFailed` raised — with the working extraction already gone, where base kept serving it. Per invariant 3 **every existing install walks this window once** on upgrade. Verifier confirmed nothing restores it: the failure branch and the outer `catch` touch only `tempPath`. Fix: defer the delete to immediately before `Directory.Move`. Verified as mechanically safe — extraction never references `finalPath`; the cost is peak disk holding both copies.

### 2. The unreadable-archive fallback covers only a `null` return, not a throwing read — `Runtime/ModelExtractor.cs:52-65` · major · CONFIRMED

The read is now the first statement in the `try`, and only its null shape is guarded. `ReadStreamingAsset` returns null for exactly two checks (`!File.Exists`; `request.result != Success`); everything else propagates — `File.ReadAllBytes` on a locked or permission-denied file (`:219`), OOM materialising ~40 MB from `downloadHandler.data` (`:212`). Those unwind to the outer `catch` (`:134`), raising `ModelLoadFailed` and returning null **on a launch where `finalPath` holds a valid, correctly-stamped cache**. At base the read sat after the cache early-return, so this was structurally impossible. This contradicts invariant 7(a) and three written claims: PR note 4, `getting-started.md:75` ("missing, **or unreadable**"), `CHANGELOG.md:22`. No test covers it — the seam is driven by `Task.FromResult`, which cannot throw, while `MissingArchive_WithValidCache_ReturnsCacheWithoutError`'s own message calls that case "An unreadable archive" (`ModelExtractorCacheTests.cs:184`). Fix: degrade a *failed* read to the same fallback, or narrow the claims to "absent".

### 3. Shipped docs claim the per-launch read is off the main thread; on Android it is not — `Documentation~/getting-started.md:166`, `CHANGELOG.md:22` · major · CONFIRMED

"Otherwise it only reads **and hashes** the archive to confirm the cache is current, off the main thread." Only the hash is (`:67`), plus the desktop read (`:219`). The Android branch runs `UnityWebRequest.Get` / `SendWebRequest` on the calling thread — the wrapper is no longer `async`, so the lambda executes before the first suspension — then `await Task.Yield()` posts back to `UnitySynchronizationContext`, so the poll loop *and* the `downloadHandler.data` copy that materialises ~40 MB (`:212`) are main-thread work, now on **every** launch rather than once. Android/Quest is this package's primary target (`troubleshooting.md:11`). Base made no threading claim at all — this one is introduced here. The deferred on-device box asks about latency; it should also cover main-thread stall and peak memory at `:212`.

### 4. The missing-archive path no longer sweeps a stale temp or deletes an invalid cache — `Runtime/ModelExtractor.cs:52-65`, `:80-81` · minor · CONFIRMED

Base order was: cache check → `Directory.Delete(finalPath, true)` → stale-temp sweep → read archive → error if null. So a missing archive still cleaned both before erroring. At head the null branch returns at `:60`/`:64`, above the sweep at `:80-81`, so an unreadable archive now leaves an invalid cache directory and a partial `.tmp_` (up to ~110 MB unpacked) in place. The verifier rejected the "base skipped it too" defence: base returned early only for a *valid* cache, and never reached the missing-archive error with one. Relatedly `getting-started.md:71` — edited in this diff — still promises the SDK "deletes it and re-extracts immediately, within the same call"; the delete half is now false when the archive is unreadable. Reachability is low (interrupted extraction *and* an unreadable archive), hence minor.

### 5. "The extraction path had no test coverage at all before this" is false — PR body · minor · CONFIRMED

The body's own validation box contradicts it: `VoxrSpeechRecogniserOwnershipTests.FailedInitialise_ReleasesClaimForNextRecogniser` depends on `ExtractModelAsync`'s keying and asserts the error the extractor raises. Its `[UnitySetUp]` also drives a full successful extraction that all four tests in the class depend on. Accurate wording: no test targeted the extraction path *directly*. Not a nit — the mis-statement erases the one pre-existing test that constrains this refactor.

---

### Notes (unverified, no action implied)

- `MissingStamp_ReExtractsExactlyOnce:157-164` — the third phase asserts only the sentinel, not `steadyState.Result` or `_errors`; the two comparable phases assert all three.
- `ChangedArchive_ReExtractsAndUpdatesStamp` is the only new test that never asserts `_errors` is empty.
- The `entry.LastWriteTime` pin in `WriteEntry` is never exercised: every test needing two equal hashes reuses one `byte[]`, so "identical contents hash identically" is asserted nowhere.
- The seam's error text hardcodes "in StreamingAssets" though the seam exists partly to serve a downloaded archive.
- `VoxrSpeechRecogniserOwnershipTests.cs:185`'s comment still describes the key as `Path.GetFileName`, now only half of it.
- The ~40 MB `byte[]` stays rooted across the hash and the cache-hit return — an LOH allocation on every launch.

### Not verifiable from here

On-device (Quest) behaviour of any kind: the Android `UnityWebRequest` StreamingAssets branch, real cold-start cost of the new per-launch read+hash, main-thread stall at `downloadHandler.data`, and the upgrade re-extraction against a real pre-stamp cache. The PR's EditMode 147 / PlayMode 602 counts were not re-run by this review; the six named tests were confirmed to exist and to assert what the PR says they assert.

*Reviewed by Claude Code — finder agents inherited no context from the session that wrote the diff.*

jinwoo1601:
Pushed `cd98656`, which addresses **all five** findings from the review above, per the maintainer's ruling to fix all rather than blockers only.

| # | Finding | Fix |
|---|---|---|
| 1 | Stamp mismatch destroyed a valid cache before its replacement existed | `Directory.Delete(finalPath, true)` moved out of the cache-check and down to immediately before `Directory.Move`, so the old extraction stays the working model until the replacement is extracted, validated **and** stamped |
| 2 | Fallback caught only a `null` return, not a throwing read | The archive fetch has its own `try`/`catch`; a throw records `readFailure` and falls through to the same valid-cache fallback. The missing-archive message is byte-identical; a failed read gets a distinct "could not be read" message naming the reason |
| 3 | "Off the main thread" false on Android | `getting-started.md`, `CHANGELOG.md` and the PR body now say the *hash* is always off-thread, the read joins it only in Editor/standalone, and on Android the `UnityWebRequest` poll and the ~40 MB array copy are main-thread work |
| 4 | Missing-archive path stopped sweeping a stale temp / invalid cache | That branch deletes an invalid `finalPath` and a stale `.tmp_<modelName>` before raising, restoring the pre-stamp hygiene |
| 5 | "No test coverage at all before this" was false | PR body corrected to "no test targeted the extraction path *directly*", naming the incidental coverage through `VoxrSpeechRecogniserOwnershipTests` |

Four new PlayMode tests pin the behaviour fixes: `CorruptArchive_WithValidCache_KeepsExistingModel` (decisive for #1 — fails on the previous commit), `ThrowingArchiveSource_WithValidCache_ReturnsCacheWithoutError` and `ThrowingArchiveSource_WithoutCache_RaisesModelLoadFailed` (#2, one using a synchronous throw and one a faulted task, since production only ever produces the latter), and `MissingArchive_WithInvalidCache_SweepsCacheAndTemp` (#4).

Re-verified on the host project: **EditMode 147/147, PlayMode 606/606**, no compile errors, no hang. The six pre-existing cache tests all still pass under the new delete ordering.

Two consequences of fix 1 worth naming rather than burying:

- Peak disk now briefly holds the old cache **and** the new temp at once. That is the price of not surrendering a working model to an unproven replacement.
- A corrupt archive fetched over a pre-existing cache now leaves that cache in place even when the cache itself was invalid — the temp-validation failure returns before the pre-move delete. Strictly better than destroying a valid one, and it is swept on the next launch with a readable archive, or immediately on the unreadable-archive path.

The `### Notes` list from the review is untouched — those were marked unverified with no action implied. The on-device Quest check remains open and is unchanged by this push.

PR #155, opened 2026-09-08
Serve an unchanged model cache without reading the archive (#153, #154)

### LINKED ISSUE #153: Cached launches still read and hash the 39 MB archive: add a cheap source-identity tier to the model stamp
## Problem

#145 (PR #152) made the extracted-model cache correct: it keys on a SHA-256 of the archive bytes, stamped into `.voxr-model-stamp`, so a changed archive is never silently served stale. That correctness is proven on hardware and is **not** in question here.

But the fix hoisted the archive read above the cache check, so **every launch now reads and hashes the ~39 MB model archive**, even when the cache is valid and nothing has changed. Before #145, a valid cache meant the archive was never opened.

## Measured cost (Quest 3, Horizon OS / Android 14, 90 Hz, 2026-09-07/08)

| | ms |
|---|---|
| **Time-to-voice-ready, cached launch** | **1303.2** |
| Archive read + SHA-256 | **~719** cold / ~614 warm |
| — of which main thread (UWR poll + `.data` copy) | ~167 |
| — of which off-thread (SHA-256, `Task.Run`) | ~552 |
| `vosk_bridge_init` (derived two independent ways) | ~547–584 |

**The always-hash read is ~54 % of time-to-voice-ready and roughly doubles it (~584 ms → ~1303 ms, ≈2.2×).**

Also measured: managed heap **+43 MB** for the byte array, not released without a GC (1.9 MB → 44.9 MB, still 44.7 MB after release). And the read blocks the main thread — 1 frame advanced across 147.8 ms of polling, `max_frame_dt = 177.2 ms` (~16 dropped frames). That stall is tracked separately; see the companion issue.

Caveat that must travel with these numbers: Development build (required, `Profiler.GetMonoUsedSizeLong()` returns 0 in release), so the timings are an **upper bound**. Measured in the `VoXR TestGround` host project, but the archive is byte-identical so the work is the same.

## The invariant, restated

The invariant wanted is *"a changed archive is never silently served stale."* Always-hash was one **implementation** of it, not the invariant itself. There is a cheaper implementation that preserves it.

## Proposed design — two-tier source identity

The stamp gains a second line:

```
sha256:<hash of archive bytes>
src:<cheap source-identity token>
```

Token: `base.apk` path + size + mtime on Android; archive file length + mtime on Editor/standalone.

On launch:

1. Compute the cheap token — a `stat`, **no archive read**.
2. **Token matches** → source provably unchanged → serve the cache. **Zero archive I/O.**
3. **Token differs or is missing** → full read + SHA-256 exactly as today.
   - hash **matches** → rewrite the token line only, **no re-extraction**
   - hash **differs** → re-extract

The failure direction is **over-invalidation, never under**: a token that moves without the bytes changing costs one extra hash; a token cannot fail to move when the bytes change, because the archive lives inside the APK and its bytes cannot change without an install.

**No user-facing configuration.** The original "no opt-out knob" ruling stands — this is not a knob, it is a cheaper way to answer the same question.

Expected effect: typical cached launch **1303 ms → ~584 ms**.

## Feasibility, already verified on device (both PASS)

- The app's own UID can `stat` **and** read `base.apk` (`-rw-r--r--`; `run-as` returned `PK\003\004`).
- Reinstalling a **byte-identical** APK moves the token: path `~~ZCyRryo7…` → `~~7Q-TpXFBt9…`, mtime `1788786682` → `1788869785`.
- `versionCode` is **not** a viable token — proven on device: a freshly rebuilt APK and the May 2026 build both reported `versionCode=1`, `versionName=0.1.0`.

## Implementation notes

- `Runtime/ModelExtractor.cs` is the only runtime file this should touch.
  - `StampFileName` :22 — the stamp format changes, so a one-line stamp from #145 must read as "token missing" and take the hash path once, then be rewritten in the two-line form. No re-extraction in that case (the hash will match).
  - `ExtractModelAsync` :38–180 — the archive read at :57 must move back *below* the cache check for the token-hit path.
  - `ReadStamp` :210 currently `Trim()`s the whole file and compares the whole string; it needs to parse two lines.
  - `ComputeArchiveKey` :197 is unchanged.
- The testable overload injects the archive as `Func<Task<byte[]>> archiveSource` (:41). The source-identity token needs a parallel seam so the token path is unit-testable without a device — worth deciding during design, not here.
- Nothing in either test suite exercises the Android `ReadStreamingAsset` path, so the token computation on Android cannot be covered by an automated test. Plan for that: unit-test the Editor/standalone token and the stamp parse/upgrade logic; the Android branch stays human-verified.

## Acceptance

- A cached launch with an unchanged APK performs **zero** archive reads (no `UnityWebRequest`, no 39 MB allocation).
- A changed archive still re-extracts — the existing #145 guarantee must not regress. The device rehearsal to reuse: change `conf/model.conf`'s `--endpoint.rule2.min-trailing-silence` 0.5 → 0.35 inside the archive and ship it over a valid stamped cache.
- A #145-era one-line stamp upgrades in place without re-extracting.
- A moved token with an unchanged hash rewrites the stamp and does **not** re-extract.

## Why it matters beyond start-up

#145 is infrastructure for two committed items: the **R11 endpointer sweep** (which tunes `conf/model.conf` *inside* the archive) and **download-on-demand models**. Both need the change-detection to be correct; neither needs it to cost 719 ms per launch.

Ruled by the maintainer 2026-09-08 on the measured evidence above. This is a new issue, not a reopening of #145 — #145's fix is correct and stays.

### LINKED ISSUE #154: Archive read blocks the main thread: Task.Yield() in ReadStreamingAsset never releases the frame
## Problem

`ModelExtractor.ReadStreamingAsset` polls the `UnityWebRequest` like this (`Runtime/ModelExtractor.cs` :230–243, the Android branch):

```csharp
using var request = UnityWebRequest.Get(fullPath);
var operation = request.SendWebRequest();

while (!operation.isDone)
    await Task.Yield();
```

`await Task.Yield()` reposts the continuation to Unity's synchronisation context, which runs it again **within the same frame** rather than releasing the frame. The loop therefore spins on the main thread instead of yielding it.

## Evidence (Quest 3, Horizon OS / Android 14, 90 Hz)

Measured by an on-device probe replicating this exact code path:

| | COLD | WARM |
|---|---|---|
| `segA_poll_ms` | 147.8 | 31.6 |
| `segA_data_copy_ms` | 19.4 | 17.6 |
| `segA_frames_advanced` | **1** | 3 |
| `segA_max_frame_dt_ms` | **177.2** | 11.9 |

**One frame advanced across 147.8 ms of polling**, with a 177 ms max frame delta — roughly 16 dropped frames at 90 Hz. In VR that is a visible hitch, not a statistic.

For contrast, the SHA-256 in the same probe runs under `Task.Run` and advanced 35–51 frames during its 552 ms, so it delays readiness without freezing rendering. The stall is specific to this poll loop.

## Why this is separate from the always-hash cost

The companion issue removes the archive read from the *typical* launch. This defect is about how the read behaves *when it does happen* — first launch, an upgrade, and every re-extraction. Those paths remain, so the stall remains, and it would be worth fixing even if the archive were read on every launch forever.

## Proposed fix — **hypothesis, not measured**

A coroutine (`yield return operation`) or a genuine per-frame yield should release the frame and remove the freeze **without changing wall time** — the download takes as long as it takes; the question is only whether the main thread is held while it does.

This has **not** been verified. The on-device probe rig can test it: instrument `frames_advanced` and `max_frame_dt` across the read the same way, and compare against the numbers above.

Converting to a coroutine means the method stops being `async Task<byte[]>`, which the `Func<Task<byte[]>> archiveSource` seam at :41 depends on — so the fix needs a shape that keeps that seam testable. Worth settling during design.

## Acceptance

- The archive read advances frames roughly in proportion to its duration (not 1 frame in 148 ms), with `max_frame_dt` back near the 11 ms frame budget.
- Measured on device with the same probe method, before and after — a green test suite is not evidence here; nothing in either suite exercises the Android `ReadStreamingAsset` path.
- Wall-clock read time does not regress.

Split out from the #145 always-hash ruling on 2026-09-08 at the maintainer's direction: distinct defect, distinct fix.

jinwoo1601:
Closes #153
Closes #154

## What

The model cache's freshness check becomes two-tier. `.voxr-model-stamp` gains an optional second line:

```
sha256:<lowercase hex>
src:<container path>|<byte length>|<last-write UTC ticks>
```

On launch the SDK computes the `src:` token with a single `stat`. If it matches what the stamp records — and the stamp also carries a hash — the cache is served and **the archive is never opened**. Otherwise it falls back to the full read + SHA-256 exactly as #145 left it; if the hash still matches, the token line is refreshed in place and **nothing is re-extracted**. Only a hash mismatch re-extracts.

The container is the file the archive lives in, derived from `Application.streamingAssetsPath` (`jar:file://…!/assets` on Android → the APK; the `.zip` itself in the Editor and standalone) rather than hardcoded to `base.apk`. That parse is a pure helper, `ResolveArchiveContainerPath`, with EditMode coverage.

`ReadStreamingAsset`'s Android poll — `while (!operation.isDone) await Task.Yield();` — is replaced with an `AsyncOperation.completed` callback awaited through a `TaskCompletionSource`.

No public API is added, and no user-facing configuration: the "no opt-out knob" ruling on #145 stands. The internal seam gained a sixth **optional** parameter, so all ten pre-existing cache tests compile unchanged and now pin the no-token path.

## Why

#145's fix is correct and stays. But it hoisted the archive read above the cache check, so every launch read and hashed the ~39 MB archive. Measured on a Quest 3 (Horizon OS / Android 14, 90 Hz, Development build so these are an upper bound):

| | ms |
|---|---|
| Time-to-voice-ready, cached launch | **1303.2** |
| Archive read + SHA-256 | ~719 cold / ~614 warm |
| `vosk_bridge_init` (derived two ways) | ~547–584 |

That is **~54 % of time-to-voice-ready**, a ≈2.2× slowdown, plus +43 MB of managed heap not released without a GC.

The invariant #145 exists for is *"a changed archive is never silently served stale."* Always-hash was one implementation of it, not the invariant. The token is a cheaper implementation with the same guarantee, because the archive lives inside the APK and its bytes cannot change without an install that replaces the container.

The failure direction is one-sided by construction: it **over-invalidates, never under**. A token that moves while the bytes did not costs one hash and never a re-extraction — a reinstall of a byte-identical build is exactly that case, and is device-verified to move both path and mtime. `versionCode` was tested as a cheaper identity and rejected: a rebuilt APK and a four-month-old build both reported `versionCode=1`.

Separately, `Task.Yield()` reposts its continuation to Unity's synchronisation context, which reruns it inside the same frame instead of releasing it — so the poll was a stall dressed as a yield. Measured at **1 frame advanced across 147.8 ms**, max frame delta 177.2 ms (~16 dropped frames at 90 Hz).

## Validation

- [x] Unity **EditMode: 153/153 pass** (147 → 153, +6). New fixture `ModelExtractorSourceTokenTests` discovered as `runstate="Runnable"`, `testcasecount="6"` — not silently skipped.
- [x] Unity **PlayMode: 612/612 pass** (606 → 612, +6). `ModelExtractorCacheTests` 10 → 16.
- [x] Both run in host project `VoXR TestGround` (Unity 6000.4.7f1), fresh results directory, `start-time` verified so no stale XML was read.
- [x] Zero-archive-I/O is pinned by a counting archive source asserting `Reads == 0`, not merely by the cached path coming back.
- [x] The legacy one-line stamp upgrades in place with no re-extraction (sentinel survives).
- [x] #145's guarantee is re-pinned: changed bytes still re-extract.
- [x] No `NativeBridge~/` change, so no bridge rebuild.

### Not verified — please read before merging

- [ ] **The #154 fix has zero behavioural coverage.** The rewritten branch is guarded by a runtime `Application.platform == RuntimePlatform.Android` check, so neither suite ever enters it — both gates prove only that it type-checks. On-device Quest verification is human-only.
- [ ] **No "after" measurement exists for #154.** The 147.8 ms / 177.2 ms figures describe the *old* code. That the completion callback actually releases the frame is a well-founded expectation, not a measurement. The probe rig used for #153 can test it.
- [ ] **Tier 1 on device is unexercised here** for the same reason — the ~584 ms figure is derived (1303 − 719), not observed after this change.
- [ ] **Split Application Binary / Play Asset Delivery is untested.** Unity builds `streamingAssetsPath` by concatenating `jar:file://` with `Application.dataPath` (confirmed as a string constant in `libunity.so`), so on an ordinary build the derived container and `dataPath` coincide. Whether it tracks the real container when StreamingAssets ship outside the main APK is unconfirmed; if it does not, a changed archive delivered that way might not move the token. Recorded in `KNOWN_LIMITATIONS.md` as untested rather than as safe or broken.

## Docs

An independent sweep drove these; `#145` is still under `[Unreleased]`, so its "the archive is read and hashed on every launch" passage was **amended rather than contradicted** — that cost will never ship.

- `CHANGELOG.md` — #145 entry rewritten to the as-shipped behaviour and re-linked to `#145, #153`; new scoped `Fixed` entry for #154 that gives the old figures as the defect and states plainly that no after-measurement exists.
- `Documentation~/getting-started.md` — the two-tier description, the stamp format (the `src:` token documented as **opaque**), the corrected "a **hash** mismatch takes the same path", the rewritten `Initialise()` cost bullet, and a new paragraph on how to force a re-extraction plus the trap that hand-edits to the *extracted* copy are invisible to both tiers.
- `KNOWN_LIMITATIONS.md` — one entry covering the residual ~19 ms main-thread copy and ~43 MB heap growth on the launches that still read, the deliberate reinstall over-invalidation, and the untested delivery mode.

`README.md:201` and `Documentation~/editor-testing.md:88` were checked and deliberately left alone — they describe `vosk_model_new`, not the archive read.


jinwoo1601 (commented):
## Review — PR #155

**Profile:** slim (issue lane), **plus `PIT`** — added because the diff rewrites an async continuation on Unity's synchronisation context and moves main-thread-only Unity statics. Angles run: `CLAIM`, `LINE`, `GONE`, `PIT`. None dropped.
**Head reviewed:** `154ae671d90ca98a417963bb84cab0fa5d356eec` · **merge-base:** `4eb034b`
Finder agents inherited nothing from the sessions that wrote the code; every numbered finding below was then adversarially verified by a separate agent. Two survived that verification with their framing corrected, and one was killed outright.

---

### 1. BLOCKER — a tier-1 hit fires `OnModelReady` *inside* `Initialise()`, breaking a documented contract

`Runtime/ModelExtractor.cs:81` returns before any `await`, so `ExtractModelAsync` completes synchronously and hands back a finished `Task`. `await` on it takes the `IsCompleted` fast path, so `VoxrSpeechRecogniser.InitialiseAsync` never suspends — and `vosk_bridge_init` (`VoxrSpeechRecogniser.cs:222`, a direct P/Invoke in the `#else` branch a Quest player build compiles) plus `OnModelReady?.Invoke()` (`:229`) both run inside the caller's stack.

`Documentation~/api/speech-recogniser.md:74` and `:77` document `Initialise()` and `StartRecognition()` as "Fire-and-forget async wrapper."

**Failure:** `recogniser.Initialise(); recogniser.OnModelReady += Handler;` — safe by construction before this PR, because the archive read always yielded — now silently never fires the handler, and `IsModelReady` is already `true`, so the handler is dead. In-package consumers escape only by Unity ordering luck (`VoxrCommandRecogniser.OnEnable` happens to subscribe before `VoxrPushToTalkController.Start` initialises).

This is **not** a new main-thread stall — `vosk_bridge_init` always blocked the main thread, and this PR removes ~719 ms from it. What regresses is call shape and event ordering. Being fixed on this branch.

### 2. MAJOR — a derived figure is presented as a device measurement, in three shipped files

`CHANGELOG.md:22` — *"Measured on a Quest 3 (…): time-to-voice-ready on an ordinary cached launch is ~584 ms"*. `Documentation~/getting-started.md:172` — *"on a Quest 3 that moved time-to-voice-ready from 1303 ms to ~584 ms"*. Only 1303 and 719 were measured; ~584 is `1303 − 719`, arithmetic on the **pre-fix** run, and the tier-1 path has never executed on a device. The PR body's own deferred box says exactly this.

The defence that the tilde marks derivation fails: `~719` is also tilde'd and *was* measured. The author marks derivation explicitly elsewhere ("derived two ways"), and the #154 entry one bullet away carries the right disclosure — it simply was not applied here.

`KNOWN_LIMITATIONS.md:359` separately attributes the whole entry to *"device measurements taken for #153"*, though its third-launch tier-1 outcome (`:357`) could not have been produced by pre-fix probes. Minor by comparison; the Repro lines are reader instructions per this file's convention, and are code-true.

### 3. MINOR — the "never under-invalidates" absolute is false off Android, and stated four times

Non-Android, `ResolveArchiveContainerPath` returns the `.zip` itself (`ModelExtractor.cs:339`), so the token is (path, length, mtime) of the very file whose contents are in question. Four sites state the property as impossible: `ModelExtractor.cs:287`, `CHANGELOG.md:22`, `KNOWN_LIMITATIONS.md:377-379`, and `getting-started.md:57` — which applies the absolute to precisely this `.zip` case.

The length half of the collision is **not** the improbable half. The verifier deflated a realistic `conf/model.conf` and applied single-digit decoder edits: 2 of 4 (`--beam=13.0→12.0`, `--lattice-beam=6.0→8.0`) produced a **byte-length-identical, byte-different** zip — the exact workflow `getting-started.md:63` recommends. What remains is a preserved mtime: a timestamp-normalising build, a `touch -r`, or a coarse-granularity volume. Then tier 1 serves the stale model and the tuning is a silent no-op — the #145 defect, in the forbidden direction.

`KNOWN_LIMITATIONS.md` also contradicts its own "never" six lines later at `:383`.

*(`cp -p` / `rsync -a` were named in the original finding and are wrong — they carry the new file's mtime and are safe. Narrow the mechanism list, not the claim.)*

### 4. MINOR — the jar-parse fallback stamps a guessed container, and the comment says it cannot

`ResolveArchiveContainerPath` returns `fallbackPath` for a malformed jar shape (`ModelExtractor.cs:343`), and the caller passes `Application.dataPath` (`:292-296`) — which on Android **is** the base APK, a stat-able file. So `ComputeSourceToken` returns a full token rather than the `null` at `:303`.

That falsifies the comment at `:324-326` — *"a token that fails to resolve degrades to the hash path, so a bad parse costs performance and never correctness"* — and the same rationale in `MalformedJarPath_WithoutSeparator_FallsBack`, whose own fixture constant is an `.apk` path and which asserts only the returned string, never a stat.

Verified as **latent, not live**: reaching it needs a path that starts with `jar:file://`, contains no `!`, and names something other than `dataPath` — and the split/OBB case is exactly where the `!` *is* present. The safety holds by coincidence of Unity's string construction, not by the reasoning given. Returning `null` for an unparseable jar shape would make the comment true.

### 5. MINOR — the in-place refresh falsifies a retained doc sentence

`getting-started.md:59` still says the stamp is *"written before the rename so a cache can never be stamped with a key it does not match"* — retained verbatim from base. There are now two writers, and `ModelExtractor.cs:149` writes into the **published** cache via non-atomic `File.WriteAllText`. The same file already says at `:55` that the identity is "refreshed in place", contradicting `:59`.

The torn-write direction was walked and is safe: a zero-length or prefix-truncated stamp fails both tiers and costs one re-extraction. Suggested: scope the clause to extraction and say a refresh interrupted mid-write reads as no match.

### 6. NIT — the tier-1 fall-through list is framed as closed and is not

`getting-started.md:55` (*"Only when this does not settle the question — no stamp, no recorded identity, or an identity that has moved"*) and `CHANGELOG.md:22` (*"Anything else — …"*) omit the gate's fourth limb, `cachedKey != null` (`ModelExtractor.cs:81`) — a stamp with a token but no hash. Unreachable by anything the SDK writes, so impact is near zero; what makes it wrong is the closed-set operator. Separately, `getting-started.md:59` tells a hand-editor the stamp will "lie about a cache the SDK will then go on serving", when deleting the hash line actually forces a re-extraction.

### 7. NIT — off-by-one in a test comment

`Tests~/Runtime/ModelExtractorCacheTests.cs:692` says *"The nine issue-#145 tests"*; the base carries **ten** (`grep -c "\[UnityTest\]"` on `4eb034b` = 10), and the PR body says ten twice. Fix: "nine" → "ten".

---

### Considered and killed

- **#154's effect stated as achieved fact in the prose files** — REFUTED. The finding's premise was that the CHANGELOG hedges where the docs do not; it does not. `CHANGELOG.md:24` makes the same mechanism claim ("hands the frame back to Unity for its duration"), and its hedge scopes the *147.8/177.2 ms numbers*, not the mechanism. All three files are epistemically aligned, and "a completion callback does not hold the frame" is a source-checkable statement about mechanism, not a claim of measurement.
- **The CHANGELOG half of finding 5** — REFUTED. *"an interrupted **extraction** cannot leave a half-written or wrongly-stamped cache behind"* is scoped to extraction and stays true; the refresh is not an extraction and the entry describes it separately.

### Clean, and worth recording

`PIT` returned **zero** findings after establishing, not assuming: the Unity statics are all reached before the first `await`; the continuation resumes on the main thread, so `request.result` and `downloadHandler.data` stay legal (`RunContinuationsAsynchronously` changes inlining, not thread affinity); no disposal race, since `using var request` outlives the await; no missed completion, per Unity's documented `completed` semantics. `GONE` confirmed all six #145 fallback behaviours preserved byte-for-byte, and that no base-written stamp can be misread by the new parser.

### Not verifiable from here

- **Anything on device.** The rewritten Android branch is behind a runtime `RuntimePlatform.Android` check, so no host test can execute it — #154 has zero behavioural coverage, as the PR states.
- **Every device figure** (1303, ~719, ~584, ~19 ms, +43 MB, 147.8, 177.2) came from earlier probe runs and could not be re-derived here.
- **The Unity suite results** (EditMode 153/153, PlayMode 612/612) were reported by a separate agent; not re-run.
- **Split Application Binary / Play Asset Delivery**, per the PR.

jinwoo1601:
Pushed `777e538`, which addresses **finding 1 (BLOCKER)** — the tier-1 hit firing `OnModelReady` inside `Initialise()`. One `await Task.Yield();` on that path restores asynchronous completion; it is not the #154 poll loop, and `MatchingSourceToken_SkipsArchiveRead` still passes, so the zero-archive-I/O optimisation it protects is intact.

Re-verified: **EditMode 153/153, PlayMode 613/613** (612 + the new test), fresh result files. The new `MatchingSourceToken_DoesNotCompleteSynchronously` ran as `runstate="Runnable"` and passed. It is non-vacuous by pairing rather than by mutation: `IsCompleted == false` alone would also hold if the call had fallen through to the read path, so it is asserted alongside `Reads == 0` on a counting archive source — together they can only both hold on a tier-1 hit that yielded.

**Findings 2–7 are deliberately left unfixed** for you to rule on. Finding 2 (the derived ~584 ms presented as measured, in three shipped files) and finding 3 (the "never under-invalidates" absolute, false off Android in four places) are the two I would want a decision on before merge — both are prose, neither touches behaviour.

jinwoo1601:
Pushed `54b79a0` — **all remaining findings addressed**. With `777e538` (finding 1, the blocker) and finding 7 already in, findings 1–7 are now closed.

**Finding 4 got a code fix rather than a reworded comment.** `ResolveArchiveContainerPath` dropped its `fallbackPath` parameter and returns `null` for every shape it cannot resolve. The fallback was `Application.dataPath`, which on Android *is* the APK, so an unparseable `jar:` shape produced a confident-looking token for a container nobody had verified — while the comment claimed a failed resolve degrades to the hash path. That comment is now true by construction instead of safe by coincidence of Unity's string building. Three EditMode tests renamed off "FallsBack" and assert `IsNull`; one added for a jar shape that *names a real-looking container* and must still resolve to nothing.

**Findings 2, 3, 5, 6 were prose and are corrected in place** — including the fourth site for finding 3, the code comment in `ModelExtractor.cs` the three doc sites were written from.

**One judgement call, flagged for you.** Finding 3 is recorded, not closed. The stale-serve window is real but confined to the Editor and standalone builds, where the "container" resolves to the loose `.zip` itself; closing it in code means never letting the cheap tier settle the question off Android, which returns every Editor launch to a full 39 MB read and hash. The Quest launch cost is what the ruling was about; the Editor never was. It now has its own `KNOWN_LIMITATIONS.md` entry with both conditions (same byte length **and** a preserved mtime), the note that `cp -p`/`rsync -a` are *safe* because they carry the new file's mtime, and the remedy. Say the word if you'd rather close it in code — it is a small change.

**Verification:** EditMode **154/154**, PlayMode **613/613**, fresh results. The fixture reports 7 cases (3 renamed + 1 new + 3 pre-existing), so the +1 is the addition and not a rename-drop masked by one.

Unchanged from before: #154's fix still has **zero behavioural coverage** — the branch sits behind a runtime Android check — and no post-fix device measurement exists. The container shapes are covered only through injected path strings in EditMode; whether a real Quest APK, split binaries included, resolves the same way at runtime is not claimable from WSL.

PR #156, opened 2026-09-09
Key per-word confidence by token position, not word text (#146)

### LINKED ISSUE #146: Per-word confidence is keyed by word text, so repeated words report the wrong value
## Summary

Per-word confidence is carried into the parser as a `Dictionary<string, float>` keyed by word **text**, first occurrence wins. A transcript that repeats a word therefore reports the wrong confidence for every later occurrence, and the command-level `Confidence` (minimum over matched tokens) can be wrong in either direction.

## Where

`Runtime/Commands/VoxrCommandParser.cs`:

- `BuildWordConfidence` `:3802-3812` and `InstanceBuildWordConfidence` `:3821-3831` — `if (!d.ContainsKey(w.Text)) d[w.Text] = w.Confidence;`
- `ComputeConfidence` `:4090-4113` — looks up `tokens[i]` by text; returns `-1` when the table is null or empty.

Example: "orient heading two two seven" with VOSK reporting `two` at 0.9 then 0.5. The table holds 0.9 for `two`; the second `two` is reported at 0.9; the command's minimum confidence is overstated. The reverse case understates it.

## Consequences beyond the wrong number

- The `-1` "no data" path bypasses the `minConfidence` gate (`Runtime/Commands/VoxrCommandRecogniser.cs:980`) and eager-flush condition 7 (`VoxrCommandParser.cs:4510-4512`). Any future change to what the decoder supplies per word (N-best mode supplies none) lands on this path silently. A token-aligned carrier makes that explicit.
- Any per-word feature (a confidence display, timing-based slot resolution) needs position, not text, as the key.

## Proposed change

- Replace the text-keyed dictionary with a token-aligned array: `float[] confidence` (and later start/end) indexed by token position, built once per parse from the `VoxrWord[]` the recogniser already receives. Tokens and words come from the same transcript, so alignment is by position; the tokeniser must produce one token per `VoxrWord` (verify, and handle `[unk]` and any normalisation the tokeniser applies).
- `ComputeConfidence` reads `confidence[i]` for `i` in the matched span.
- Keep the `-1` contract for "no word data" so existing behaviour of both gates is unchanged when words are absent.
- Remove the pooled dictionary from the hot path (an allocation-free array reuse replaces it).

## Constraints

- No change to which commands fire on any existing fixture or corpus where words are unique within the transcript; the 699-row parser corpus and the WAV-replay suite must pass unchanged except for transcripts with repeated words, which must be listed in the PR with old and new confidence.
- `ZeroAllocPollPathTests` must stay green.

## Acceptance

- Test: repeated word with different confidences yields the per-position value, and the command minimum reflects the lower one.
- Test: words absent → `Confidence == -1` and both gates bypass as today.
- Diagnostics (`VoxrMatchDiagnostics`, session log `SlotDto.confidence`) read the aligned values.
- `Documentation~/api/data-types.md` (`VoxrCommand.Confidence`) wording checked; `CHANGELOG.md` under Unreleased (Fixed).

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), critique verification (h3) and item R23; promoted by ruling 4 (2026-09-06) as the carrier for the per-word signal N-best parsing will need.

jinwoo1601:
Closes #146

## What

Per-word VOSK confidence was carried into the command parser as a `Dictionary<string,float>` keyed by word **text**, keeping each distinct word's **first** occurrence, and every token of a matched span was then looked up by text. A repeated word was scored at the first occurrence's confidence wherever it appeared — including when that first occurrence lay *outside* the match — so `VoxrCommand.Confidence` (the minimum over the span) was wrong in either direction.

The carrier is now a `float[]` indexed by **token position**.

- `VoxrCommandParser.InstanceBuildWordConfidence(string[] tokens, ReadOnlySpan<VoxrWord> words)` builds it; `ComputeConfidence` reads `wordConfidence[i]`.
- `internal const float NoConfidence = -1f` names the per-token "no data" sentinel.
- The pooled `Dictionary` is replaced by a pooled `float[]`, resized only when an utterance is longer than any seen before.
- `ParseInternal`, `TryEagerCommit`, `BuildSiblingRivalCommand`, `PendingCommandHandler.TryFollowUpSlotFill`, and the recogniser's `TryBuildAmbiguity` / `BuildAttempt` / `BuildBarredAttempt` all take `float[]`.
- The dead static `BuildWordConfidence(VoxrWord[])` is **removed** rather than ported — it had no call sites anywhere in the repo. Flagging that explicitly since it is a deletion the issue did not literally ask for.
- `ProcessParsedResults`' two overloads collapsed into `ProcessParsedResultsCore(string, ReadOnlySpan<VoxrWord>, float[])`. They existed only to build the old dictionary. The `float[]` is supplied by the buffered path and is `null` on the unbuffered one, where it is built after the token split.
- `UtteranceBuffer` now owns the per-token confidences for a buffered utterance (`ConfidenceBuffer` / `ConfidenceCount`), built one result at a time — see the next section for why.

## Why alignment is per *result*, not across the join

**Revised after review — the original version of this PR got this wrong, and the history is on the PR.** It built the array with a single greedy walk over the whole joined utterance. `UtteranceBuffer.Append` takes `text` unconditionally but `words` only when a result supplies them, so the token stream and the word stream can be offset, and the walk then credited a word to the first token whose text matched — possibly a token from a *different* result. `InjectText("cease fire")` with no words, then a result `"cease fire"` carrying `[cease 0.2, fire 0.2]`, produced `[0.2, 0.2, -1, -1]`: the wordless segment credited, the segment that actually supplied the words voided. The voided round then computed `-1`, which **bypasses** `minConfidence` rather than failing it, so a command fired that the pre-#146 code rejected.

Alignment is a property of a single result, so `UtteranceBuffer` now builds the array itself, one segment per `Append`. A word can never reach a token from another result. Within a segment:

- no words -> every token of that segment takes `NoConfidence`;
- `words.Length == tokenCount` -> straight positional copy (the shape the decoder actually emits, measured 1:1 below);
- ragged -> a two-cursor walk over *that segment's own* tokens and words, so a mis-pairing stays inside the result that caused it.

`ProcessParsedResultsCore` takes the array back as a parameter; `null` means "no buffer segmented this utterance" and it is built from that single result's own words.

**The walk asks no alignment guarantee of its caller.** An earlier revision advanced the word cursor only on a match, which could not distinguish a word *waiting for a later token* from a word belonging to *no token at all* — so one foreign word parked the cursor and cost every later token its confidence. That question is directly answerable, and the walk now asks it: on a mismatch, if the word's text appears in a later token only the token cursor advances; if it appears in no remaining token the word is dropped and the same token is retried. Both cursors are monotone, so it terminates in `tokens + words` steps and stays allocation-free. A whitespace-only word text — which slipped past the old `IsNullOrEmpty` guard and parked the cursor just as effectively — is now classified as foreign like any other.

**There is no unverified positional fast path anywhere.** A length-equal copy would credit a token with another word's confidence whenever the two streams merely happen to be the same length, which contradicts the claim the walk exists to make. The parser verifies every token; the buffer keeps its allocation-free path but *verifies* it, comparing token spans of the text against the words without cutting substrings, and falls through to the aligned walk on the first mismatch. On the decoder's own 1:1 output the walk is exactly N successful ordinal comparisons and never looks ahead, so verifying costs a comparison per token and nothing else.
## Validation

- [x] **Unity EditMode: 155/155 pass.** Host project `VoXR TestGround`, Unity 6000.4.7f1.
- [x] **Unity PlayMode: 633/633 pass.** Both platforms actually executed — most of the affected tests are PlayMode. (Counts are after the review fixes; the rounds were 154/618, then 155/626, then 155/633.)
- [x] **The allocation pin the review found missing now exists.** No test previously drove `InstanceBuildWordConfidence` with a non-empty `words` argument — the existing `Is.Not.AllocatingGCMemory()` pins pass a prebuilt `null` carrier, so the builder never ran inside a measured region. `BuildWordConfidence_AllocatesNothingPerCall` measures both the 1:1 and the ragged path, warms the pool deliberately (`Array.Resize` would otherwise fire inside the measured region), and carries a live-instrument control asserting that a deliberate allocation IS caught — so it cannot pass vacuously.
- [x] **The `text`/`result` 1:1 assumption was measured, not assumed.** Ran the committed 16-fixture corpus through the desktop harness (`NativeBridge~/harness`) on real libvosk 0.3.45 + `vosk-model-small-en-us-0.15`, dumping the raw final JSON: `result` is exactly 1:1 with the space-separated tokens of `text` in all 16 finals, `" ".join(words) == text` exactly, and `[unk]` **is** a first-class `result` entry at its own index. Empty finals are `{"text": ""}` with no `result` key, so tokens and words are both 0. The harness was restored byte-identical afterwards and re-run green (16/16, exit 0).
- [x] **New tests are real pins, not tautologies** — each carries a comment recording what the old carrier produced:
  - `Sequential_ConfidencePropagatedPerCommand` — rewritten. It previously *encoded the bug*: it paired 4 tokens with a deliberately 3-element `VoxrWord[]` and asserted both commands came out at `0.72`, with a comment admitting it was reading the first `"fire"`. Now supplies all four words 1:1 with the second `fire` at `0.41`; asserts `0.72` / `0.41`.
  - `RepeatedWord_WeakSecondOccurrence_LowersCommandConfidence` — asserts `0.35`; old produced `0.90`.
  - `RepeatedWord_WeakFirstOccurrenceOutsideSpan_DoesNotLowerCommandConfidence` — asserts `0.90`; old produced `0.20`.
  - `PartialWordData_TokensWithoutWords_AreSkipped` — pins the buffer-merge shape at `0.70`. Same value as the old code: this is a **non-regression** pin, not a behaviour change, and is labelled as such in the test.
  - `CorpusTranscript_RepeatedDigit_CommandConfidenceUnchanged` — asserts `0.5`, unchanged by design.
  - `NoWordData_BypassesConfidenceGate` — new coverage with a control (`null` -> `Commit`, `[0.2, 0.2]` -> `None`). No dedicated pin for this existed before; it was only asserted incidentally in comments.
- [ ] **On-device (Quest) check: not required and not claimed.** No `NativeBridge~/` change, no MonoBehaviour lifecycle or mic-path change; binding 2 (arm64 build + desktop harness regression) was correctly skipped for the diff, though the harness *was* run for the measurement above.

## What moves on the committed corpus

The corpus contains exactly one transcript that exercises the defect. `fire_two_torpedoes_at_bravo_two.wav` decodes as `fire two torpedoes at bravo two` with `two` at **0.50** (index 1) and **1.00** (index 5); the old table read both as `0.50`. Under the pinned grammar-constrained decode confidence is otherwise near-degenerate — 60 of 62 word entries came back at `1.0`, and the only other non-1.0 value is a non-repeated `two` at `0.5` in `set_heading_two_seven_zero`.

- **Command-level confidence: `0.50` both before and after**, because the weak occurrence is the earlier one *and* lies inside the matched span. So no command's fire/no-fire decision moves on this corpus.
- **Slot-level: the `target` slot covering `"bravo two"` (token span `[4,6)`) goes `0.50` -> `1.00`.** Per-slot confidence is Editor diagnostics and session-log only and is read by no gate. **This figure is *derived*** from the two measured word confidences and the span, not separately observed — flagged as derived in the CHANGELOG and in `scoring.md` too.

**No before/after has been run over the corpus as a whole**, so nothing is claimed about aggregate verdict changes beyond the above.

Issue #146's Constraints also name the 699-row parser corpus. It is **inapplicable by construction**: the A/B rig drives it through the text-only `Parse(text)` overload, which supplies no `VoxrWord[]` at all, so no per-word confidence is computed for any of its rows and none can move. It was not run, and that is why.

## What does not change

The `-1` contract (it still bypasses `minConfidence` and eager-flush condition 7), the `[unk]` skip, the minimum-not-average rule, the leading-required-miss bar, both gates, `minScore`, the scoring model, and which commands fire on any transcript without a repeated word. The carrier is `internal`: **no public API change** — no new field, no changed signature, no new event.

## Docs

`Documentation~/scoring.md` — the "Repeated words resolve by text, not by position" block is inverted and the old behaviour recorded as pre-#146 history, with the measured worked example. `Documentation~/api/data-types.md` — `VoxrCommand.Confidence` row tightened. `CHANGELOG.md` — `[Unreleased]` / `Fixed`. `command-recognition.md:131`, `api/command-recogniser.md:12`, `troubleshooting.md` and `editor-testing.md` were each checked against the new mechanism and deliberately left alone; their claims survive unchanged.


jinwoo1601 (commented):
## Independent review — PR #156

**Profile:** slim (issue lane) — `CLAIM` + `LINE`, `GONE` — **plus `XFILE`**, added because this diff changes a carrier contract across three runtime files and its correctness rests on a `text`↔`words` alignment contract spanning the JSON parser, the utterance buffer and the parser. No angle was dropped.
**Head reviewed:** `d8bc5b5a625ac6003f6e8288eadbb974b84db273` · **merge-base:** `330b956`
Four finder agents ran against the diff with no context from the sessions that wrote it. Six candidate findings survived dedup; all were adversarially verified. **2 CONFIRMED, 2 REFUTED.**

---

### 1. BLOCKER — cross-segment misattribution loosens `minConfidence`; fires a command the base rejected

**CONFIRMED.** Found independently by `LINE`, `XFILE` and `GONE`.

`Runtime/Commands/VoxrCommandParser.cs:3987-3998` consumes `words[j]` at the **first** token whose text matches, with no check that token and word came from the same buffered result. `UtteranceBuffer.Append` (`Runtime/Commands/UtteranceBuffer.cs:24-32`) adds `text` unconditionally but `words` only when a result supplies them, so the two streams can be offset.

Minimal repro (defaults `bufferWindow 0.5`, `minConfidence 0.4`, eager flush off): `InjectText("cease fire")` with no words, then within the window a result `"cease fire"` carrying `[cease 0.2, fire 0.2]`. Tokens are `[cease, fire, cease, fire]`; the walk yields `[0.2, 0.2, -1, -1]`. Round 2's span `[2,4)` is all `-1`, so `ComputeConfidence` returns `-1`, which **bypasses** the gate at `VoxrCommandRecogniser.cs:1036` — the command fires. At `330b956` the text-keyed lookup returned `0.2` for both rounds and both were rejected.

The uncovered tokens are credited and the covered ones are voided — the exact inversion of what the change intends, in a shape `Documentation~/scoring.md:286` documents as supported. It also falsifies the new comment at `VoxrCommandParser.cs:3962` ("no token is ever credited with a confidence that came from a different word"). `PartialWordData_TokensWithoutWords_AreSkipped` cannot catch it: its uncovered head shares no token text with its covered tail.

**Fix:** align per buffer segment — tokens of a result against the words of that same result — so a word can never be attributed across a segment boundary.

---

### 2. MAJOR — a word matching no later token parks the cursor, voiding every token after it

**CONFIRMED**, severity lowered on verification: **no in-repo producer can create the shape**. The JSON parser copies VOSK words verbatim (1:1, as measured), `CreateSimulatedWords` splits the same string, and `UtteranceBuffer` preserves subsequence order. It is reachable only through `InjectText`/`InjectResult`, which apply no cross-validation.

`VoxrCommandParser.cs:3987-3998` advances `j` only on a match. `InjectText("cease fire", [please 0.9, cease 0.2, fire 0.2])` never matches, so the array is all `NoConfidence` → `-1` → gate bypassed, where the base returned `min(0.2,0.2)=0.2` and rejected. A word whose `Text` is whitespace-only passes the `IsNullOrEmpty` guard and can match no token (tokens come from a `RemoveEmptyEntries` split), so one such word parks the cursor for the whole utterance.

**The claim is wrong unconditionally, independent of reachability.** `VoxrCommandParser.cs:3961-3964`, this PR's body, and `Documentation~/scoring.md:269` all state that the walk "resynchronises on the next match" and degrades "per token, not per utterance". That holds only when `words` is an in-order subsequence of `tokens` — a precondition none of the three states, and one the new test at `Tests~/Runtime/VoxrCommandParserTests.cs:370` quietly relies on.

---

- **Pooled array grows to exact size rather than doubling, so a lengthening run allocates per utterance.** REFUTED: the pool is a high-water mark that never shrinks or resets, so lifetime allocations are bounded by `maxTokensEverSeen − 32` and the ramp is warm-up, not steady state. The base's pooled `Dictionary` allocated on growth too. The linear-vs-geometric asymmetry against `UtteranceBuffer.cs:29` is real but changes event count, not the invariant.
- **"The corpus contains exactly one transcript that exercises the defect" is false — three fixtures carry a repeated word.** REFUTED: the PR's very next sentence disposes of the other two (60 of 62 word entries at `1.0`; the only other non-1.0 is a non-repeated `two`), so `launch_one_missiles_target_hotel_one` and `filler_launch_three_torpedoes_target_alpha_three` decoded entirely at `1.0` and provably cannot move. The section is headed "What moves". The reading of issue #146's listing constraint as "every transcript containing a repeated word" is also untenable — 224 of the 699 corpus rows repeat a token and none carry per-word data at all.

---

### Notes

Unverified, below the numbered-finding bar; recorded for the human, not acted on.

- No allocation test drives `InstanceBuildWordConfidence` with a non-empty `words` span. Both `Is.Not.AllocatingGCMemory()` tests pass `TryEagerCommit(tokens, null, …)`, so the carrier is pre-built as `null` and the builder never executes inside a measured region — the invariant-5 pin is inert for this path.
- Issue #146's acceptance asks that "words absent → both gates bypass". The new test pins only `TryEagerCommit`; the recogniser gate at `VoxrCommandRecogniser.cs:1036` has no `-1` pin.
- The rewrite of `Sequential_ConfidencePropagatedPerCommand` removed the suite's only ragged-**and**-repeated input shape (3 words against 4 tokens) without re-pinning its new answer.
- The 699-row A/B rig corpus named in issue #146's Constraints is inapplicable by construction — the rig calls the text-only `Parse(text)` overload, so no confidence is computed — but the PR does not say so.

---

### Not verifiable from here

The Unity suites were not re-run by this review (WSL cannot drive Unity); the PR's EditMode 154/154 and PlayMode 618/618 were audited for internal consistency against the diff's test count and found sound, not reproduced. The measured per-word decoder figures were likewise audited, not re-decoded. On-device (Quest) behaviour is never claimable from here and this diff does not touch it.

Finding 1 is a blocker and is this run's own unfinished work — a fix follows on this branch. Finding 2 and all Notes are left unruled for the human.

jinwoo1601:
**Blocker fixed** — this push addresses finding **1** (cross-segment misattribution loosening `minConfidence`) from the review above. Head is now `106a9a9`.

Alignment is a **per-result** property, so `UtteranceBuffer` now builds the per-token confidence array itself, one segment per `Append`, and a word can never be attributed to a token from another result. `ProcessParsedResultsCore` takes the array back as a parameter; `null` means "no buffer segmented this utterance" and it is built from that single result's own words. On the verified repro the array is now `[-1, -1, 0.2, 0.2]` instead of the reversed `[0.2, 0.2, -1, -1]`, so the round that actually supplied the words is the one judged against the gate, and it is rejected at `0.2 < 0.4` as it was before #146.

One consequence worth stating rather than burying: round 1 — the wordless injected segment — now reports `-1` and therefore bypasses the gate and fires, where the pre-#146 text-keyed dictionary leaked `0.2` into it from the *later* result and rejected it. That leak was the defect; `-1` is the correct "no confidence known" value for a segment that supplied no words, and it matches what an unbuffered wordless `InjectText` has always done.

**Finding 2 is not code-fixed, and that is a judgement the reviewer flagged for you rather than one I made quietly.** The cursor stall is reachable only through a hand-built `VoxrWord[]` passed to `InjectText`/`InjectResult` that is not an in-order subsequence of the text; no in-repo producer can construct it. What *was* unconditionally wrong is the claim, so the precondition and the gate-loosening consequence are now stated plainly in the `InstanceBuildWordConfidence` comment, in `Documentation~/scoring.md`, and in the `CHANGELOG.md` entry's "what does not move" clause. Building lookahead machinery for caller misuse looked like the wrong trade against the issue lane; say the word if you would rather have it hardened.

Also corrected here: **the PR body above still carries the unqualified "the walk resynchronises on the next match" sentence** that finding 2 refuted. Read it with the precondition attached — `words` must be an in-order subsequence of `tokens` by text.

The four Notes are untouched and unruled.

### Tests

- `Tests~/Runtime/UtteranceBufferTests.cs` (new) — pins segment alignment both ways round, the `ConfidenceCount == token count of PeekText()` invariant across an empty append and one padded with repeated spaces, and the `ClearWords`/`Reset` reset.
- `BufferedSegments_WithCollidingText_ReportConfidencePerRoundInRoundOrder` (EditMode) — the executable discriminator. A "how many commands fired" assertion does **not** separate the bug from the fix: exactly one fires either way, because whichever round holds the words is the one rejected. This asserts the per-round confidences in round order; pre-fix they were reversed.
- `Sequential_RaggedWordData_ConfidencePropagatedPerCommand` — restores the ragged-**and**-repeated input shape the earlier rewrite dropped (Note 3).
- `InjectText_NoWordData_BypassesMinConfidenceGate` + the buffered variant — pins the second gate named in #146's acceptance (Note 2). Honest label: these pass pre-fix too, so they are coverage, not regression pins; each carries an in-line control that must *not* fire, so neither can pass vacuously.

### Validation

- [x] Unity **EditMode 155/155**, **PlayMode 626/626** — both platforms actually executed against host project `VoXR TestGround` (Unity 6000.4.7f1). Counts reconcile: +1 EditMode, +8 PlayMode against the previous run's 154/618.
- [ ] On-device (Quest): not required and not claimed — no `NativeBridge~/`, lifecycle, or mic-path change.

jinwoo1601:
**All review findings and notes are now addressed.** Head is `cb659ce`.

### Finding 2 — fixed in code, not just documented

I previously left this one documented, on the grounds that it was caller-misuse-only. On reflection that was the wrong call: it was a *gate-loosening* path, and the underlying question turned out to be directly answerable rather than needing a heuristic.

The walk advanced its word cursor only on a match, so it could not distinguish a word **waiting for a later token** from a word belonging to **no token at all** — and it treated both as "hold the cursor", which parked it and cost every later token its confidence. It now asks which case it is: if the word's text appears in a later token, only the token cursor advances; if it appears in no remaining token, the word is dropped and the same token is retried. Both cursors are monotone, so it still terminates in `tokens + words` steps and remains allocation-free. A whitespace-only word text — which slipped past the old `IsNullOrEmpty` guard and parked the cursor just as effectively — is now foreign like any other.

The subsequence precondition is therefore gone, and **the caveat it forced into `scoring.md` and the CHANGELOG has been removed with it.** Both files are net reductions. The two copies of the walk are now one shared `FillAlignedConfidence`.

### One thing the fix surfaced, which I also closed

Removing the stall exposed that the equal-length **positional fast path** contradicted the claim the walk exists to make: a length-equal copy credits a token with another word's confidence whenever the two streams merely happen to be the same length. That was the third overclaiming comment this PR has produced, so rather than re-word it I closed the gap. The parser now verifies every token. The buffer keeps its allocation-free path but *verifies* it, comparing token spans of the text against the words without cutting substrings, and falls through to the aligned walk on the first mismatch. **No token is credited with a value from a different word on any path now**, so the claim is true as written.

Cost: on the decoder's own 1:1 output the walk is exactly N successful ordinal comparisons and never looks ahead.

### The four Notes

1. **Allocation pin inert for the builder** — closed. `BuildWordConfidence_AllocatesNothingPerCall` measures both the 1:1 and the ragged path (the latter driving the `AppearsFrom` lookahead), warms the pool deliberately so `Array.Resize` cannot fire inside the measured region, and carries a live-instrument control asserting a deliberate allocation IS caught — so it cannot pass vacuously.
2. **`-1` bypass pinned on only one gate** — closed earlier by the two `..._NoWordData_BypassesMinConfidenceGate` tests, each with a control that must not fire.
3. **Ragged-and-repeated shape dropped by the rewrite** — closed earlier by `Sequential_RaggedWordData_ConfidencePropagatedPerCommand`.
4. **699-row corpus not mentioned** — now stated in the PR body. It is inapplicable by construction: the A/B rig drives that corpus through the text-only `Parse(text)` overload, which supplies no `VoxrWord[]`, so no per-word confidence is computed for any row and none can move.

The two refuted findings stand refuted and no code was changed for them, with one exception I am flagging rather than slipping in: the parser pool now grows geometrically to match its two sibling pools in `UtteranceBuffer`. That is a consistency tidy, not a defect fix — the reviewer's claim that exact-size growth breached the allocation invariant was correctly refuted, since the pool is a high-water mark.

### New tests

Genuine regression pins (fail before this push): `ForeignLeadingWord_IsDropped_AndTheRealWordsKeepTheirConfidence` and `WhitespaceOnlyWordText_IsDropped_NotLeftParkingTheCursor` (both `0.2`; both `-1` before, i.e. the gate was bypassed), `Append_EqualLengthButMisPaired_FallsThroughToTheAlignedWalk` and `Append_ForeignTrailingWord_IsDroppedRatherThanCreditedToItsToken` (both previously took the wrong word's value).

Coverage that passes either way, labelled as such in the tests themselves: `TransposedWords_NoTokenIsCreditedWithAnotherWordsConfidence`, `Append_EqualLengthAndCorrectlyPaired_...` (guards against a fix that rejected good pairings too), and the allocation pin.

### Validation

- [x] **EditMode 155/155, PlayMode 633/633** — both platforms actually executed against `VoXR TestGround` (Unity 6000.4.7f1). The +7 reconciles exactly against the previous 626.
- [ ] On-device (Quest): not required and not claimed — no `NativeBridge~/`, lifecycle, or mic-path change.

Nothing from the review is left open. Still not merged, and still yours to rule on.

PR #157, opened 2026-09-09
Reject a malformed grammar before it reaches VOSK (#150)

### LINKED ISSUE #150: Malformed grammar JSON crashes the process inside vosk_recognizer_new_grm instead of returning NULL
## Summary

`vosk_recognizer_new_grm` **segfaults** on several malformed-grammar shapes rather than returning `NULL`. Because it dies inside the native call, the managed `NULL` check never runs and the process — Unity Editor, or the game on device — is killed outright. There is no exception to catch and no error event.

Found while verifying an unrelated review finding on #143/PR #149. Pre-existing; nothing in PR #149 causes or worsens it.

## Measured

Against the vendored `libvosk` (`vosk-linux-x86_64-0.3.45`) and `vosk-model-small-en-us-0.15`, calling `vosk_recognizer_new_grm(model, 16000, <input>)` in a standalone C harness and reading the process exit status:

| input | exit | result |
|---|---|---|
| `["[unk]", "fire", "cease"]` | 0 | returns non-NULL (correct) |
| `["fire", 42]` | 1 | returns **NULL** — graceful, handled |
| `[]` | **139** | **SIGSEGV** |
| `["a"b", "c"]` | **139** | **SIGSEGV** |
| `{"grammar": ["fire"]}` | **139** | **SIGSEGV** |
| `not json at all` | **139** | **SIGSEGV** |

Each crashing case logs `WARNING (VoskAPI:UpdateGrammarFst():recognizer.cc:283) Expecting array of strings, got: '...'` and then dies before returning.

The `["fire", 42]` row is the interesting one: VOSK **can** fail gracefully. It guards "array containing a non-string" but not "not an array at all" and not "empty array".

## Why it matters here

`Runtime/EditorMicBackend.cs:290` handles the failure VOSK does not produce:

```csharp
if (_recognizer == IntPtr.Zero)
{
    fireError?.Invoke(VoxrBridgeErrorCode.ModelLoadFailed, "vosk_recognizer_new failed during SetGrammar.");
    return;
}
```

For every crashing shape above that branch is unreachable — control never comes back.

**This is not Editor-only.** The Android bridge calls the same function at `NativeBridge~/src/vosk_bridge.cpp:363` (`g_recognizer = vosk_recognizer_new_grm(g_model, g_sample_rate, grammar_json);`), so the same input crashes the app on device.

## Reachability

Not reachable through normal authoring, but not sealed off either:

1. **`VoxrSpeechRecogniser.SetGrammar(string)` is public** (`Runtime/VoxrSpeechRecogniser.cs:437`). Any integrator-supplied string reaches the native call — Editor via `EditorMicBackend`, Android via `BridgeNative.vosk_bridge_set_grammar`.
2. **A quote in a pattern literal escapes unnoticed.** `GenerateGrammarJson` serialises with no escaping pass (`Runtime/Commands/VoxrCommandParser.cs:3163-3174`). Slot values and alias keys containing punctuation *are* warned about (`:747`, `:780`), but **pattern literals are not checked at all**, so a pattern element containing `"` produces malformed JSON with no warning and then crashes.
3. `[]` crashes, though `GenerateGrammarJson` always emits `[unk]` (`:3157`) and so never produces a bare `[]` itself.
4. The `{"grammar": [...]}` shape already circulates in the repo — `Tests~/Runtime/VoxrSpeechRecogniserOwnershipTests.cs:134` passes it. That call is rejected as a non-owner before reaching the decoder, so it is safe today, but it shows the shape is one a caller can plausibly construct.

## Suggested fix

Validate before handing the string to the native call — the package already knows what a well-formed grammar looks like:

- In `EditorMicBackend.SetGrammar` and `vosk_bridge_set_grammar`, reject a grammar that is not a non-empty JSON array of strings, raising the existing `VoxrBridgeErrorCode` path instead of calling through.
- Optionally extend the punctuation check at `VoxrCommandParser.cs:747` to pattern literals, so the authoring-time warning catches the quote case before it can ever be serialised.

Upstream hardening of `UpdateGrammarFst` would be the real fix, but guarding on our side is cheap and keeps a bad grammar from taking the whole app down.

## Out of scope

Any change to what a *well-formed* grammar does. This is purely about not crashing on a malformed one.

jinwoo1601:
Closes #150

## What

`VoxrSpeechRecogniser.SetGrammar` now proves a grammar is well formed before any native call is made. A grammar that fails raises the existing `VoxrBridgeErrorCode.ModelLoadFailed` path plus a `Debug.LogError`, and returns — leaving the recogniser already in use untouched and still decoding.

- **New** `Runtime/VoxrGrammarValidator.cs` — an `internal static` hand-rolled check (no JSON dependency, in keeping with `VoxrJsonParser`) answering exactly one question: is this a complete JSON array holding at least one element, every element a JSON string, with escapes and control characters honoured as JSON defines them. It is total — it never throws, for `null`, empty, whitespace-only or truncated input — and reports a reason naming the character index.
- **Modified** `Runtime/VoxrSpeechRecogniser.cs` — the guard, placed after `RejectIfBridgeOwnedByOther` and before the `#if UNITY_EDITOR_WIN` branch.
- **New** `Tests~/Editor/VoxrGrammarValidatorTests.cs` (30 cases) and `Tests~/Runtime/VoxrGrammarGuardTests.cs` (4 cases).
- `CHANGELOG.md` entry under `[Unreleased]` → `Fixed`.

`null` and `""` still mean "clear the grammar" and still take the no-grammar path; the check is asked only about a non-empty string. **No public API changes** — no new error code, no new member, no changed signature. Nothing about a well-formed grammar moves.

## Why

`vosk_recognizer_new_grm` does not validate its grammar: it dereferences its way through the parse and dies. Measured in the issue against libvosk 0.3.45, `[]`, `[\"a\"b\", \"c\"]`, `{\"grammar\": [\"fire\"]}` and `not json at all` each raised SIGSEGV inside the native call and took the whole process with it — no catchable exception, no `OnError`, nothing in the log. Only `[\"fire\", 42]` failed politely by returning NULL, which is the one shape the existing managed NULL check at `EditorMicBackend.cs:290` could ever have seen.

The likeliest way to author a crashing shape is not hand-written JSON but a double quote inside a command pattern literal or a slot value: `VoxrCommandParser.GenerateGrammarJson` serialises those verbatim with no escaping pass. `SetGrammar` is also `public`, so any integrator string reaches the call.

**Why the guard is managed-only.** Every path in this package that reaches `vosk_recognizer_new_grm` funnels through `VoxrSpeechRecogniser.SetGrammar` — Windows Editor via `EditorMicBackend.cs:282`, everything else via `BridgeNative.vosk_bridge_set_grammar` → `vosk_bridge.cpp:363`. One guard above both branches therefore covers the Android bridge too, and the shipped `.so` is byte-identical and was not rebuilt.

## Deliberate deviations from the issue's suggested fix

Both are judgement calls, flagged for you rather than buried:

1. The issue suggested guarding **in `EditorMicBackend.SetGrammar` and in `vosk_bridge_set_grammar`**. The native entry point is left unguarded. It is reachable only from the `NativeBridge~/` desktop harness, not from the package, so guarding it would buy nothing for a shipped app while costing an arm64 rebuild, a committed `.so`, and a deferred on-device verification on this PR. `vosk_recognizer_new_grm` itself likewise remains unhardened — anything reaching it by another route still crashes the same way. Both facts are stated in the CHANGELOG rather than glossed.
2. The issue's second, explicitly **optional** bullet — extending the authoring-time punctuation warning to pattern literals — is **not** included. It does not affect the crash now that the chokepoint is guarded, and it would land in `RunValidationWarnings`, whose ordering is load-bearing for `VoxrCommandRecogniserInjectionTests`' queued log expectations. Worth a follow-up issue if you want it; the rejection message names the quote-in-a-literal cause in the meantime.

## Validation

- [x] **EditMode — 185/185 pass**, 0 failed / 0 inconclusive / 0 skipped. Host project `VoXR TestGround`, Unity 6000.4.7f1, working tree referenced by local path. Includes the new `VoxrGrammarValidatorTests` fixture 30/30 — all four measured SIGSEGV shapes rejected, `PackageGeneratedGrammar_IsAccepted` pinning that real `GenerateGrammarJson` output still passes the check, and `HostileInput_NeverThrows` over every truncation of a valid grammar.
- [x] **PlayMode — 637/637 pass**, 0 failed / 0 inconclusive / 0 skipped. Includes the new `VoxrGrammarGuardTests` 4/4. The two rejection cases (`ObjectShapedGrammar`, `EmptyArray`) go red if the guard is removed; the other two (`WellFormedGrammar`, `NullOrEmpty`) are controls pinning that the guard does not over-reject, and stay green under that mutation because the uninitialised recogniser's `_editorBackend` is null. *(An earlier revision of this body claimed all four were mutation-sensitive. That was wrong — corrected per finding 1 of the review below.)*
- [x] **`VoxrSpeechRecogniserOwnershipTests` still 8/8**, including `SecondSetGrammar_RejectedLoudly` — it passes a malformed grammar to a non-owner and asserts exactly **one** error. The guard sits after the ownership check specifically to keep that true; the captured log confirms the single ownership error with no second grammar error.
- [x] **Every shape measured against the real decoder.** A standalone C probe against the vendored `libvosk 0.3.45` + `vosk-model-small-en-us-0.15` (one grammar shape per process, exit status read), validated first by reproducing all six rows of #150's own table exactly. Results:
  - **Everything the validator accepts is safe** — all 12 accepted shapes return non-NULL, `[""]` included. So accepting the empty element string is correct, not a gap.
  - **Three crashers #150 never found**: a whitespace-only string, an unterminated string (`["a`) and a bare `[` each SIGSEGV. All three were already refused — their rejection is load-bearing, not tidiness.
  - **Three shapes the guard refuses are actually tolerated by libvosk**: trailing content after the array, an invalid backslash escape, and a raw control character inside a string all return non-NULL. Refusing them is a deliberate strictness choice, not a crash averted — see the trade-off note below. This corrects a claim in the posted review, which had assumed VOSK rejected them too.
- [x] `NativeBridge~/` and `Runtime/Plugins/` untouched — the NativeBridge binding was correctly not run, and no `.so` changed.
- [ ] **No on-device (Quest) check needed.** The change is managed-only with no lifecycle, mic, or native surface; the shipped `.so` is byte-identical.

## Deliberate strictness beyond the decoder

The measurement above showed libvosk parses three malformed shapes leniently rather than refusing them. The guard refuses them anyway. That is a judgement call worth your attention, because it is the one place this PR changes behaviour for a grammar that previously *worked*: an author who puts a stray backslash or tab inside a slot value or pattern literal now gets the whole grammar loudly rejected, where before the decoder loaded it and silently dropped just the affected word.

Kept strict because a precise error naming the offending index is better authoring feedback than a word that silently never matches; because `GenerateGrammarJson` cannot emit any of the three from valid authoring, so this is reachable only from a typo or a hand-written grammar; and because encoding one vendored libvosk version's lenient JSON quirks as a contract would be brittle. Say the word if you'd rather the guard mirror the decoder exactly.


jinwoo1601 (commented):
## Independent review — PR #157

**Profile:** slim (issue lane) — `CLAIM`, `LINE`, `GONE`. No angles dropped; none added — the diff crosses no ABI, serialized-field, or cross-assembly contract (so no `XFILE`) and touches no MonoBehaviour lifecycle, coroutine, or main-thread-only API (so no `PIT`).
**Head reviewed:** `b337dc8a7c63e3e97ca4665b1a641b8327e5b7c3` · **Base:** merge-base `10d255b`.

**Two confirmed findings, both minor, both in the PR's own prose rather than in the code.** The guard itself verified clean: all three angles independently re-derived it and found no defect. `GONE` returned no findings at all.

---

### 1. The PlayMode validation box overstates the fixture — CONFIRMED (minor, high)

The ticked box says the four new `VoxrGrammarGuardTests` "assert on recorded `OnError` invocations and `LogAssert` … **so all four go red if the guard is removed**." Two of the four cannot.

`SetGrammar_WellFormedGrammar_RaisesNothing` and `SetGrammar_NullOrEmpty_RaisesNothing` are absence assertions. Delete the guard block (`Runtime/VoxrSpeechRecogniser.cs:449-469`) and control reaches `_editorBackend?.SetGrammar(...)` at `:471`; `_editorBackend` (`:68`) has no initialiser and is assigned only inside `InitialiseAsync` (`:197`), and the fixture deliberately never initialises. Null-conditional no-op — nothing raised, nothing logged, both tests stay green. They are controls pinning that the guard does not over-reject, not mutation evidence.

The count is exactly 2 of 4: the two rejection tests do go red three ways each (unsatisfied `LogAssert.Expect`, `Assert.AreEqual(1, _errors.Count)`, and `_errorCodes[0]` on an empty list).

**The claim as written is withdrawn.** Accurate: 2 of 4 are mutation-pinned, 2 are controls. The safety conclusion survives — the guard *is* pinned — but by half the evidence the box advertised.

### 2. "Where in the string it was wrong" is falsified by the headline case — CONFIRMED (minor, high)

`CHANGELOG.md` (`[Unreleased]` → `Fixed`, #150 bullet) states the rejection logs name "what was wrong, **where in the string it was wrong**, and the quote-in-a-literal cause", unhedged. The PR body likewise: "reports a reason naming the character index."

Of 14 distinct rejection reasons in `Runtime/VoxrGrammarValidator.cs`, two carry no index — `:32` `"the grammar is null or empty"` and `:49` `"empty array"`. The first is unreachable through `SetGrammar`, which short-circuits on `string.IsNullOrEmpty` first (`VoxrSpeechRecogniser.cs:450`). The second is not a corner case: `[]` is one of the four measured SIGSEGV shapes this fix exists for. The message interpolates only `{reason}` into otherwise static text (`:454-465`), so for `[]` the shipped log names no position at all.

Arguably an empty array *has* no meaningful index — which argues the sentence should be qualified, not that it is satisfied. Accurate: "names the character index wherever a position is meaningful."

---

### What verified clean

- **All four measured SIGSEGV shapes are refused.** `LINE` hand-executed each against the scanner: `[]` → `"empty array"`; `["a"b", "c"]` → `ScanString` returns on `'b'`, neither `,` nor `]`; `{"grammar": …}` and `not json at all` → no leading `[`.
- **Legitimate package output still passes.** `GenerateGrammarJson` emits `[`, entries quoted and joined by `", "`, `]`, with `[unk]` added unconditionally — so a bare `[]` is unreachable from it, `[`/`]` inside an entry are ordinary characters to `ScanString`, and multi-word entries are plain. Invariant 2 holds.
- **The validator is total.** Every index read is bounds-guarded (`i + 1 >= s.Length` before `s[i+1]`; `i + 5 >= s.Length` before the four hex digits), every loop body advances `i`. No throwing input found.
- **Null/empty still clears the grammar.** `!string.IsNullOrEmpty(…) &&` short-circuits before the validator. This also keeps `Documentation~/command-recognition.md:736` true, which documents `SetGrammar` with an empty grammar as the free-dictation escape hatch.
- **Ownership rejection still first and still single** — `RejectIfBridgeOwnedByOther` at `:439` precedes the guard at `:449`, and `FireError` only invokes `OnError` without logging.
- **The structural claim holds.** `CLAIM` and `GONE` independently established it: `vosk_recognizer_new_grm` has one call site (`EditorMicBackend.cs:282`, reached only from `VoxrSpeechRecogniser.cs:471`); `BridgeNative.vosk_bridge_set_grammar` has one (`:477`); both P/Invoke declarations are `internal` and `InternalsVisibleTo` names only first-party assemblies, so no integrator route bypasses the guard. The native `vosk_bridge_set_grammar`'s only C caller is `NativeBridge~/harness/main.cpp:186`. Deviation 1's justification is sound.
- **No public API change** — the only new type is `internal static VoxrGrammarValidator` with one `internal` method.
- **Nothing lost by interception.** `GONE` enumerated it: a refused grammar no longer reaches `EditorMicBackend`'s `NotInitialised`/`AlreadyRunning` rejections (no caller reads those codes; `GrammarManager.ForceApply` stops recognition itself first) or `WarnOnUnknownWords` (a string failing the array check has no reliable word list to warn about). `GrammarManager`'s unconditional `IsApplied = true` is pre-existing, not introduced here.

- `[""]` is accepted (pinned deliberately by `Accepts_EmptyElementString`). Valid JSON, and `GenerateGrammarJson` cannot emit it, but it was not among the shapes measured against libvosk 0.3.45 — so whether a single empty entry is as fatal as `[]` is untested either way.
- The `element` counter in the reason strings is 0-based, so `["fire", 42]` reports "element 1" for what a reader would call the second element. Internally consistent with the 0-based character indices; presentation only.
- `Documentation~/api/speech-recogniser.md:81` describes `SetGrammar` in one table row with no mention of the new rejection path. Not false — incomplete.
- `Rejects_ShortUnicodeEscape` (`Tests~/Editor/VoxrGrammarValidatorTests.cs`) is actually a bad-hex-digit case, not a short escape. Naming only.
- A backslash or tab authored into a pattern literal or slot value is now hard-rejected where it previously reached VOSK. Both produce genuinely invalid JSON, so this is the fix working — but no claim mentions it.

### Not verifiable from here

- **The Unity test counts** (EditMode 182/182, PlayMode 637/637). These agents cannot run Unity — the figures came from a separate host-project run and were not re-derived. What *was* checked against the diff: the case counts match (27 and 4), all four SIGSEGV shapes are present as rejections, and `VoxrSpeechRecogniserOwnershipTests` still holds its 8 methods.
- **On-device (Quest) behaviour.** The PR asserts none is needed; the diff supports that — `NativeBridge~/` and `Runtime/Plugins/` are absent from it, and nothing native, mic-related, or lifecycle-related is touched.

No blockers, so nothing was fixed on the branch. Both findings are the human's to rule on.

jinwoo1601:
## Both findings addressed — plus a measurement that corrects one of my own claims

Pushed as `deeb86c`. Suites re-run green: **EditMode 185/185** (was 182 — three new cases), **PlayMode 637/637**.

### Finding 1 — PlayMode validation box overstated the fixture
Fixed in the PR body. It now says the two rejection cases are mutation-sensitive and the other two are controls, and flags that the earlier wording was wrong. No code change — the finding was about the claim, not the tests.

### Finding 2 — "where in the string it was wrong"
CHANGELOG bullet revised in place: the reason names the character index *wherever a position is meaningful*, and the fault alone where one is not, as with the empty array. Same fix applied to the PR body's wording.

### Notes — all addressed, one by measurement

I built a probe against the vendored **libvosk 0.3.45** + `vosk-model-small-en-us-0.15` (one grammar shape per process, exit status read) and validated it by reproducing all six rows of #150's own table exactly before trusting it. Three results:

**`[""]` does not crash** — it returns non-NULL, as does every one of the 12 shapes the validator accepts. The open note is settled: accepting the empty element string is correct, and no logic changed on its account.

**Three crashers #150 never found.** A whitespace-only string, an unterminated string (`["a`) and a bare `[` each SIGSEGV. All three were already refused, so nothing had to change — but their rejection is now known to be load-bearing rather than tidiness, and the CHANGELOG and test comments say so.

**One claim in my posted review was wrong.** I wrote that trailing content, an invalid backslash escape and a raw control character "produce genuinely invalid JSON, so this is the fix working." libvosk tolerates all three — they return non-NULL. So the guard is *deliberately stricter than the decoder* there, and refusing them is a choice, not a crash averted. I kept the strictness and documented it in the CHANGELOG and in the validator's comment; the PR body now has a section laying out the trade-off, since it is the one place this PR changes behaviour for a grammar that previously worked. Happy to loosen it to mirror the decoder if you'd rather.

Remaining notes: the ambiguous 0-based element ordinal is gone from the reason strings (character index alone now — unambiguous, and less code); `Documentation~/api/speech-recogniser.md`'s `SetGrammar` row now names the rejection path and that `null`/`""` still clears the grammar; `Rejects_ShortUnicodeEscape` was misnamed and is now `Rejects_BadHexDigitInUnicodeEscape`, with a genuine truncated case (`["a\u1`) added alongside it — and thanks to the check that caught my first attempt at that case, which pinned a different input but still landed on the same branch.

Still not verifiable from here: on-device Quest behaviour (the diff touches nothing native, mic-related, or lifecycle-related, and the `.so` is byte-identical). Nothing merged, nothing approved.

PR #158, opened 2026-09-10
Expose the input level and ship a Voice Check sample (#147)

### LINKED ISSUE #147: Expose the input level and ship a Voice Check sample
## Summary

The bridge exposes a rolling input-level meter, `vosk_bridge_get_input_level()` (~300 ms RMS of pre-DSP audio, linear 0..1), bound in C# at `Runtime/Native/BridgeNative.cs:64` and documented in `Documentation~/native-bridge.md` as having no runtime consumer. The Editor backend has `EditorMicBackend.ComputeRms` (`Runtime/EditorMicBackend.cs:542`). Neither reaches a game.

A first-run "voice check" is the standard way to catch the two silent failures the troubleshooting guide lists first: no microphone permission or no audio reaching the recogniser, and a mic gain too low for the AGC to rescue. Expose the level and ship a sample that uses it.

## Proposed change

- Public `float InputLevel { get; }` on `VoxrSpeechRecogniser`: bridge level on device, Editor RMS in the Editor, `0` when not running. Read on the main thread; no allocation.
- Sample `Samples~/VoiceCheck`: a scene with a level meter, a "say this phrase" prompt drawn from the active grammar, and a pass/fail readout (level seen above a threshold; a command recognised). Registered in `package.json` `samples`.
- Optional: `OnInputLevelLow` is **not** added; games poll the property.

## Acceptance

- PlayMode test through the WAV-replay path: level rises above zero during a fixture and returns to zero after stop.
- Editor: property reflects live mic RMS.
- Device: reported as deferred (human-only) per the verification bindings.
- Docs: `Documentation~/api/speech-recogniser.md` (new property), `Documentation~/troubleshooting.md` (point the "no transcription" entries at the sample), `CHANGELOG.md` under Unreleased (Added).

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), critique item R40 (verified zero consumers).

jinwoo1601:
Closes #147

## What

**`VoxrSpeechRecogniser.InputLevel`** — a new public `float`, a rolling ~300 ms RMS of pre-DSP audio, linear `0..1`. Reads `0` when nothing is being captured or replayed, `0` before `InitialiseAsync` has built the Editor backend, and `0` on a component that does not own the process-wide bridge (the same `OwnsBridge` guard `IsInitialised`/`IsRecognising` already carry, #57). Allocates nothing; safe to poll every frame; main thread only.

- **Device:** returns `BridgeNative.vosk_bridge_get_input_level()`, wrapped in the file's existing `_bridgeAvailable` latch + `catch (DllNotFoundException)` pattern.
- **Windows Editor:** `EditorMicBackend` now computes the level itself, using the *identical* EMA the C++ uses (`alpha = 1 - exp(-chunk_ms / 300)` over the RMS of the pre-DSP 48 kHz chunk). Because alpha is derived from each chunk's own length rather than assumed, the ~300 ms window holds whatever chunk size the producer delivers — which is what makes this real parity, since the Editor mic path's chunk length varies with frame rate.
- Zeroed at the four points the native zeroes at (mic `Start`/`Stop`, `StartPlayback`/`StopPlayback`) and deliberately **not** on `Reset()`, matching `vosk_bridge_reset`.

**`Samples~/VoiceCheck`** — a first-run mic check: a live meter driven by `InputLevel`, a "say this phrase" prompt, and two independently latched checks (audio seen above a threshold; a command actually recognised). Registered in `package.json` and the root README table.

## Why

The native getter has shipped since the push-audio tier as a **dead binding** — bound in C#, called by nothing. Meanwhile the two failures `troubleshooting.md` lists first (no permission / no audio reaching the recogniser, and gain too low for the AGC) are both silent and look identical from a game's side. The property plus the sample make them distinguishable.

## Notable decisions

- **No `IsRecognising` guard on the property.** Both backends zero their *own* level on stop, so the guard is unnecessary — and it would be actively wrong: the WAV-replay path runs with `IsRecognising == false` by design, so a guard would make the level read `0` for an entire fixture and the issue's own acceptance criterion unsatisfiable.
- **`PreAgcRms`/`PostAgcRms`/`AgcGain` untouched.** They stay instantaneous, post-downsample 16 kHz values for the Editor debug window. `InputLevel` is an additive third quantity, not a redefinition.
- **No `OnInputLevelLow`**, per the issue: games poll the property.
- **No new public grammar API.** The sample authors its own three-phrase grammar through the public `VoxrCommandDefinition`/`Configure` surface and renders the prompt from it, because there is no public way to read the active grammar back.
- **No public API removed or changed**, no new error code, no behavioural change to recognition, parsing, scoring, or what fires.

## Validation

- [x] **Unity EditMode — 186/186 passed**, 0 failed / 0 inconclusive / 0 skipped (host project `VoXR TestGround`, Unity 6000.4.7f1).
- [x] **Unity PlayMode — 638/638 passed**, 0 failed / 0 inconclusive / 0 skipped.
- [x] Both new tests confirmed **executed and passed** by reading the NUnit `<test-case>` nodes, not inferred from suite totals: `InputLevel_UninitialisedRecogniser_ReturnsZero` (EditMode) and `Playback_LoudSine_RaisesInputLevelThenReturnsToZero` (PlayMode, 0.84 s). Test assemblies confirmed recompiled during the run, so the passes come from current source.
- [x] `NativeBridge~/` **not modified** — no arm64 rebuild, no desktop harness run; the shipped `.so` is byte-identical.
- [x] Scene validated structurally: every local `fileID` reference resolves to a declared anchor, `m_Component`↔`m_GameObject` and `m_Children`↔`m_Father` verified both ways, all 10 `m_Script` guids resolved against real `.cs.meta` files, and all 10 serialized keys matched against the C# field names.
- [ ] **On-device (Quest) check needed** — see below.

## What is *not* covered

Stating this rather than glossing it:

- **The native branch of `InputLevel` has no automated coverage at all, and none is possible here.** The WAV-replay path is `UNITY_EDITOR_WIN`-only and reaches libvosk directly, never `libvosk-bridge` — so both new tests exercise only the Editor branch. The device branch is one `try`/`catch` around a getter that was already verified working in C++, but it has not been executed.
- **The sample scene has never been opened in Unity.** `Samples~` is excluded from asset import by its `~` suffix, so nothing in CI or the test run compiles the scene or `VoiceCheckDemo.cs`. The structural validation above is thorough but cannot catch a compile error or a layout that simply looks wrong. **Please open it once before merging.**
- **On-device verification of the meter is outstanding** and is human-only per the verification bindings.

## Pre-existing issue noticed, not fixed

`Editor/VoxrDebugWindow.cs:231` comments its meter scale as `0.0–0.5 RMS maps to full bar` but the line below is `Mathf.Clamp01(rms * 3f)`, which puts full scale at `0.333`. The comment contradicts its own code. Predates this branch; left alone.

https://claude.ai/code/session_01LWb8hCSPtQ6gH6C5E5RZNF


jinwoo1601 (commented):
## Independent review — PR #158

**Profile:** issue-lane **slim** (`CLAIM`, `LINE`, `GONE`), widened with **`XFILE`** (the hand-authored scene binds serialized field names across a file boundary) and **`PIT`** (MonoBehaviour lifecycle, a main-thread-only polled property). Five finder agents, fresh context, then one adversarial verifier per numbered finding.
**Head reviewed:** `ff13006a0a5efb3f458354cdf53ac794e1c1da40`.

`GONE` and `PIT` returned **no findings**. `XFILE` cleared the whole scene↔code contract — all 10 serialized keys match by name *and* type, all 10 `m_Script` guids resolve, the `Button`→`Retry` persistent call is well-formed, `coverageWeight` is correctly absent, and both minted guids are unique. The EMA parity with `vosk_bridge.cpp` was re-derived independently and is formula-exact. Validation ticks #4 and #5 were re-checked from scratch and hold.

Five findings, all CONFIRMED. **None is a blocker**, so nothing was fixed — all five are yours to rule on.

---

### 1. `Retry()` reports a healthy "Listening" state on a stopped recogniser — medium, CONFIRMED

`Samples~/VoiceCheck/VoiceCheckDemo.cs:139` clears `_errorLatched`, then `:146` calls `RefreshStatus()`, which falls through to `:179` `"Listening -- say the phrase above."`. But `StartRecognition()` appears exactly once in the sample, in `Start()` (`:114`) — the button re-arms nothing.

Worst on the sample's headline case: a denied `RECORD_AUDIO` fires `OnError` from `VoxrSpeechRecogniser.cs:371-374` **without** calling `StartRecognitionInternal()`, and nothing re-requests. `InputLevel` then returns `0f` forever, so check 1 can never latch. `troubleshooting.md:33`/`:41` now send users here specifically to distinguish "no audio at all" from "audio arriving but unrecognised" — and after one click the diagnostic asserts the exact healthy state it was imported to disprove. On Quest there is no Console, so the on-screen text is the only signal.

Two defensible fixes, and the choice is a design call: have `Retry()` re-issue `StartRecognition()`, or stop it clearing `_errorLatched` at all.

### 2. Three new copies of an arithmetically false meter scale, one Inspector-visible — minor, CONFIRMED

The PR body itself certifies the sentence false ("puts full scale at `0.333`. The comment contradicts its own code"), then the same diff ships three fresh copies: `VoiceCheckDemo.cs:45` (inside a `[Tooltip]`, so it is Inspector-visible), `:154`, and `Samples~/VoiceCheck/README.md:86`. With `meterFullScale = 3f`, `Clamp01(level * 3f)` saturates at `0.333`, not `0.5`.

The neighbouring "speech rarely exceeds 0.3" makes it worse rather than better: it advertises headroom above ordinary speech when the true ceiling is `0.333` — effectively none, so the meter pegs on normal talking.

To be fair to the PR body: "Predates this branch; left alone" is *literally true* — it refers to `VoxrDebugWindow.cs:231`, which is genuinely untouched. The accurate charge is **incomplete disclosure**: a reader of that section concludes the bad arithmetic is confined to one legacy line.

### 3. Sample README claims two different quantities "mean the same thing" — low (doc), CONFIRMED

`Samples~/VoiceCheck/README.md:95`: "The same `* 3f` scaling is used by the Editor's Command Debug Window level meters, so a reading here means the same thing as a reading there." The window is fed `PreAgcRms`/`PostAgcRms` (`VoxrDebugWindow.cs:209-210`) — instantaneous, post-downsample 16 kHz, one of them post-gain — while the sample's bar is `InputLevel`, a 300 ms EMA of pre-DSP 48 kHz.

It fails even on the charitable "same visual scale" reading, measured two ways: the decimating FIR's 15 coefficients sum to **1.3358** (`Runtime/Dsp/Downsampler.cs:23-27`), so a steady tone reads ~34 % higher on the Pre-AGC meter; and the Post-AGC meter parks near 38 % fill for almost any input (`Agc.cs:28`, `MaxGain = 20f` toward `-18 dBFS`), diverging up to ~12× on quiet speech. The PR body contradicts the README verbatim, and `api/speech-recogniser.md` and `editor-testing.md` both side with the PR body — the sample README is the only place that conflates them.

### 4. The CHANGELOG's reset-parity claim is false for a live session — low (doc), CONFIRMED

`CHANGELOG.md:12`: "matching `vosk_bridge_reset`, deliberately *not* on `Reset()`." On a **running** session `vosk_bridge_reset` zeroes the level twice, transitively: `vosk_bridge.cpp:304` `vosk_bridge_stop()` → `:291` store `0.0f`, then `:311` `start_internal()` → `:179` store `0.0f`. `EditorMicBackend.Reset()` touches `InputLevel` on no path, so the Editor meter carries through uninterrupted. Both arms hang off the one public `ResetRecogniser()` (`VoxrSpeechRecogniser.cs:446-464`).

The literal reading — `vosk_bridge_reset`'s own body has no `g_input_level.store` — is true, and is probably how the sentence got written; it does not rescue a claim whose whole job is to justify an observable-behaviour omission. Impact is small: `ResetRecogniser()` has **no caller anywhere in the package**, and the device recovers in ~300 ms. Arguably the Editor's behaviour is the nicer one — but the changelog says they agree, and they don't.

### 5. The new EditMode test does not pin what its comment claims — low, CONFIRMED

`Tests~/Editor/AudioMetricTests.cs:88` comments `// _editorBackend is null before InitialiseAsync`, but `InputLevel`'s **first** statement is `if (!OwnsBridge) return 0f;` (`VoxrSpeechRecogniser.cs:94-95`), and `s_bridgeOwner` is static and set only inside `InitialiseAsync` — there is no `Awake`/`OnEnable` in the type — so a freshly `AddComponent`ed instance returns there and never reaches the null-backend path. The comment is a copy-paste from `:39`, where it *is* the operative cause, because `EditorPreAgcRms` has no ownership guard.

It is also non-discriminating: delete the `OwnsBridge` guard and `_editorBackend?.InputLevel ?? 0f` still yields `0f`, so the assertion survives. Note `VoxrSpeechRecogniserOwnershipTests.cs` pins non-owner behaviour for `IsInitialised` on seven paths and never once for `InputLevel`.

**(b), same family:** issue #147's acceptance item *"Editor: property reflects live mic RMS"* is exercised by nothing — `grep -rn "\.Tick(\|Microphone" Tests~/` returns zero hits; the only Editor coverage is the WAV-replay path. It is genuinely **not** automatable here (`Tick` reads `Microphone.GetPosition` with no seam; `Start` bails with no device), so the fix is one line in the deferral list, not a test. The PR's "What is *not* covered" section opens "Stating this rather than glossing it" and then omits it. Real risk is low: the EMA lives in `ProcessChunk`, which playback does cover; the mic-only residue is the two `InputLevel = 0f` stores.

---

### Considered and refuted

- **"Both new tests exercise only the Editor branch" is false** — REFUTED. Read in its own sentence it asserts neither test reaches the *native* branch, which is true. The PlayMode test is correctly characterised: `SetUp` awaits `InitialiseAsync` and asserts `IsInitialised`, so it does hold ownership and does read the backend.
- **`EditorMicBackend.cs:63`'s comment overstates reset parity** — REFUTED. Its referent is the getter's formula, which genuinely matches; only the CHANGELOG sentence (finding 4) overreaches.

- `Release()` and a stop-while-idle can leave a stale non-zero level readable; unobservable today because both callers null the backend on the next line.
- The sample README's "What's in the scene" table omits `Canvas/Title` and `Canvas/PromptLabel`.
- The scene is Screen Space–Overlay with a plain `GraphicRaycaster` while its README promises Quest controller-raycast input — identical to all three shipped samples, so pre-existing convention, not introduced here.
- The scene pulls `com.unity.inputsystem` via its `EventSystem` while `package.json` declares no dependencies — again identical to the three existing sample scenes.
- `Runtime/Native/BridgeNative.cs:50` still names the abandoned Tier D rig as the push seam's consumer; pre-existing, and this PR's doc rewrite now names the desktop harness instead.
- `Documentation~/editor-testing.md:15` lists the debug window's three meters and is the natural place a reader would look for `InputLevel`; it doesn't mention it.

### Not verifiable from here

- **On-device (Quest) anything** — the native `InputLevel` branch, mic capture, permission flow. Human-only per the project's verification bindings; no agent claimed otherwise.
- **The Unity suites were not re-run** by this review. Ticks #1–#3 were audited for *consistency* only: both named tests exist verbatim, are correctly attributed, compile against public API, and the PlayMode test's asserted bound follows from its own arithmetic (10 chunks, α ≈ 0.2835, level ≈ 0.341 ≫ 0.05).
- **Whether the scene opens correctly in Unity.** `Samples~` is never imported by any automated run, so the static checks above are the ceiling — they cannot catch a compile error or a bad layout. Worth one Editor open before merge.

https://claude.ai/code/session_01LWb8hCSPtQ6gH6C5E5RZNF

jinwoo1601:
## All five review findings addressed — `531f8f6`

Maintainer ruled "address all findings, take your judgement", so all five are fixed rather than left for triage. Three were judgement calls worth stating:

**Finding 1 — `Retry()`:** fixed by **re-arming**, not by suppressing the error text. The button says "run again", so it should. Guarded on `IsRecognising`, which is load-bearing: an unguarded restart hits `ALREADY_RUNNING` on the device path and would fire a spurious `OnError`. Verified the guard is also safe against `Start()`'s existing `StartRecognition()` call — `InitialiseAsync` has an `_initialising` re-entry guard, so the second call is a no-op, and if the model is already ready the first start completes synchronously and the guard skips.

**Finding 4 — reset parity:** fixed the **doc, not the code**. The device zeroes because `vosk_bridge_reset` genuinely restarts capture; making `EditorMicBackend.Reset()` mimic that would fake an artifact rather than achieve parity of meaning. The CHANGELOG now records the divergence explicitly.

**Finding 5:** comment only, **no test added**. A discriminating Editor-branch test cannot exist — owning the bridge is what creates the backend, so "non-owner with a backend" is not a constructible state. The guard's real work is on the device branch, and the comment now says so.

Findings 2 and 3 were text-only. Scope note: finding 2 also touches `Editor/VoxrDebugWindow.cs:230`, which is outside this issue — but the sample cites that line by name, so leaving it would have shipped a self-contradiction. Comment only, no code.

### Verification

- **EditMode 186/186, PlayMode 638/638**, 0 failed / 0 inconclusive / 0 skipped. Both #147 tests confirmed `Passed` + `runstate="Runnable"` from the `<test-case>` nodes; test assemblies confirmed recompiled in-run (`CacheMiss` → `CacheWrite` on `Tests.Editor.dll`).
- The compiled delta since the last green run is **comment-only**, and I want that on the record rather than implied away: the suites confirm nothing broke, not that the behavioural fix works.
- **`VoiceCheckDemo.cs` was compiled out-of-band** — 0 errors, 0 warnings, against Unity 6000.4.7f1's engine modules plus the real `Jinwoo1601.VoXR.Runtime.dll`. This is the first time anything has compiled that file; `Samples~` is excluded from Unity's importer, so no gate in this repo reaches it. The harness was mutation-verified because a `~` in the path is exactly what MSBuild silently globs to nothing: removing a `;` produced `CS1002` at the `Retry()` call site, and renaming `IsRecognising` produced `CS1061` against the real assembly, so member lookup genuinely binds.
- **Still unverified:** the `Retry()` behaviour itself. It compiles and its members resolve, but nothing has *run* it. The scene was opened and passed before this commit, so that pass predates the change — one more Play through the sample is worth it before merge.

https://claude.ai/code/session_01LWb8hCSPtQ6gH6C5E5RZNF

PR #159, opened 2026-09-15
Add the slot resolver hook (#148)

### LINKED ISSUE #148: Slot resolver hook: fill an omitted required slot from game state (ellipsis)
## Summary

Let a game fill an omitted required slot from game state, so "launch missiles" can mean "launch missiles target &lt;the main target&gt;" when the game knows there is one obvious target. This is ellipsis in the narrow sense ruled on 2026-09-06: **game-state resolution of an omitted slot**, never "reuse the last-mentioned value". The package ships the hook; the game decides what "the obvious target" is.

## Definition (from the maintainer's own game note)

A track number becomes optional in an aggressive order when one obvious target exists: the player designated one ("track 1721 is the main target"), interception implicitly highlighted one, or the encounter has only one hostile. That knowledge lives in the game, so the package must not guess it.

## Proposed change

- `VoxrCommandRecogniser.RegisterSlotResolver(string slotName, Func<VoxrSlotResolution> resolver)` and `UnregisterSlotResolver`, alongside the existing `RegisterSlotValueProvider` (`Runtime/Commands/VoxrCommandRecogniser.cs:330`). `VoxrSlotResolution` carries the value (or none) and a short `Reason` string ("main target", "only hostile") so crew readback can name what it filled.
- When a parsed command is complete except for one or more required slots, each missing slot's resolver is consulted **before** the incomplete ruling. Resolved → the command is complete and proceeds through the normal gates. Not resolved → today's behaviour (pending with `allowPartialMatch`, otherwise rejected). `VoxrCommand` exposes which slots were resolver-filled and the reason.
- The resolver is consulted only for a round that actually produced a candidate. A barred round is never offered to a resolver, and a resolver can never supply a pattern's first required element (ruling 3: the bar is absolute). Concretely: if the missing slot is the pattern's anchor, the round is barred already and no resolver runs.
- Resolver runs on the main thread at parse time, so it must be cheap; no grammar change, no rebuild.

## Acceptance

- Tests entirely through `InjectText`: (a) resolver returns a value → command fires with the slot filled and the reason recorded; (b) resolver returns none → pending/reject exactly as today; (c) a pattern whose first required element is the resolvable slot, spoken without it, stays barred and the resolver is never called; (d) a slot with both a value provider and a resolver keeps its provider semantics for grammar and matching.
- Session log records resolver-filled slots distinguishably.
- Docs: `Documentation~/command-recognition.md` (Pending Commands: a new step before "missing required slot"), `Documentation~/api/command-recogniser.md`, `Documentation~/api/data-types.md`; `CHANGELOG.md` under Unreleased (Added).

## Out of scope

Pronouns and demonstratives, gaze or pointing fusion, and any last-mentioned-target memory in the package. Those are deferred to a design branch after the human corpus exists.

## Provenance

Research pass `Planning~/research/2026-09-05-next-level` (local), item R36a per ruling 9 (2026-09-06). Backed by the maintainer's `Planning~/anaphora-resolution-analysis.md` (integration point: slot-time hook between parse and accept).

jinwoo1601:
Closes #148. Feature lane, design track skipped — ruling 9 (2026-09-06) is the lock. **G2 is your explicit in-conversation ruling**; I never merge.

Two commits: `0d9aceb` the hook and its behaviour, `f61da95` the observability. **No product-doc changes** — those come after G2, per the close-out order.

---

## Read this first — the feature is close to inert for the use case that motivated it

This is measured, not suspected, and it is the thing to decide before anything else in this PR.

On the **fresh-parse path**, a resolver is only offered a candidate that already cleared `minScore`, and a resolver-filled slot deliberately earns **no score** (a slot nobody spoke is not evidence). So an elided utterance must clear the gate on what was actually spoken.

**Scope correction (review F-4).** That premise holds on the fresh-parse path only. The *follow-up* path has no `minScore` gate by design (#77, #113) — its floor is a positive score — so a follow-up fill re-scoring below 0.6 **is** offered to a resolver. The measurement below classifies fresh-parse occurrences only, and is therefore a **floor**, not the whole picture: some utterances it counts as "below the gate" may be reachable in two steps (pend first, then complete via a follow-up plus a resolver). That two-step reachability is **unmeasured** — the verifier assigned to it did not run. With `RequiredSlotMissPenalty = -1.0`, `RequiredLiteralMissPenalty = 0` and `minScore = 0.6`:

> **(a) drop the slot alone: `(D−2)/D`, needs `D ≥ 5`.  (b) drop the slot *and* its orphaned literal — the natural spoken form: `(D−3)/D`, needs `D ≥ 8`.**  (`D` = count of **required** elements; optionals drop out of both sides.)

Measured against the real parser on `VR FTL-Like 3`'s `Set_Combat` (20 live commands, 40 pattern×required-slot occurrences, all three scenes at the shipped `minScore: 0.6`):

| | (a) slot only | (b) natural form |
|---|---|---|
| Resolver reachable | 18/40 (45%) | **5/40 (12.5%)** |
| Clears the gate but loses its round to a rival | 5 | 4 |
| Below the gate | 17 | 27 |
| **Barred by the anchor rule** | **0** | **0** |
| Degenerate (no utterance left) | 0 | 3 |

- The `{track}` aggressive orders, 15 occurrences: **(a) 7/15, (b) 2/15** — and both survivors are `adjust_range`'s 8-element pattern, not an aggressive order. **0 of 4 weapon-release occurrences are reachable in natural form.** `"launch jackal"` scores **0.250** — identical to #148's own example, so the issue's example is not an unlucky pick.
- A majority of `{track}` orders would need `minScore ≤ 0.25`, i.e. effectively no gate. `minScore` is one global `[SerializeField]`, shared with the sibling-tie scan and the eager gate.
- **The anchor rule costs this grammar nothing** — not one pattern opens with a required slot. I counted it separately expecting it to matter; the bucket is empty.
- The grammar already solves this for `{burn_level}` **with no resolver**: a slot-less sibling pattern matches the elided utterance at 1.000 and wins the round. That idiom costs a second pattern and a per-handler fallback, with no uniform record and no way to *ask* when the game does not know.

Provenance: staged from `0d9aceb` into a desktop harness, validated first against the committed 66-claim DocCheck pin, and mutation-verified (moving the slot penalty to `-0.5` moved `"launch missiles"` 0.2500 → 0.3750). Full detail in `Planning~/features/slot-resolver/requirements.md` §6 E-2.

**Not measured:** acoustics, the confidence gate (`Parse()` supplies no word confidences), and which elision shape real speakers produce — the 699-utterance corpus's `delete` arm is single-**word** deletion, not slot elision.

**My recommendation:** merge this — the hook, the record, the readback reason and the ask-when-unknown path are correct — and take the scoring question to a **design branch**, since `scoring-model.md` is locked and complete. (An earlier draft called this "the substrate any fix would need". Review finding F-14 showed that overstates it: option 4 acts at scoring time *inside the parser*, while `_slotManager` is recogniser-private and the parser receives only a materialised array, so the follow-up would have to cross the very boundary E-1(3) deliberately drew. The types and records are reusable; the placement is orthogonal.) A fourth option the measurement surfaced: exempt a resolver-fillable slot from the *denominator* rather than crediting it as matched (`0.667`, not `0.75`) — declining to charge for what the game supplies rather than manufacturing evidence it was spoken. If you would rather this not merge until the feature visibly works, that is a legitimate call.

---

## What it does

`RegisterSlotResolver(string slotName, Func<VoxrSlotResolution> resolver)` / `UnregisterSlotResolver`, shaped after the slot-value-provider pair but contracted differently: one value or none, per utterance, no parser rebuild, filling what the parser did **not** match. New public `VoxrSlotResolution` (value-or-none + a short `Reason` for crew readback) and `VoxrResolvedSlot`. `VoxrCommand` gains `ResolvedSlots` and `GetSlotResolutionReason`.

**Additive:** the public `VoxrCommand` constructor keeps its signature and the **three construction sites outside the struct** are unedited — a private all-fields constructor carries the new field, following the existing `WithScore` precedent. (The fourth site *is* `WithScore`, inside the struct, and it was edited to carry the field through; the PR originally said "all four", which review finding F-15 corrected.)

**The rule that organises it:** *resolution changes exactly one thing, the completeness answer, and may never move a command across any other gate.* That is why it is gated on `minScore` on the fresh-parse path and `Score > 0` on the follow-up path — without those, a command moves pending→rejected and re-armed→refused respectively. My first draft had a different rule and it was false; the conclusions survived, the reasoning did not.

**Uniform across all three sites that consult the completeness ruling**, via a single resolution pass whose result the sites read, so they cannot disagree. Resolving only on the fresh-parse path would make the same utterance behave differently depending on whether a pending happened to be live — invisibly to every instrument in the project.

**The leading-miss bar gets no new gate.** A barred round constructs no `VoxrCommand`, so there is nothing to offer a resolver; ruling 3 holds by structure rather than by a second statement of it that could drift.

## Found by a pre-implementation plan audit, not by tests

Each of these would have shipped:

- **`TryBuildAmbiguity` re-read the pooled buffer**, so `choices[0]` would have been the *unresolved* command while `PendingCommandHandler` fires `Choices[i]`, not `Command`. Answering a sibling tie would have fired a command missing its required slot — the shape #73 refuses.
- **The follow-up path resolved against a definition that is not always the one it fires under** (`:806-813` documents that divergence as reachable). Both definitions must now agree.
- **A resolver that cancels the pending** mid-utterance would have dereferenced an empty `Nullable` on the new read *and* the shipped one. Fixed at the premise — a follow-up fill is meaningless without the pending it merged into — rather than by guarding one read and leaving the other crashing.

## Verification

EditMode **190/190** (186 + 4), PlayMode **664/664** (638 + 26), `Samples~` compile green and mutation-verified. `Samples~` matters because a Unity test run does not compile it at all — a public-API change can break every shipped sample with the suite green. That gate was unbound before this branch; it is now binding 3.

**Mutation evidence — six tests proved capable of failing** (each applied alone, reverted, and confirmed green again; every filtered run reported `testcasecount="1"`, so none was a zero-test false green):

| Test pins | Mutation | Failure |
|---|---|---|
| The bar keeps the resolver unreachable | neuter the barred `continue` | `Expected: 0, But was: 1` — the game *was* asked |
| One question per utterance | drop the memoisation | `Expected: 1, But was: 2` |
| The sibling-tie defect above | re-read the pooled buffer | fired command lacks the slot |
| The `minScore` gate | delete it | rejected instead of pending |
| Step 4 reads the effective command | read `resultBuf[i]` | `RawText` merged into the pending's |
| The cancelled-pending discard | delete it | `InvalidOperationException: Nullable object must have a value` |

## One existing test was edited, deliberately

`SchemaVersion_IsThree` → `SchemaVersion_IsFour`. The exported log's shape changed, and every prior field addition bumped the version (`3493256` #28, `09e91cd` #112, `4576ce8` #144). That test exists **to pin the version**; retargeting it is the correct response to a legitimate bump, not a green-keeping edit. It is the only existing test that differs in this PR.

## Known gaps, stated rather than left to be found

- **F24 (debug-window marking) has code but no test.** The logic is inline in a private `EditorWindow` method that hands its label straight to IMGUI; there is no seam and nothing in `Tests~/` touches that window. An IMGUI harness would be larger than the code under test. A **Should**. The minimal fix, if wanted, is extracting the line formatting to an `internal static` helper — a production-design call I did not take unilaterally.
- **F23 (readme prose)** untested: it would pin a string against itself, and no precedent for asserting on `Readme` exists. A **Should**; the prose is written.
- **F25 is a Must and does not close with this PR.** The `VoxrBatchTestRunner` comment is in; its `KNOWN_LIMITATIONS.md` entry is post-G2 product-doc work.
- **The zero-allocation NFR is not discharged.** `ProcessParsedResultsCore` allocates unconditionally (`text.Split`, plus Editor-only diagnostics), so the resolution step cannot be measured in isolation through `InjectText` without a new internal seam. Claimed nowhere as verified.
- **Slot names are global**, so a resolver on `track` makes `{track}` resolvable on every command using it — including `self destruct target {track}`. No per-command fence exists within ruling 9's per-slot shape; the fences a game has are register/unregister timing, or a different slot name. This leads the product docs.
- **On-device verification: not applicable.** No mic path, no lifecycle change, no native bridge.

## Post-G2, if you accept

`Documentation~/command-recognition.md` (a new Pending-Commands step before "missing required slot"), `api/command-recogniser.md`, `api/data-types.md`, `KNOWN_LIMITATIONS.md` (F25 + the global-slot-name hazard + the reachability arithmetic), `CHANGELOG.md` under `[Unreleased]`, and `Documentation~/editor-testing.md:68-72`, which states the schema version as `3` and enumerates the per-slot fields — both now stale, and it never says "resolver", which is exactly the sweep-for-the-outcome trap.

https://claude.ai/code/session_01Vyb74D5vGjKgrFH7q88w6F

---

# Review fixes — `47c94f5`

The `review-pr` full-profile pass (12 angles, posted above) produced 23 findings. **All are addressed.** Two were maintainer rulings; one turned out to be a wrong *test* rather than wrong code.

## The three severe ones

**A resolver that cancels the pending threw out of the recognition callback.** Found independently by three angles and confirmed by a verifier. The guard added for exactly that hazard sat *above* Step 5's own resolver call — and Step 5 is routinely the utterance's first resolver invocation, because Step 3b resolves nothing when the parse yields no results and stops at the first unresolvable slot when it does. This PR's own shipped test was the reproduction minus the cancel. Fixed at the premise on Step 5's own side; when the pending is gone the branch falls *out* of Step 5 rather than returning, because an early return would hand the speaker silence after the game's own resolver ended the exchange.

**Sibling rivals are no longer offered resolution.** The bar's structural guarantee — a barred round constructs no command, so there is nothing to resolve — covers the round **winner** only. Issue #126 ruled `LeadingRequiredMissed` "recorded, not gated", so a barred rival *does* get a command built, and resolving one would let game state supply the very anchor the bar refused. #126 ruled it acceptable for **the speaker** to supply a missing anchor by answering the question; never for the game to supply it from state. D-11's actual defect (the winner re-read from the pooled buffer) stays fixed.

**The resolution gate needed three floors, not one.** It was gated on `minScore` so a command could not move from *pending* to *rejected*. The confidence and debounce gates sit **below** the same pending branch, so a resolvable command that would have prompted the speaker instead resolved, skipped the branch, and died at a gate it had never been measured against — with `anyThresholdFiltered` suppressing even `OnUnrecognisedSpeech`. Silence where there was a question. The debounce arm was worse: Step 4 does not test cooldown, so such a command cancelled a live pending *on its way to being debounced*.

## Maintainer rulings

**The resolver now receives its asking context** (`Func<VoxrSlotResolutionRequest, VoxrSlotResolution>`, carrying slot name, intent and matched pattern index). Slot names are global, so a `track` resolver was consulted on every command using it — including a destructive one — with no way to tell which was asking. Keying stays per-slot per ruling 9; only the ask gained context. **The memo key widened to match**: keying on the slot alone would serve one intent's answer to another, which is the hazard the change exists to remove. So the bound is no longer "once per slot per utterance" but once per distinct `(slot, intent, pattern)` triple.

**The triplicated completeness walk gets a test now and a refactor later** (#160). Three hand-written copies answer "which required slots are unfilled?" for three consumers — whether a command fires, what the crew asks for, and what the resolver is offered — and two had already silently diverged. A divergence produces no error: it produces commands that don't fire, or fire with wrong arguments, or exchanges that ask for the wrong thing.

## Also fixed

Three invariants could previously be **deleted with the whole suite green** and now have tests: the per-utterance cache clear (the no-memory rule ruling 9 forbids breaking), D-6's follow-up score floor, and the completeness mirror. The resolver's re-entrancy contract is now stated (`CancelPendingCommand` is safe; `InjectText`/`Configure`/`SetActiveSets` are not). Two comments claimed things that were not true — the no-allocation claim and *"a single pass, so they cannot disagree"* — and now state what is. `PendingCommandHandler.TryFollowUpSlotFill` silently dropped `ResolvedSlots`; unreachable today, silent when it isn't. Plus the debug window's pending panel, duplicate-slot dedup, and four convention nits.

**One finding was a wrong test, not wrong code.** A pattern naming a slot twice needs **one** record: `HasSlot` is a lookup *by name*, so a single record satisfies every occurrence. The test asserted two, with a justification that was factually false. Both were corrected.

## Verification

EditMode **190/190**, PlayMode **676/676**, `Samples~` green and mutation-verified. No pre-existing test went red — notably none in `VoxrPendingCommandTests`, which is what establishes the `TryFollowUpSlotFill` restructure is behaviour-neutral today.

**Six tests proved breakable** (each mutation applied alone, reverted, re-run green; every filtered run reported `testcasecount="1"`, so none was a zero-test false green):

| Test pins | Mutation | Failure |
|---|---|---|
| The F-1 crash fix | unwrap the Step 5 `if (_pending.HasPending)` | `InvalidOperationException: Nullable object must have a value` |
| The confidence floor | delete that clause from the gate | `Expected: 0, But was: 1` — the resolver was reached |
| The debounce floor | delete that clause | `Expected: 0, But was: 1` |
| The cache clear (no-memory) | delete `_resolutionCache?.Clear()` | `Expected: 2, But was: 1` — the refusal was remembered |
| The widened memo key | narrow it back to slot name | `Expected: 2, But was: 1` — one intent served the other's answer |
| The divergence mirror | drop the `IsOptionalSlot` guard | expected `["track"]`, got `["flavour","track"]` |

**One honest limit on that evidence:** three of those mutations trip a *call-count* assertion first, which short-circuits the rest of the test. So "it goes red" proves the resolver's reachability is pinned — it does not prove the downstream assertions in those tests (the pending still opens, the slot stays unfilled) were exercised by the mutation. They were not reached.

## Still open

- **F-2's reachability is unconfirmed.** The mechanism is fixed regardless, but whether the mixed-anchor grammar shape is actually constructible was never established — that verifier was killed by a rate limit.
- **The E-2 reachability figures are a floor, not a total** (see the scope correction above). The two-step follow-up path is unmeasured.
- **Seven of the ten finding-verifiers never ran.** Most findings in the review above carry only their finder angle's confidence. They should be re-run against this head before G2.
- **F-25 remains the only unmet Must** — its `KNOWN_LIMITATIONS.md` entry is post-G2 product-doc work, per the project's close-out order.


jinwoo1601 (commented):
# `review-pr` — full profile

**Head reviewed:** `f61da952058587bb8455cfd8a5f472ee584e1cbe` · **Base:** merge-base `03831a8` · **Profile:** full (feature lane, G2 at stake) — **all 12 angles ran**: CLAIM, LINE, GONE, XFILE, PIT, WRAP, REUSE, SIMP, EFF, ALT, CONV, GAP. None dropped, none added.

## ⚠ Verification status — read before weighing anything below

The verifier pass was **cut short by an API weekly limit**. Of the findings below:

- **1 cluster was independently verified** by a `finding-verifier` with no prior context (F-1).
- **2 were verified on the main thread by the same session that wrote the code** (F-2, F-3) — traced against source and quoted, but *not* independent. Weight them accordingly.
- **The rest carry their finder angle's own confidence only.** They are reported because suppressing them would be worse, but they have had no adversarial second pass.

Nothing here should be treated as fully adjudicated. Re-running the verifiers is the first thing to do when limits reset.

---

## F-1 · BLOCKER · CONFIRMED (independently verified)

**A resolver that cancels the pending throws `InvalidOperationException` out of the recognition callback.** Found independently by three angles (PIT, WRAP, XFILE).

`Runtime/Commands/VoxrCommandRecogniser.cs:1025` and `:1031`.

The D-15 guard — `if (followUpResult.HasValue && !_pending.HasPending) followUpResult = null;` (`:886`) — is justified by the comment at `:881`: *"Step 3b above is the first game code this method has ever run between Step 2's fill and Step 5's use of it."* **That is false of this PR's own code.** Step 5 invokes the resolver itself at `:954`, *after* the guard, and then dereferences the pending unconditionally at `:1025`/`:1031`.

The decisive evidence is a **shipped, green test in this PR**: `Resolver_CompletesAFollowUpFill_FiresInsteadOfReArmingThePending` (`Tests~/Runtime/VoxrPendingCommandTests.cs:2191`) is the reproduction minus the cancel. It injects `"three"`, which parses to `resultCount == 0`, so Step 3b is skipped by `:850` and the resolver necessarily fires for the first time from the unguarded Step 5 site. That test proves every gate on `:951-960` is passable on this shape.

Both arms throw, from one grammar:
- resolver returns a value → `Complete(followUp, _pending.Current.Value.Definition, …)` throws at `:1031`;
- resolver returns `None` → `AdvanceSlotFill` throws at `PendingCommandHandler.cs:271`.

`Configure()` has the identical hole (both overloads call `InterpretResolution(_pending.Cancel())`).

**A second, wider route:** even when `resultCount > 0`, Step 3b's all-or-nothing pass returns at the *first* slot with no resolver (`:1603`), so a resolvable slot later in pattern order is never asked there either — the `:886` guard is a no-op for it too.

The code contains the contradiction in plain sight: `:830` says *"Step 5 is reachable with resultCount == 0"*; `:881` says Step 3b is the first game code. Both are in this PR.

**Repro:** pend `PartialLaunch` (`UnfilledSlots == {tube, track}`); register a `track` resolver whose body calls `CancelPendingCommand()` then returns a value; `InjectText("three")`.

---

## F-2 · BLOCKER (canon) · premise confirmed on the main thread — NOT independently verified

**A resolver can supply a pattern's first required element, through a tied sibling rival — which ruling 3 forbids absolutely.** (LINE)

`Runtime/Commands/VoxrCommandRecogniser.cs:1481` — the rival-resolution line this PR added under D-11:

```csharp
var rival = parser.BuildSiblingRivalCommand(i, n, tokens, wordConfidence);
if (TryResolveMissingSlots(rival, rivalDef, out var resolvedRival))
    rival = resolvedRival;
```

Invariant 1's structural argument — *"a barred round constructs no `VoxrCommand`"* — covers only the round **winner**. `VoxrCommandParser.cs:3036` is explicit that it does not cover rivals:

> *"`matchResult.LeadingRequiredMissed` is in scope here and is deliberately NOT carried into the record (issue #126, ruled 2026-08-28: recorded, not gated). … Do not read this as 'a leading-missed candidate can never fire': it can, here, by being chosen."*

**What is pre-existing versus what this PR adds.** #126 ruled that a barred rival may be offered as a disambiguation choice and fired when chosen — and the same comment says why that is acceptable where the discriminator is the barred member's anchor: *"the answer SUPPLIES the missing anchor"*. The **speaker** supplies it, by saying it. This PR lets the **game** supply it, from state, with no one having said it. Ruling 3's wording is *"the accepted transcript must contain the first required element"* — a resolver-filled anchor means it does not.

Nothing carries `LeadingRequiredMissed` onto `VoxrCommand`, so `TryResolveMissingSlots` walks the rival's pattern from `p = 0` with no way to know its element 0 was barred.

**This also makes a shipped public XML guarantee false** (`:370`): *"it can never supply a pattern's first required element: a command whose leading required element went unspoken is refused before any command is built, so there is nothing to offer a resolver."*

I could not finish establishing whether the specific `{?track}`/`{track}` mixed-anchor grammar shape actually ties — that was the verifier's remit and it did not complete. **The mechanism is confirmed; the reachability is not.** If the shape is unconstructible this collapses to a documentation fix; if it is constructible, it is a canon violation that a design branch owns, not this PR.

---

## F-3 · MAJOR · confirmed structurally on the main thread — NOT independently verified

**Resolution moves a command across the *pending* gate — into silence.** (GAP; ALT-4 is the shape-level version of the same concern.)

Step 7's pending branch is `if (cmd.Score < minScore || incomplete)` at `:1167`. The confidence gate is at `:1218` and the debounce at `:1231` — **both below it**. Verified by reading.

A candidate above `minScore`, incomplete, on an `allowPartialMatch` definition, with word confidence below `minConfidence`:
- **before:** `incomplete` is true → enters the branch → `EnterPending` → `continue`. The confidence gate is never reached. **The speaker is prompted.**
- **after, with a resolver:** resolves → complete and above `minScore` → branch not entered → falls to `:1218` → `anyThresholdFiltered = true`, `continue`. No command, no pending, no prompt — and `anyThresholdFiltered` suppresses `OnUnrecognisedSpeech` too. **Silence.**

Same shape for a resolved command whose intent is on cooldown.

This is exactly the class D-5 exists to forbid. D-5 guards the score gate; the same mechanism one gate later is unguarded. The architecture's justification (*"score and confidence untouched by the copy-with, so no candidate moves across those gates"*) reasons about the carried **values** — the defect is about which **branch becomes reachable**.

No test covers it: every resolver fixture sets `CommandCooldown = 0f`, and the only resolver test supplying real confidences uses `0.9f`, above the gate.

---

## F-4 · MAJOR · unverified — the PR's own framing is wrong

**"A resolver is only offered a candidate that already cleared `minScore`" is false on the follow-up path.** (CLAIM)

The fresh-parse path gates at `:867`; the follow-up path gates only at `&& followUp.Score > 0f` (`:953`), and the comment there says so: *"This path has no minScore gate by design (#77, #113)."* So a follow-up fill re-scoring at 0.25 **is** offered to a resolver — the range the PR body reports as unreachable.

The E-2 reachability measurement classifies **fresh-parse** occurrences only. Nothing in the body says so. The figures are not wrong; their scope is narrower than stated, and the "close to inert" framing rests on the unqualified version. A two-step path (pend below the gate, then complete via follow-up + resolver) may be reachable where E-2 reports "below the gate" — **unverified, and it is the question that decides how much the E-2 conclusion needs revising.**

---

## F-5 · MAJOR · unverified

**The resolver delegate is context-free, so the global-slot-name hazard cannot be fenced by the game.** (ALT)

`Func<VoxrSlotResolution>` (`:376`), invoked as `resolver()` (`:1537`), keyed by slot name alone. Slot names are global, so a `track` resolver is consulted on **every** command using `{track}` — including `self destruct target {track}`. Requirements §6 E-1 calls this "the feature's sharpest edge" and offers only register/unregister timing or renaming the slot: a stateful timing dance imposed on a game by a package whose defining property is holding no state.

The finding argues this is *not* the per-command hook requirements §9.2 closes off — keying would stay per-slot, only the **ask** gains context — and that it is the same latitude already taken in departing from ruling 9's `Func<string>` to carry a reason. **Whether ruling 9 forbids context on the ask or only fixes the keying is the crux, and it is unverified.** It matters now rather than later because the shipped signature is the one integrators write against.

---

## F-6 · MAJOR · unverified

**The "no memory" invariant — the thing ruling 9 forbids outright — has no executable guard.** (GAP)

`_resolutionCache?.Clear()` (`:835`) is the only barrier to cross-utterance memory. **Delete it and every test in this PR still passes.** All thirteen new resolver tests register constant-answer resolvers, unregister before a second utterance, or assert a call count after exactly one resolving utterance. None has a resolver whose answer *changes* between utterances. A permanently-stale cache — including a cached `None` freezing a slot as unresolvable for the session — is invisible to the suite. The PR's mutation table pins only the opposite direction.

Code believed correct; the invariant rests on prose.

---

## F-7 · MAJOR · unverified — a stated invariant that is false

**The no-allocation claim does not hold once a resolver is registered.** (EFF)

The field comment at `:186` claims no allocation for *"every utterance in a game that registered no resolver, **and most of them in one that did**."* The first half holds (`HasResolvers` short-circuits). The second does not: the walk calls `VoxrCommandParser.ExtractSlotName` (`VoxrCommandParser.cs:4355`), a substring factory — one heap string per slot element, **two** for an optional `{?foo}`, allocated before `IsOptionalSlot` discards it on the next line. Paid per above-gate candidate per utterance, including commands the speaker said in full. Two further sites pay it again (follow-up, and per rival in `TryBuildAmbiguity`).

Small in magnitude; it is the stated invariant that is wrong. Fix is either an allocation-free slot-name path or a narrowed comment.

---

## F-8 · MAJOR · unverified

**`TryResolveMissingSlots` is the third hand-written copy of the same guarded pattern walk.** (REUSE and ALT independently.)

Alongside `VoxrCommandParser.HasUnfilledRequiredSlot` (`:4387`) and `PendingCommandHandler.ComputeUnfilledSlots` (`:395`). The sharpest part: **the two pre-existing copies have already silently diverged** — `ComputeUnfilledSlots` carries only two of the three guards. This PR makes a divergence *correctness*-bearing rather than cosmetic, and its own comment says so: *"Any divergence there is a command that resolves and still does not fire, or fires still missing an argument."* Nothing enforces the mirror — no shared enumerator, no test that the two agree — and architecture §5.2 records that the first draft of this very method copied only two of three guards.

The angles split on severity (REUSE: minor; ALT: major).

---

## Minor

- **F-9 · Re-entrancy corrupts in-flight resolution state.** (PIT, LINE, GONE, WRAP.) A resolver calling `InjectText` re-enters and clears `_effectiveBuf`, `_resolutionCache` and both scratch lists mid-walk; the outer frame then reads `default(VoxrCommand)` (null `Slots`) or materialises a command carrying the inner utterance's slots. The existing "handlers should queue" caveat is written for *event subscribers*; the resolver contract says only "must be cheap". Also `TryBuildAmbiguity` (`:1481`) now calls game code *between* two reads of the parser's pooled rival buffers, so `choices[n]` and `choiceValues[n]` can come from different utterances.
- **F-10 · "Uniform … so they cannot disagree" overstates it.** (CLAIM, ALT.) Steps 4 and 7 read one pass; the follow-up site runs a **second**, independently gated resolution against a different command and definition. What makes them agree is the per-utterance memo, not a single pass. The same overstatement is in the Step 3b comment (`:815`). Relatedly, the uniformity is a convention — "never read `resultBuf[i].Command` after Step 3b" — stated nowhere checkable; this PR's own audit caught one violation of it.
- **F-11 · The public XML doc's precondition is wrong twice.** (ALT, LINE.) `:370` omits the `minScore` gate — the one that decides the outcome in the measured majority (E-2: 17/27 below the gate versus 0 barred) — and makes the categorical bar claim F-2 disputes.
- **F-12 · D-6's `Score > 0f` floor has no test.** (GAP.) Its twin D-5 got a named regression guard; delete this one and the suite stays green.
- **F-13 · The debug window's *pending* panel is unmarked.** (XFILE.) `Editor/VoxrDebugWindow.cs:361` prints a resolver-filled slot indistinguishably from a spoken one, while the attempt panel above it marks them. Outside F24's letter, same hazard.
- **F-14 · "The substrate any fix would need" is overstated.** (ALT.) Option 4 acts at scoring time inside the parser; `_slotManager` is recogniser-private and the parser receives only a materialised array. The types are reusable, but the placement is orthogonal to the recommended follow-up rather than a step toward it.

## Nits

- **F-15** "All four construction sites are unedited" — three are; the fourth *is* `WithScore`, which this PR edited (`VoxrCommand.cs:95`). Substance intact, count wrong.
- **F-16** The PR body drops a row when reproducing E-2's table, so the columns no longer account for their denominator; and it says "the 15 aggressive orders" while noting two of them are not.
- **F-17** `PendingCommandHandler.cs:246` rebuilds through the public constructor and silently drops `ResolvedSlots`. Unreachable today; latent, unpinned. (GAP, XFILE.)
- **F-18** A pattern naming the same required slot twice appends duplicate entries; `Slots.Length - ResolvedSlots.Length` stays correct, but the identity is now load-bearing on an unpinned grammar shape. (LINE.)
- **F-19** Step 3b re-walks a pattern Steps 4 and 7 each walk again, with no cheap incompleteness pre-gate. (EFF.)
- **F-20…F-23** Conventions (CONV): a `//` block wedged between an XML doc and its member (the only instance in the codebase); a `========` divider introduced into three fixtures that use `--------`; a double blank line; `_resolvers` declared below a method body while its sibling field is at the top of the class.

## Angles that found nothing

**SIMP** — clean, and it specifically tested the `resolved` + `resolvedReason` two-field argument and found it survives. **GONE** — enumerated all 27 deletions and traced each to its replacement, including all ten `followUpResult.Value` rewrites mapping 1:1; one low-confidence note folded into F-9. **CONV** cleared the banner, `.meta` format, `Voxr` prefix, test naming, `#if UNITY_EDITOR` placement and assembly placement, and ruled the XML-doc asymmetry an improvement rather than a violation, with evidence.

## Verified clean, worth recording

- Both scoring formulas `(D−2)/D` and `(D−3)/D` re-derived independently from the constants; `0.250` for `"launch jackal"` confirmed. (CLAIM)
- The added test counts (4 EditMode, 26 PlayMode) match the diff; all six mutation-table rows map to tests that assert what the table says. (CLAIM)
- `SchemaVersion_IsFour` is the **only** existing test line modified anywhere in the diff. (CLAIM, GONE)
- `JsonUtility` null-string handling is correct, and the log is write-only, so the schema bump carries no deserialisation hazard. (PIT)
- Invariant 8 (editor-only) holds by construction. Invariant 6 (append-last) is pinned by a real test. (GAP, XFILE)
- `VoxrCommand` copy semantics, the `in` parameter, and `matchedCount` non-negativity all hold. (PIT, LINE)
- F25 is the only unmet **Must**, and no Must is deferred silently. (GAP)

The Unity EditMode/PlayMode counts (190/190, 664/664), the `Samples~` compile, and the six-row mutation-evidence table are **author claims**. Angles confirmed the tests exist and assert what the table says they assert; that they *ran* cannot be checked from here. On-device behaviour is human-only and claimed not applicable.

---

**No fixes have been applied.** Under the feature lane nothing is fixed without your ruling; this review is evidence for G2, not a mandate. F-1 is a confirmed crash on a path the PR's own test demonstrates is reachable, and F-2 is a possible canon violation — I would not merge on either until you have ruled.

jinwoo1601:
## Two open items from the review are now closed

### F-2 was **REACHABLE** — the fix closed a live hole, not a theoretical one

The review confirmed the *mechanism* but never established reachability (its verifier was killed by a rate limit). It has now been derived from source **and confirmed empirically** against the compiled parser at this head.

**The review's own example grammar does not reach it** — it fails at the tie, tripping `MissedRequired > MatchedRequired`. The shape is real but needs a longer pattern. A grammar that does tie (probe-verified: both score **0.6250**, `rivalCount = 1`):

```
slots:   {track} = "one"        {target} = "zulu"
A_intent : {track}  {target} set course for the gate alpha
B_intent : {?track} {target} set course for the gate bravo
utterance: "one set course for the gate"      ({target} and the discriminator both elided)
```

`NormalizeElement` folds `{?track}` against `{track}`, so these are siblings. A matched slot scores the same optional or required, but only a *required* one credits `matchedRequired` — so `B` latches `LeadingRequiredMissed` on `{target}` while `A`, whose `{track}` credited first, does not. Both are admissible, so `B` is recorded as `A`'s tied rival. The general condition is `k/(k+3) ≥ minScore`, i.e. **k ≥ 5** matched required elements after the anchor at the default gate.

**Isolating controls** — the only thing changing between rows 2 and 3 is whether `B`'s anchor was spoken:

| grammar / utterance | results |
|---|---|
| A alone, `"one set course for the gate"` | 1 (0.6250) |
| **B alone, same utterance** | **0 — barred** |
| B alone, `"one zulu set course for the gate"` (anchor spoken) | 1 (0.8750) |

**The chain at `f61da95`:** Step 3b resolves the *winner* `A` (which also misses `{target}`) — and that is required, because without it `A` is incomplete and never reaches the disambiguation branch at all. The same resolver that opens the door then filled `B`'s anchor. The answer path fires the stored choice verbatim with no completeness re-test, so saying **"bravo"** fired `B_intent` carrying a `target` the speaker never uttered. Ruling 3 violated.

**How this differs from what #126 ruled acceptable:** there the answer word *is* the missing anchor, so the speaker supplies it. Here the answer word is `bravo`, one element right of the anchor, and the anchor came from `ResolveOnce`. Nothing the speaker says at any point identifies the target.

**Preconditions** — three, all author choices, none of them defaults: `disambiguateSiblingTies` on (off by default); a resolver registered for the anchor slot; and a mixed-anchor sibling pair whose barred member's anchor is a **required slot**, with ≥5 matched required elements after it. Not present in the demo grammar. Reachable in a configuration a real integrator would plausibly build, not by accident.

A regression test is being added, with the isolating control shipped alongside it rather than left in this comment.

### A re-review of the fix commit found one residual hole — in the fix, not the original code

`47c94f5` was re-derived line by line. Five of the six fixes check out, including the `TryFollowUpSlotFill` behaviour-neutrality claim (derived, not merely trusted: the two consumers are name-lookups, so the slot *order* change is immaterial).

The sixth: **Step 3b's debounce floor is measured as of the start of the utterance.** It reads the debouncer before Step 7 records any of *this* utterance's fires, so a later candidate sharing an intent with one Step 7 already accepted reads off-cooldown here and on-cooldown there. Resolved for that reason, it skips Step 7's incomplete branch and is dropped by the debounce — where unresolved it would have entered pending and prompted the speaker. **What is lost is the prompt, not the utterance** (`acceptedCount` is already non-zero from the sibling that fired, so `OnUnrecognisedSpeech` stays suppressed either way).

**Left rather than chased, and the comment narrowed to say so.** Closing it means predicting in Step 3b what Step 7 will record later, and a gate that models another gate's future is a second place for the two to disagree — the exact failure the three-floor rule exists to prevent. The accurate statement of the rule is that resolution moves no command across a gate *as the gate stands when the utterance arrives*, and that is now what the code says.

### Bearing on #126

Both fixtures behind #126's ruling use a **literal** anchor, where "the speaker supplies the missing anchor by answering" holds. A **slot** anchor is the uncovered third case: a resolver can fill it silently, so no human ever names the argument. This PR removes the resolver from that path, so what #126 still permits is a chosen rival firing with its anchor slot **empty** — visible downstream as an unfilled required slot rather than as a wrong value. That is a materially different residue from the one the ruling was written against.

jinwoo1601:
## E-2 ruled: ship as built, scoring goes to a design branch

The maintainer has ruled **option 1** on the reachability finding. The hook ships with its measured limits stated honestly; the scoring question is now **#161**, where it will need G0/G1 because every option changes `scoring-model.md`, which is locked and complete.

**What this does not settle.** It is not a G2 acceptance — that is separate and still outstanding. And it does not answer the scoring question, only where it gets answered. Option 4 (exempt a resolver-fillable slot from the *denominator* rather than crediting it as matched — `"launch missiles"` reaches 0.667 instead of 0.75, declining to charge for what the game supplies rather than manufacturing evidence it was spoken) is recorded there as the strongest candidate the measurement surfaced, not as a decision.

**A documentation obligation this creates, for the post-G2 pass.** The product docs must carry the arithmetic rather than a vague caveat: the two formulas, the `D ≥ 5` / `D ≥ 8` thresholds, and plainly that an author's levers are pattern length or a lowered global `minScore`. An author who marks a slot resolvable and finds it never fires is overwhelmingly looking at a pattern too short to clear the gate, and they should be able to diagnose that without reading the parser. Recorded in the feature's requirements so it cannot be lost between here and the docs.

Everything else on this PR is unchanged and green: EditMode 190/190, PlayMode 678/678, `Samples~` mutation-verified, all 23 review findings addressed.

